Issue #73, opened 2015-05-07
str is too verbose


gaborcsardi:
From https://github.com/igraph/igraph/issues/507#issuecomment-99860452

gaborcsardi:
YEah, but we'll have to live with it....

sam-s:
why do we have to live with this bug?!

PR #77, opened 2015-05-12
comply with the compactness requirement of str

sam-s:
https://github.com/igraph/rigraph/issues/73
full=TRUE only if max.level is not supplied or bigger than 1

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -350,6 +350,31 @@ as.undirected <- function(graph, mode=c("collapse", "each", "mutual"), edge.attr
 }
 
 
+#' Reverse the direction of all edges.
+#'
+#' Return a new graph with the same edge and vertex attributes
+#' but with the edge directions reversed.
+#' If the graph is undirected, it is returned as is.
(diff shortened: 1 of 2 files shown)


gaborcsardi:
What's the `rev.igraph`?

gaborcsardi:
Also, this is kind of an abuse of `max.level`, so I'll rather close this. Something that makes proper use of `max.level` would be fine, though. E.g. 
- `1`: only summary
- `2`: summary + edges
- `3`: sumamry + edges + all attributes
Maybe.

sam-s:
@gaborcsardi: rev.graph: Reverse the direction of all edges.

gaborcsardi:
Yes, I get that, but what does it have to do with this pull request?

sam-s:
`str` is _required_ to always produce compact output.
it can _never_ print edges

sam-s:
this was supposed to be a separate pull request.
I am not very good with these github tricks.
sorry.

PR #80, opened 2015-06-01
Update curve_multiple to avoid warnings generated for graphs with loops (self-edges)

josilber:
Update curve_multiple function from R/plot.common.R to not generate warnings when processing graphs with loops (self-edges) -- see http://stackoverflow.com/questions/30556493/why-igraph-plot-warns-that-number-of-items-to-replace-is-not-a-multiple-of-repl/30558216#30558216 for the reported bug and the resolution


gaborcsardi:
Thanks!

PR #81, opened 2015-06-03
R: fix bug plotting named graphs with marked groups

aaronwolen:
Plotting marked groups works fine with unnamed graphs:

```r
karate <- graph.famous("Zachary")
plot.igraph(karate, mark.groups = list(grp = c(33, 34)))
```
Plot marked groups in unnamed graph

But fails with named graphs:

```r
V(karate)$name <- paste0("v", 1:vcount(karate))
plot.igraph(karate, mark.groups = list(grp = c("v33", "v34")))
```

```
## Error in layout[v, , drop = FALSE] : no 'dimnames' attribute for array
```

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -158,6 +158,8 @@ plot.igraph <- function(x,
   if (!is.list(mark.groups) && is.numeric(mark.groups)) {
     mark.groups <- list(mark.groups)
   }
+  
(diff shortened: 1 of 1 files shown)


gaborcsardi:
This seems to work with the latest version of igraph, so I am closing this.

PR #131, opened 2015-11-25
Remove strange printing behaviour

DexGroves:
`igraph` started printing numbers while plotting one day. It was unexpected.

Is this supposed to be here? Seems like a debugging print that got left behind.


gaborcsardi:
Thanks!

PR #142, opened 2016-02-10
DOC: typo weight instead of weigth

stnava:
there may also be a typo of similar nature in inst/benchmarks/local.scan.R


stnava:
feel free to ignore the pull request and just fix on your own.

gaborcsardi:
Thanks!

PR #144, opened 2016-03-04
Move documentation from remaining Rd files to roxygen

rundel:


ntamas:
Closing this as I moved around the remaining .Rd files so they are no longer in `man/`. Thanks anyway!

PR #145, opened 2016-03-04
Added support for make clean

rundel:
Existing Makefile did not have any ability to cleanup after itself. The files that are removed come from the exclusions listed in .gitignore (other than printr.R which seems like it shouldn't be there).

This change requires pull request #144 to prevent the current Rd files from being incorrectly deleted.


ntamas:
Thanks a lot! In the end I merged this without merging #144 because I moved the .Rd files around so that non-generated documentation is in `doc/` and not `man/`.

Issue #135, opened 2016-01-08
Fails to install on Mac OS X 10.11


rustyconover:
```
> devtools::install_github("igraph/rigraph")
Downloading GitHub repo igraph/rigraph@master
Installing igraph
'/usr/local/Cellar/r/3.2.2_1/R.framework/Resources/bin/R' --no-site-file  \
  --no-environ --no-save --no-restore CMD INSTALL  \
  '/private/var/folders/11/8ydpws752f100kjcy0jzhpdh0000gn/T/Rtmp80DI80/devtoolsaa70783b2c0f/igraph-rigraph-057cc9d'  \
  --library='/usr/local/lib/R/site-library' --install-tests 

* installing *source* package ‘igraph’ ...
....[good lines]...
clang -I/usr/local/Cellar/r/3.2.2_1/R.framework/Resources/include -DNDEBUG -I/usr/local/include  -I/usr/local/opt/gettext/include -I/usr/local/opt/readline/include -I/usr/local/opt/openssl/include -I/usr/local/include  -I/usr/local/include  -DUSING_R -I. -Iinclude -Ics -Iglpk -Iplfit -ICHOLMOD/Include -IAMD/Include -ICOLAMD/Include -ISuiteSparse_config -g -O2 -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/libxml2 -g -O2 -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk/usr/include/libxml2 -DNDEBUG -DNPARTITION -DNTIMER -DNCAMD -DNPRINT -DPACKAGE_VERSION=\"1.1.0\" -DINTERNAL_ARPACK -DIGRAPH_THREAD_LOCAL=/**/ -fPIC  -g -O2  -c CHOLMOD/Check/cholmod_check.c -o CHOLMOD/Check/cholmod_check.o
CHOLMOD/Check/cholmod_check.c:170:2: error: no member named 'print_function' in 'struct cholmod_common_struct'
        PRINTVALUE (Xx [p]) ;
        ^~~~~~~~~~~~~~~~~~~
CHOLMOD/Check/cholmod_check.c:112:2: note: expanded from macro 'PRINTVALUE'
        P4 (" %23.15e", value) ; \
        ^~~~~~~~~~~~~~~~~~~~~~
[...]

Issue #111, opened 2015-08-21
include directories listed in CFLAGS and CXXFLAGS instead of PKG_CPPFLAGS


pooryorick:
In R, only PKG_CPPFLAGS overrides the include directories that R records in its configuration when it is built, so header conflicts can occur if -I arguments are used in PKG_CFLAGS and PKG_CXXFLAGS, as rigraph does.  In our environment, the following change allowed the build to succeed:

``` diff
--- igraph/src/Makevars.in      2015-06-25 21:04:46.000000000 -0400
+++ igraph.new/src/Makevars.in  2015-08-21 17:56:50.000483000 -0400
@@ -1,12 +1,12 @@

-PKG_CFLAGS=-DUSING_R -I. -Iinclude -Ics -Iglpk -Iplfit \
-        -ICHOLMOD/Include -IAMD/Include -ICOLAMD/Include \
+PKG_CPPFLAGS=-DNDEBUG -DUSING_R -DUSING_R -DIGRAPH_THREAD_LOCAL=/**/ \
+       -DNDEBUG -Iprpack -I. -Iinclude -DPRPACK_IGRAPH_SUPPORT -Iprpack \
+       -I. -Iinclude -Ics -Iglpk -Iplfit -ICHOLMOD/Include -IAMD/Include \
+       -ICOLAMD/Include \
        -ISuiteSparse_config \
        @CPPFLAGS@ @CFLAGS@ -DNDEBUG -DNPARTITION -DNTIMER -DNCAMD -DNPRINT\
        -DPACKAGE_VERSION=\"@PACKAGE_VERSION@\" -DINTERNAL_ARPACK \
        -DIGRAPH_THREAD_LOCAL=/**/
-PKG_CXXFLAGS= -DUSING_R -DIGRAPH_THREAD_LOCAL=/**/ -DNDEBUG -Iprpack -I. \
-       -Iinclude -DPRPACK_IGRAPH_SUPPORT
 PKG_LIBS=@XML2_LIBS@ @GMP_LIBS@ @GLPK_LIBS@ $(FLIBS) $(LAPACK_LIBS) $(BLAS_LIBS)

 all: $(SHLIB)
```

gaborcsardi:
Thanks! Could you please send a PR?

Btw. the current version builds on several platforms, where did you have problems with it?

rundel:
[...]

PR #146, opened 2016-03-04
Change Makevars.in to use CPPFLAGS

rundel:
Fix for #111 and #135 (when combined with homebrew/homebrew-science#3386)


vtraag:
I believe this issue is no longer a problem, so I am closing this. 

You can install the package using `remotes::install_github("igraph/rigraph@master")`. Tests pass on the CI on macOS using R 4.0. Thanks for the contribution though @rundel ! If you do come across any remaining issues, feel free to report back!

PR #167, opened 2016-09-06
$ operator

ttriche:
as referenced in issue

### DIFF
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -311,21 +311,6 @@
 }
 
 
-#' @details
-#' The dollar-sign operator selects the induced subgraph 's' of graph 'x' 
-#' that results when a single vertex 'i' and its neighbors of order 1 are 
-#' used as the set of vertices to induce s from x.
-#' 
-#' @param x The graph.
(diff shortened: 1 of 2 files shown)


ttriche:
The ego-subgraph selection operation selects the subgraph s of graph g "around" vertex v, so 

g %around% v 

seemed like a reasonable syntactic compromise.   This addresses https://github.com/igraph/rigraph/issues/168

IgnatiusPang (commented):
Could probably be replaced by magrittr pipe %>%. This is probably just a syntactic sugar and perhaps not a priority to merge.

vtraag:
I think the crucial point was raised by @mbojan in https://github.com/igraph/rigraph/issues/168#issuecomment-246999705: how is this different from `make_ego_graph`? Indeed, he also suggests that this can simply be done by a `magrittr` pipe, similar to your observation @IgnatiusPang. 

I would be inclined to close this, unless somebody has good arguments for introducing such a functionality.

ntamas:
Closing as suggested above by @vtraag

Issue #161, opened 2016-07-05
User-friendly vertex size


gvegayon:
Working with vertex size is not easy on igraph as that is a function of the plotting region itself. This can be solved by rescaling the vertex size considering the plotting region, further, as done in Gephi, users could provide vertex size and relative scale (minimum/maximum sizes) to the plotting area (e.g. 0.5% and 1.0% of the plotting area). The following code does exactly that:

``` r
rescale_vertex_igraph <- function(
  vertex.size,
  par.usr=par("usr"),
  minmax.relative.size=c(.05,.01),
  adjust=200
) {
  if (!length(vertex.size)) return(
    rescale_vertex_igraph(1, par.usr, minmax.relative.size, adjust))

  # Adjusting x
  xrange <- range(vertex.size)
  xscale <- (par.usr[2] - par.usr[1])*minmax.relative.size
  vertex.size     <- (vertex.size - xrange[1] + 1e-15)/(xrange[2] - xrange[1] + 1e-15)*
    (xscale[2] - xscale[1]) + xscale[1]

  return(vertex.size*adjust/2)
}
```

The default in this function will return a vector of length `n` that can be passed to `plot.igraph(...,add=TRUE)`, resulting on vertices that take sizes between 0.5% and 1% of the plotting region, so if you change the layout (and hence the plotting region) the vertices will have the same size.

The `adjust` parameter follows a rescaling that igraph does internally when drawing figures via `symbols`. I would like to include this a patch in igraph but that implies adding extra parameters to the `igraph.plotting` function and I'm not completely aware of how doing that.

I write this post as a suggestion from Gabor during the useR! 2016 conference.

PR #172, opened 2016-10-06
feat: add relative size scaling to vertices in plots

gvegayon:
Following issue https://github.com/igraph/rigraph/issues/161, this PR makes a significan change in the way `vertex.size` is processed. In particular, instead of using the scaling `(1/200)` as seen throughout the code, `plot.igraph` now uses the function `i.rescale.vertex` which rescales the vertex size such that the smallest vertex is of size `vertex.relative.size[1]` and the largest one of size `vertex.relative.size[2]` (`vertex.relative.size` is a new parameter that has been added as `relative.size` in the list `i.vertex.default` and has a default values of `c(.01,.025)` [1%,2.5% of the plotting region]).

This new feature has been included to be compatible with `rglplot` and `tkplot` so all three plotting functions have somewhat same vertex sizes.

A neat example of this new feature is:

``` r
igraph_options(plot.layout=layout_nicely)
set.seed(312)
g <- barabasi.game(10)

oldpar <- par(no.readonly=TRUE)
par(mfrow=c(2,2))
set.seed(1);plot(g, main="Default")
set.seed(1);plot(g, vertex.size=degree(g), main="size=degree")
set.seed(1);plot(g, vertex.size=degree(g), vertex.relative.size=c(.05,.1),
                main="size=degree,relative.size=c(.05,.1)")
set.seed(1);plot(g, vertex.size=degree(g), vertex.relative.size=c(.025,.2),
                main="size=degree,relative.size=c(.025,.2)")
par(oldpar)
```

example

which has been included at the end of the file `plot.common.Rd`. Furthermore, the changes do not affect significantly examples in the `plot.common.Rd` file since by default all vertices occupy 2.5% of the screen.

One important drawback of this PR is that "simple" changes in vertex.size won't be reflected, e.g.

``` r
set.seed(1211)
g <- barabasi.game(10)

# These three are equivalent
oldpar <- par(no.readonly=TRUE)
par(mfrow=c(2,3), mar=rep(0,4))
set.seed(1);plot(g)
set.seed(1);plot(g, vertex.size=1)
set.seed(1);plot(g, vertex.size=2)

# But this not
set.seed(1);plot(g, vertex.size=1, vertex.shape="rectangle", vertex.size2=1)
set.seed(1);plot(g, vertex.size=1, vertex.shape="rectangle", vertex.size2=2)
set.seed(1);plot(g, vertex.size=2, vertex.shape="rectangle", vertex.size2=1)
par(oldpar)
```

example2

Finally, this PR fixes a couple of 'failures' that `testthat` showed due to a relatively recent update in which using `print` started to be a requirement within 'expectations'.

I understand that this may be a big change but the benefits are greater (in my opinion) since picking vertex sizes will now be more intuitive than before.


gvegayon:
Following CONTRIBUTING.md, ping @gaborcsardi @ntamas

vtraag:
In general, this possibility looks quite useful @gvegayon, thanks for contributing! It has taken slightly longer than a week to get back to you on this :wink: 

I will have to look at this in a bit more detail, but I had a more general question. The ability to specify *relative* vertex sizes is definitely useful, however, it is a pity that the ordinary `vertex.size` argument no longer works as originally. This will break some existing code of users that simply wanted to create larger (or smaller) vertices. I was wondering whether it would not be possible to do the following:
- Rename `vertex.size2` that you have now to `vertex.size`.
- Rename the `vertex.size` to `vertex.relative.size` to clarify you are specifying a *relative* size.
- Rename `vertex.relative.size` to `vertex.relative.scaling` to indicate the scaling of the relative vertex sizes that you have.

That way, users continuing to use `vertex.size` would see no difference, but users would be able to specify a (possibly more useful) `vertex.relative.size`, where the scaling of the `vertex.relative.size` is controlled by `vertex.relative.scaling`. Let me know what you think.

gvegayon:
No worries wrt to the timing, @vtraag! I see myself reflected on this 😛. Sure, that makes sense. I will take a look at this and try to implement one of your suggested solutions.

szhorvat:
If a breaking change is made, this would be an opportunity to think carefully about what the best way is to set vertex sizes. I'll admit I haven't yet looked into how this is done in the R interface yet, but for lack of time, I'll just throw a few Mathematica-inspired ideas here:

 - It is useful to set vertex size as a fraction of the shortest edge length, or as a fraction of the distance of the closes two vertices (in case R already has facilities to compute this in better than O(n^2) time).
 - It is also useful to set it as a fraction of a bounding region of all vertices (bounding box, bounding disk, whatever)
 - Finally, if R plotting generally supports this, it is also useful to set it in device units (e.g. millimeters/points for printing, pixels for screen, etc.) that is independent of plot coordinates. This is what is normally useful for text: the text size should be readable, regardless of whether the plot range is from 0 to 1 or 0 to a million.

krlmlr:
This now has conflicts, my apologies. Are you still interested in the change?

gvegayon:
Just solved the conflicts. I haven't tested the code, but it should still work. Will try it out tomorrow.

krlmlr:
The vignettes now fail: https://github.com/igraph/rigraph/actions/runs/8446422057/job/23140221959?pr=172#step:7:34 . Can you please take a look?

maelle:
What about the comments in https://github.com/igraph/rigraph/pull/172#issuecomment-791498917? Do we want to make this a breaking change?

maelle:
``` r
library("igraph")
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
g <- make_graph("Zachary")
plot(g)
#> Error in symbols(x = coords[, 1], y = coords[, 2], bg = vertex.color, : invalid symbol parameter
```



<sup>Created on 2024-05-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

For some reason in this reprex vertex.size becomes a vector of NA, that is passed to `symbols()`. Any idea what could be wrong?

krlmlr:
Want to sync main after #1413 and see what remains.

maelle:
@krlmlr @schochastics could we try and finalize this?

schochastics:
yes I think we need to finish this. Given I had my hands on the plot code I will take over if you dont mind. @gvegayon I promise we finish this before the PR is10 years old :D

schochastics:
Not completely thought everything through yet but would it make sense to simply introduce a logical switch `vertex.size.scaling` that defaults to `FALSE`. When it is `TRUE`, the new scaling is used. That way, the old sizing keeps working, but one can switch to the new scaling. It has to be made clear to the user that if the switch is `TRUE`, `vertex.size` is essentially ignored.

Any thoughts (cc'ing all involved: @szhorvat, @vtraag, @gvegayon, @maelle)

gvegayon:
> Not completely thought everything through yet but would it make sense to simply introduce a logical switch `vertex.size.scaling` that defaults to `FALSE`. When it is `TRUE`, the new scaling is used. That way, the old sizing keeps working, but one can switch to the new scaling. It has to be made clear to the user that if the switch is `TRUE`, `vertex.size` is essentially ignored.
> 
> Any thoughts (cc'ing all involved: @szhorvat, @vtraag, @gvegayon, @maelle)

That sounds good! What about highlighting the new feature somehow? Thanks for taking over, @schochastics!

schochastics:
It works now but before I finalize it, we need consensus that the logical switch is ok to use

``` r
igraph_options(plot.layout=layout_nicely)

g <- make_famous_graph("Zachary")
par(mfrow = c(3,3), mar=rep(0,4))

set.seed(1);plot(g, vertex.size = 10, vertex.size.scaling = FALSE)
set.seed(1);plot(g, vertex.size = 20, vertex.size.scaling = FALSE)
set.seed(1);plot(g, vertex.size = 30, vertex.size.scaling = FALSE)

set.seed(1);plot(g, vertex.size = 10, vertex.size.scaling = TRUE)
set.seed(1);plot(g, vertex.size = 20, vertex.size.scaling = TRUE)
set.seed(1);plot(g, vertex.size = 30, vertex.size.scaling = TRUE)

set.seed(1);plot(g, vertex.size = degree(g), vertex.size.scaling = TRUE)
set.seed(1);plot(g, vertex.size = degree(g), vertex.size.scaling = FALSE)
set.seed(1);plot(g, vertex.size = degree(g) * 2, vertex.size.scaling = FALSE)
```



<sup>Created on 2025-03-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

szhorvat:
I'm fine with the logical switch.

gvegayon:
Same here :D. I'm sooo happy this is coming to live after ~10 years!

schochastics:
almost there! @maelle can you help me with these `\dontshow{if (interactive()) withAutoprint` changes? I am note sure where they come from

schochastics:
need to fix some issues with rectangle vertex shapes

krlmlr:
Thanks!

Issue #194, opened 2017-05-16
get.adjlist very very slow, and potential fix


cfhammill:
For large graphs get.adjlist is very slow, for a graph with 125000 vertices, and 735000 edges

```
system.time(adj1 <- igraph::get.adjlist(graph, "out"))
   user  system elapsed 
665.323   1.508 667.865 
```

The C-code to extract the neighbours isn't the culprit

```
system.time(.Call("R_igraph_get_adjlist", graph, 1, package = "igraph"))
   user  system elapsed 
  0.011   0.000   0.012 
```

The slowness can be traced to:

```
res <- lapply(res, function(x) V(graph)[x + 1])
```

I suspect the proper vertex list indexing is overkill here and causing an extreme slowdown. Since the return-type from c-code should be guaranteed to be a list of integer vectors, and the doc promises to return a list of integer vectors could we get away with:

```
vvec <- unclass(V(graph))
res <- lapply(res, function(x) vvec[x + 1])
```
Or potentially even:

```
res <- lapply(res, `+`, 1)
```

Benchmarks:
```
system.time(na1 <- new_adj_list1(graph, mode = "out"))
   user  system elapsed 
  0.187   0.000   0.188

system.time(na2 <- new_adj_list2(graph, mode = "out"))
   user  system elapsed 
  0.069   0.000   0.069

all.equal(na1, na2)
[1] TRUE

all.equal(lapply(adj1, function(x) as.numeric(unclass(x)))
        , na1)
[1] TRUE
```

gaborcsardi:
Thanks! Would you like to submit a pull request?

cfhammill:
Sure thing. Do you prefer the first or second version, the second has a precedent in  `ego` https://github.com/igraph/rigraph/blob/665d71ebd40cdfe9b996a4f50c35d65b791e4102/R/structural.properties.R#L1702

gaborcsardi:
Second is good. But we should probably add a class. G

On 25 May 2017 07:12, "Chris Hammill" <notifications@github.com> wrote:

Sure thing. Do you prefer the first or second version, the second has a
precedent in ego https://github.com/igraph/rigr
aph/blob/665d71ebd40cdfe9b996a4f50c35d65b791e4102/R/
structural.properties.R#L1702

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/igraph/rigraph/issues/194#issuecomment-304019908>, or mute
the thread

[...]

PR #196, opened 2017-05-25
Speed up `as_adj_list` if igraph_opt("return.vs.es") is false

cfhammill:
Also updates the docs to point out that this option exists. Not a fully satisfying solution to #194, but provides a simple work around.


gaborcsardi:
Thanks! I assume that you checked that after addind `create_vs`, it is still fast. :)

cfhammill:
It is fast if you set igraph_options(return.vs.es = FALSE)  yes it is still fast. I don't know if `create_vs` could be sped up.

vtraag:
@cfhammill , I guess that PR #197 renders this PR obsolete and should be preferred over this one?

ntamas:
This PR seems to be superseded by #197 so I'm closing this.

PR #197, opened 2017-05-25
Vertex access speedups

cfhammill:
A speed up for `ego` and `as_adj_list`, this goes without adjusting `igraph_opt("return.vs.es")` so it will benefit users automatically. Time reduction for `as_adj_list` on a 40000 node graph dropped from 500 -> 9 seconds.  The key time saver is only computing `V(graph)` once. This is done with the new `unsafe_create_vs` function documented in code and the commit message. Still ~100 fold slower than `igraph_options(return.vs.es = FALSE)`, but you get the benefit of reclassing the integer vectors `igraph.vs` and associating the graph pointer.


R/iterators.R
```diff
@@ -306,7 +317,7 @@ create_es <- function(graph, idx, na_ok = FALSE) {
 
 simple_vs_index <- function(x, i, na_ok = FALSE) {
   res <- unclass(x)[i]
-  if (!na_ok && any(is.na(res))) stop('Unknown vertex selected')
+  if (!(na_ok | !any(is.na(res)))) stop('Unknown vertex selected')
```
vtraag, line 320: I don't exactly understand this change. In your commit message you say

> This prevents `any(is.na(vs))` from getting evaluated if `na.ok = TRUE`

but isn't this also the case with the original conditional `!na_ok && any(is.na(res))`? That is, if `na_ok = TRUE`, there is no need to evaluate the remaining condition, since the entire condition is already known to evaluate to `FALSE`. Or is there some specific reason why this is not the case here?

R/iterators.R
```diff
@@ -208,6 +208,17 @@ create_vs <- function(graph, idx, na_ok = FALSE) {
   add_vses_graph_ref(res, graph)
 }
 
+# Internal function to quickly convert integer vectors to igraph.vs
+# for use after c code, when NA and bounds checking is unnecessary
+unsafe_create_vs <- function(graph, idx, verts = NULL){
+  if(is.null(verts))
+    verts <- V(graph)
+
+  res <- simple_vs_index(verts, idx, na_ok = TRUE)
```
vtraag, line 217: I guess the reasoning for using `na_ok = TRUE` here is that `simple_vs_index` then does not need to check `any(is.na(res))`? Semantically this is a bit strange. Perhaps the argument should then be renamed `check_na` or something, instead of `na_ok`.

vtraag (commented):
This looks potentially quite useful, thanks for contributing! 

I was thinking about one thing however. You are saying that the key time saver is the computation of `V(graph)` only once. I therefore wonder whether it would not be better to simply keep the existing `create_vs` function, but add the optional `verts` argument you now added to `unsafe_create_vs`? I am not sure of the performance impact of the `ifelse`, but I can imagine that is relatively minor. That is, we could hence have no need of the `unsafe_create_vs` and only have the "safe" `create_vs` function, but which can potentially be sped up if we pass in the vertices. There are a few other such use cases in the code base that might benefit from a similar speed up. What do you think about this?

R/iterators.R
ntamas: `na_ok` is pretty common in other places in `R/iterators.R` and this is all internal anyway, so I'm okay with this.

Semantically, `na_ok` basically means "it's okay if there is an NA value anywhere in the input vector". This _can_ make sense in other invocations of `simple_vs_index`, even though in this particular case we also know that there won't be any NA in the input vector.

R/iterators.R
ntamas: Not sure about this; there would clearly be a difference here if `na_ok` is a vector because the `&&` operator is not vectorized (it only looks at the first argument of `na_ok` if it is a vector), while `|` is vectorized. However, I think `na_ok` is meant to be a single logical value anyway so I don't think this change is necessary.

ntamas:
This PR seems okay to be, but there are multiple occurrences of `lapply(...,create_vs, ...)` in the codebase; @vtraag do you think it would be safe to also replace them with `lappy(..., unsafe_create_vs, ...)`? Most of the examples are in `cliques.R` and `structural.properties.R`.

ntamas (approved):
I think I am familiar enough with the R codebase to judge that this is safe to merge. Thanks a lot for your contribution, and sorry about the delay!

PR #222, opened 2017-08-30
Export of the as_graphnel function

David-J-R:
The export-tag for the graphnel function was commented twice withe #'. Therefore the function isn't exported and the tag is just displayed in the examples.


vtraag:
This is now already correct in the latest version:

https://github.com/igraph/rigraph/blob/75e495a2f94f2c881160b49a3eb2e08293db383c/R/conversion.R#L459

Thanks for contributing though!

PR #246, opened 2017-12-06
graph_from_graphnel wasn't being exported

rmflight:
Somehow looks like a typo got introduced, and the export directive on this function was lost, which means it was not available to others to use


rmflight:
Never mind, see there is a pull request already here from Aug 30 dealing with the same issue.

PR #250, opened 2018-01-22
Fix warning due to factors in graph_from_data_frame

clhunsen:
When converting a data.frame with an POSIXct column to an igraph object,
a warning is issued due to ineffective checking of classes:
> Warning message:
> In if (class(newval) == "factor") { :
>   the condition has length > 1 and only the first element will be used

Th warning is true as a POSIXct object has actually two classes:
"POSIXct" and "POSIXt".

To remove the warning, the class of edge and vertex attributes is now
checked using the infix notation "%in%".

------------------


clhunsen:
Any news on this?

IgnatiusPang (approved):

IgnatiusPang (approved):
The changes look fine to me.

vtraag:
@clhunsen and @IgnatiusPang , I have the impression that it would be idiomatic in R to check using `inherits`, i.e. `inherits(newval, "factor")`. Is there any reason why it should be done using the `%in%` operator?

clhunsen:
> @clhunsen and @IgnatiusPang , I have the impression that it would be idiomatic in R to check using `inherits`, i.e. `inherits(newval, "factor")`. Is there any reason why it should be done using the `%in%` operator?

There is absolutely no reason to use the infix operator `%in%`. I just haven't thought about using `inherits`. 

What is your preferred way to proceed, @vtraag: Do you want me to perform a rebase onto the current `dev` and to change the PR through a force-push? Or do you want to apply the change yourself somehow? Or ...?

vtraag:
I already merged in the latest code from `dev` to this PR, and all checks pass. Feel free to change it to `inherits`, you should be able to simply push to your branch again (just make sure you pull first in include the merge I did). If that is done, I should able to squash-merge it all together.

clhunsen:
I added a commit with the small code adjustment. Thanks for the suggestion, @vtraag! :slightly_smiling_face:

vtraag:
Thanks for contributing @clhunsen !

PR #260, opened 2018-03-13
helper function for magrittr and igraph

beemyfriend:
It just changes the first argument of the RHS magrittr function, 'graph' is now the second argument and the first argument is either an edge vector or a node vector


beemyfriend:
Whoops, it turns out the order of the argument doesn't matter with magrittr.

PR #300, opened 2018-10-06
add a little more descriptive error message in norm_coords function

mirca:
This is a tiny improvement to the error message given when `layout` is not a valid matrix.


gaborcsardi:
Thanks!

PR #304, opened 2018-11-18
plot.igraph can return coordinates

MariaHei:
plot.igraph now has a parameter 'returnCoordinates'. If set to TRUE, the coordinates and labels of the vertices are returned as a data.frame. This makes it possible to analyse the location of vertices in space and the function would be applicable to a lot more use cases.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -59,9 +59,13 @@
 #' the marked vertex groups. It is in the same units as the vertex sizes. If a
 #' vector is given, then different values are used for the different vertex
 #' groups.
+#' @param returnCoordinates A boolean specifiying if the coordinates and labels 
+#' of the vertices are to be returned.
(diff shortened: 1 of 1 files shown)


vtraag:
You can obtain the layout from a graph already using the various layout functions, which already provides you the coordinates. For example, you can simply do

```R
layout <- layout.kamada.kawai(g)
```
In fact, `plot.igraph` is also calling such layout functions to obtain the coordinates, unless the coordinates are explicitly provided, by storing them as `g$layout <- layout`, or by passing them as an argument `plot(g, layout=layout)`.

I think this specific functionality would therefore not be required. What is exactly the rational behind doing this explicitly in the plotting function, instead of obtaining the layout for a graph explicitly?

ntamas:
Closing due to inactivity.

PR #307, opened 2018-12-12
coerse membership to a factor before as.numeric

aghaynes:
this should allow strings to be passed as membership, rather than just numerics.

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -422,7 +422,7 @@ modularity <- function(x, ...)
 modularity.igraph <- function(x, membership, weights=NULL, ...) {
   # Argument checks
   if (!is_igraph(x)) { stop("Not a graph object") }
-  membership <- as.numeric(membership)
+  membership <- as.numeric(as.factor(membership))
   if (!is.null(weights)) weights <- as.numeric(weights)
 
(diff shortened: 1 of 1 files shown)


aghaynes:
my motivation for this is that RStudio/R completely crashed (without any error message) when I used a string rather than a numeric.

vtraag:
Thanks for contributin this @aghaynes! It would indeed be nice to have "named" communities, so that arbitrary vectors would also be suitable as membership, and that these names would also be consistently used throughout the code base. However, this is a much more extensive change. By allowing this in only a single function (`modularity`), but not elsewhere, this will presumably raise questions.

Instead of allowing arbitrary vectors to be used as an input for `modularity`, I think it would be better to simply raise an error and stop if the membership is not numerical. This should also prevent R from crashing, I would think. If have opened PR #455 to address this. Could you take a look to try to see if this solves your problem?

aghaynes:
Sorry @vtraag, I have absolutely no idea where i was using this anymore... I should have included a reprex... live and learn I suppose...

vtraag:
No problem, this happens!

Issue #268, opened 2018-05-02
Build against external igraph


QuLogic:
You can build python-igraph against an external/system copy of igraph, but there seems to be no option for it in this R version.

I can almost write a patch with an option to `Makevars.in` to do so, but there are a couple things that hinder it a bit:
1. All the sources get dropped directly in `src`, which is not a problem, but does make it a little hard to remove the extra files. It would be a bit nicer if they were all put in a single subdirectory.
2. The linked igraph commit does not seem to be a released version, which would generally be the one that was packaged. It would be nice if non-tagged versions were avoided.

gaborcsardi:
This is not possible currently, as the sources are compiled with different flags for the R package.  See the  `USING_R` flag.

QuLogic:
OK, good point; I had a look through those, and I _think_ most of them are benign:
1. All functions using printf are commented because R warns about it in shared libraries, but this shouldn't matter for an external dependency. Since they're all commented, the R code could never have called them and the one that _is_ called from C is avoided by `igraph_set_progress_handler`.
2. `igraph_error` has an `abort` but that should have already been avoided by calling `igraph_set_error_handler`.
3. The RNG seems to be linked to R. I'm not totally sure of the benefits/costs for this one are.
4. One `IGRAPH_ERROR` is replaced by `REprintf` in `src/community.c`; I'm not sure why this one call is special?

gaborcsardi:
Before we go into the details here, what is the benefit of linking to an external lib?

QuLogic:
Sorry, I should have provided more context earlier. I am trying to create a package for Fedora of igraph. We already have an igraph package (of the C library) and two python-igraph packages (for 2 and 3) which link against it. It is no longer a hard requirement, but I would _like_ it if the R package also linked against the common C library to minimize duplication. If things are super-entangled that would be impossible, but it does not appear to be the case here.

gaborcsardi:
The releases of the C lib and the R package are not synchronized, sadly. So I am afraid that you would have complications from this setup.

QuLogic:
So long as the C lib followed something like semver and the R version only used releases, then it would remain workable. However, for the time being, I have decided to keep the bundled igraph, and only unbundle arpack and uuid. But it would still be nice to be able to do this in the future.

[...]

PR #372, opened 2020-02-24
Updated C igraph core to 0.8.1, updated tests and updated Travis build configuration

vtraag:
This PR is meant to update the R interface to the most recent `igraph` C core version. 

There have been a number of API changes, which also need to be reflected in the R interface. 

Some issues were fixed, which affected some of the tests (actually now leading to correct results, where some tests were previously incorrect). I've also updated all statements of the form `expect_that(..., is_true())` to `expect_true(...)` (and similarly for the `false` variant), since I got too many warnings that `is_true()` was deprecated. 

There are still a number of errors in the tests, which I need to check out. Incidentally, there also seems to be some bug in the `avg_nearest_neighbour_degree` when using a weighted graph, presumably there is an error there in the `igraph` C core, but I have not yet managed to nail it down and make a reproducible example. For now I commented out that test in order to proceed with other tests (otherwise it aborts with a `free(): invalid pointer` problem). There are also still some other errors in the tests that need to be resolved. 

However, I already wanted to share the progress so far, to make sure that there are no obvious things I'm overlooking or doing incorrectly. Feedback is welcome @gaborcsardi.

Note that the `functions.def` was still incorrect in `igraph`. I have created PR https://github.com/igraph/igraph/pull/1338 to address this, which of course affects to what commit the `cigraph` submodule should point.


vtraag:
I believe the PR is essentially good to go, and the tests pass locally. I only encounter one issue still: after all tests succeed, I receive an error:
```
Error in vapply(issues, issue_summary, FUN.VALUE = character(1)) : 
  values must be length 1,
 but FUN(X[[1]]) result is length 90
```
This seems to have to do with something with the `reporter` in the `testthat` framework, but I don't yet fully understand what exactly. I'm not sure what can be done about it, as it doesn't seem to be something that can be changed by any of the code in `rigraph`.

@gaborcsardi it seems that the PR build fails because `r-builder` does not support the latest R version  (currently 3.6.2). I am not sure if it is still necessary to use the `r-builder` script? It seems that Travis supports building [R packages](https://docs.travis-ci.com/user/languages/r) without requiring specific `r-builder` scripts. Perhaps it would be possible to switch to building using the Travis supported configuration? I'd be happy to give it a try, but I don't have much experience with R.

vtraag:
@gaborcsardi , I updated the C core to version 0.8.1. Everything now builds correctly on R 3.5 and R 3.6. Two warnings remains:
1. ```* checking pragmas in C/C++ headers and code ... WARNING``` This goes for GLPK included source files, I don't think we can correct this right now.
2. Some problem with `plfit`:
```
Found ‘abort’, possibly from ‘abort’ (C)
    Object: ‘plfit/sampling.o’
  Found ‘rand’, possibly from ‘rand’ (C)
    Objects: ‘plfit/mt.o’, ‘plfit/sampling.o’
```
This will be corrected in the C core version 0.8.2, but this should not pose any problems.

The current development version of R shows a few issues, these mainly arise because `class(obj)` returns multiple values, most problematic cases involve `c("matrix", "array")`. Perhaps some of the `class(obj) == "type"` can be replaced by `inherits(obj, "type")`? I am not 100% whether this is the best approach. I assume that the current development version of R is R 4 which is due to appear on April 24, which still allows for some time to correct this, so this PR could still be merged.

If this PR can be merged, I can start working on adding some of the new functionality that has become available in the latest release of the C core. If there are any additional things that need to be done before this can be merged, let me know.

szhorvat:
@gaborcsardi Is it going to be an issue for this PR that recent R/igraph versions have not been pushed to this GitHub repo? The PR is therefore based on an older version than the current (the only version number I could see in this repo was 1.1 but the latest on CRAN is 1.2.5 ...)

gaborcsardi:
@szhorvat Latest releases are from the `cran` branch.

vtraag:
@gaborcsardi , I'm a bit lost as to the development on `rigraph`. In the [contributing guide](https://github.com/igraph/rigraph/blob/dev/CONTRIBUTING.md) is says that contributions should be made against the `dev` branch. Do you (intend to) use the `dev` branch at all? Should this PR be (re)targeted at the `cran` branch? I'd be happy to rebase against the `cran` branch.

gaborcsardi:
I can merge the cran branch into dev. OTOH, the C lib repo also has a cran branch that is used in the R package...

vtraag:
OK, good, whatever is easiest to work with. Good to know that the `cran` branch from `rigraph` uses the `cran` branch from `igraph`, I wasn't aware. We should probably also merge back in those commits in the `cran` branch in the master branch, shouldn't we (cc @szhorvat and @ntamas) ?

ntamas:
> We should probably also merge back in those commits in the cran branch in the master branch

Yes, it would be great.

QuLogic:
It would be nice if, since this is based on a tagged version, this meant that rigraph could be compiled against an existing igraph library (#268).

vtraag:
@gaborcsardi , thanks for updating the `dev` branch, I will resolve the conflicts in the coming week.

vtraag:
We can close this in favour of PR #380.

PR #380, opened 2020-03-23
Updated C igraph core to 0.8.1 and updated Travis build configuration.

vtraag:
This is a continuation of PR #372, but rebased against a more recent version of the `dev` branch (after PR #379 was merged). During the rebasing I also cleaned up a bit the different commits.

This PR updates the R interface to the recently released C core version 0.8.1 and it provides an improved Travis build script that uses Travis' native approach for R.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 /src/
 /version_number
 /configure.ac
+igraph.Rcheck/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,29 +1,39 @@
 
-language: c
-sudo: required
+language: r
+
+warnings_are_errors: false
+
+r:
+  - oldrel
+  - release
+  - devel
 
 before_install:
-- curl -OL https://raw.githubusercontent.com/metacran/r-builder/master/pkg-build.sh
-- chmod 755 pkg-build.sh
-- ./pkg-build.sh bootstrap
-- export TRAVIS_COMMIT_MSG="$(git log --format=%B --no-merges -n 1)"
-- export PATH="$PATH:$(./pkg-build.sh r_path)"
+  - Rscript -e 'install.packages(c("roxygen2", "devtools", "irlba", "pkgconfig"))'
+  - make
 
-install:
-- ./pkg-build.sh install_github hadley/lazyeval igraph/igraphdata gaborcsardi/roxygen gaborcsardi/pkgconfig
-- ./pkg-build.sh install_r magrittr falsy roxygen2
-- git submodule update
-- make
-- ./pkg-build.sh install_deps
-- ./pkg-build.sh install_bioc graph
-- ./pkg-build.sh install_r rgl ape scales
+apt_packages:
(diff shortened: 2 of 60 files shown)


szhorvat:
@vtraag Is this ready? @gaborcsardi Any comments? I'd be nice to bring the R interface up to date with the C core.

vtraag:
> @vtraag Is this ready?

Yes, as far as I know. There might be some updates necessary because of the recent release of R 4.0.0, but I think that could be done after this PR is merged. Most importantly, after this PR is merged, I can also more easily start working on integrating some of the other contributions from the igraph core.

gaborcsardi:
This is too big for me to review, and it has a lot of unrelated changes. It makes sense to break these into smaller PRs. 

Also, the latest R releases were from the cran branch, so we should probably rename that to dev, and continue development there.

In particular the cran branch has an `expect_that()` function, so there is no need rewriting all test cases.

vtraag:
> Also, the latest R releases were from the `cran` branch, so we should probably rename that to `dev`, and continue development there.

I'm fine with working either against the `dev` branch or the `cran` branch. I understood from [this comment](https://github.com/igraph/rigraph/pull/372#issuecomment-602107210) that you merged `cran` to `dev` branch. This is why I rebased this PR against the updated `dev` branch.

What do you propose to do? Feel free to rename the `cran` branch to `dev`, and then I'd be happy to open a new PR.

Or are you referring to the `cran` branch from `igraph` itself? We should indeed still merge those changes back into the `master` branch (or `develop` branch if there are API breaking changes). I will try to take a look shortly. I think it would be preferable if we would not work with a specific `cran` branch, but only with the `master` branch, do you agree also @ntamas, @szhorvat?

> In particular the `cran` branch has an expect_that() function, so there is no need rewriting all test cases.

I'm fine with not rewriting all the test. I just made all necessary changes in order for the build to pass without errors.

> It makes sense to break these into smaller PRs.

I'd be happy to break it into smaller PRs. Do you have any suggestion?

gaborcsardi:
Right, of course I forgot that I merged cran into dev...

gaborcsardi:
I'll merge this and we'll see if it needs any changes later...

gaborcsardi:
Thanks!

vtraag:
Regarding the status of the `master` branch, in the `CONTRIBUTING.md` it states

> (We build the `master` branch from the `dev` branch automatically, to make sure that the repo is compatible with the `devtools` R package which uses the `master` branch by default.)

Is this still true? It seems that the last commit in the `master` branch dates back to 2015. Is it necessary to keep on using `dev` and `master` branches separately, just to satisfy `devtools`? As far as I see, it is also possible to specify a `ref` to any branch (or tag) when using [`install_github`](https://remotes.r-lib.org/reference/install_github.html). Does `install_github` not work correctly on the `dev` branch? 

If we could simply use the `master` branch consistently it might perhaps simplify some things.

PR #393, opened 2020-05-20
Corrected check on NA for weights and capacities.

### LINKED ISSUE #392: Incorrect passing of vectors that contain NA
Vectors that contain an `NA` value are incorrectly passed to the C library. This appears in several places in `types-R.def`. The incorrect code is

https://github.com/igraph/rigraph/blob/a14fc50abf9bf3ae17b870ccbc21f88d0eefed6e/tools/stimulus/types-R.def#L189-L193

This means that the vector will be passed if there is at least one element that is not NA. What is intended is that the vector should be passed if not any element is NA, instead of any element not NA. In other words, this should be

```R
if (!is.null(%I%) && !any(is.na(%I%))) { \
    %I% <- as.numeric(%I%) \
} else { \
    %I% <- NULL \
}
```

vtraag:
This fixes #392. Vectors of weights or capacities are now only passed to the C core if there are no NA values at all in the vector.

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -456,10 +456,13 @@ modularity_matrix <- function(graph, weights=NULL) {
   if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
   weights <- E(graph)$weight
   }
-  if (!is.null(weights) && !any(is.na(weights))) {
-  weights <- as.numeric(weights)
-  } else {
-  weights <- NULL
+  if (!is.null(weights)) {
+    if (!any(is.na(weights))) {
+      weights <- as.numeric(weights)
+    } else {
+      warning("Weights vector contains NA elements, ignoring the weights vector.")
+      weights <- NULL
+    }
   }
 
   on.exit( .Call(C_R_igraph_finalizer) )
@@ -930,10 +933,13 @@ cluster_spinglass <- function(graph, weights=NULL, vertex=NULL, spins=25,
   if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
     weights <- E(graph)$weight
   }
-  if (!is.null(weights) && !any(is.na(weights))) {
-    weights <- as.numeric(weights)
-  } else {
-    weights <- NULL
+  if (!is.null(weights)) {
+    if (!any(is.na(weights))) {
+      weights <- as.numeric(weights)
+    } else {
+      warning("Weights vector contains NA elements, ignoring the weights vector.")
+      weights <- NULL
+    }
   }
 
   update.rule <- igraph.match.arg(update.rule)
@@ -1027,17 +1033,23 @@ cluster_leiden <- function(graph, objective_function=c("CPM", "modularity"),
   if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
     weights <- E(graph)$weight
   }
-  if (!is.null(weights) && !any(is.na(weights))) {
-    weights <- as.numeric(weights)
-  } else {
-    weights <- NULL
+  if (!is.null(weights)) {
+    if (!any(is.na(weights))) {
+      weights <- as.numeric(weights)
+    } else {
+      warning("Weights vector contains NA elements, ignoring the weights vector.")
+      weights <- NULL
+    }
   }
 
   # Parse initial_membership argument
-  if (!is.null(initial_membership) && !any(is.na(initial_membership))) {
-    initial_membership <- as.numeric(initial_membership)
-  } else {
-    initial_membership <- NULL
+  if (!is.null(initial_membership)) {
+      if (!any(is.na(initial_membership))) {
+      initial_membership <- as.numeric(initial_membership)
+    } else {
+      warning("Initial membership vector contains NA elements, ignoring the initial membership vector.")
+      initial_membership <- NULL
+    }
   }
 
   # Parse node weights argument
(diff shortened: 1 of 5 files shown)


vtraag:
@gaborcsardi for some reason the notification from Travis to GitHub that the build is complete seems to be have gone wrong. But all builds did pass.

szhorvat:
Question: Do I understand it right that this approach effectively ignores weights **with no message** whenever there are some `NA`s? Isn't that very likely to cause user frustration? Wouldn't a warning message be better?

ntamas:
Your understanding is correct, and yes, a warning message would probably be even better.

vtraag:
Yep, a warning message would indeed be preferable. Perhaps it is documented somewhere that vectors containing `NA` are not used, but still, a warning message would be preferable.

This PR only ensures this actually works correctly. I'd be happy to throw in some warnings in the relevant places.

vtraag:
I've rebased it against the `dev` branch and added warnings.

gaborcsardi:
You can change this for the functions that really don't support NA weights, but don't change the templates, because some functions do.

szhorvat:
There is a lot of code duplication. Would it make sense to have a small helper function for this?

vtraag:
> You can change this for the functions that really don't support NA weights, but don't change the templates, because some functions do.

I'm not sure I understand. At the moment, any vector is passed if *at least* one element is not NA. Are you saying this is correct for some functions? These NA values are translated to `NaN` in the C core, if my understanding is correct, which typically leads to problems.

gaborcsardi:
> I'm not sure I understand. At the moment, any vector is passed if at least one element is not NA. Are you saying this is correct for some functions? 

Yes. Not all functions you all the edge weights.

gaborcsardi:
In general, I don't think it is the R interface's job to check for NA and NaN values. Ideally the C code should handle that, and error if they are not allowed.

ntamas:
I think @gaborcsardi has a point here. If we force high-level interfaces to do sanity checks, then we need to do it in all high-level interfaces eventually; a lot more work than delegating it to the C core so the high-level interfaces can throw all sorts of junk at the C layer.

szhorvat:
On the other hand:

 - Some high-level interfaces simply cannot pass in a NaN, so the time spent on the check is wasted.

 - In R, there is a difference between NA and NaN. In my limited R experience, people readily use NAs in R and expect them to behave. NaNs, not so much. Also, people generally understand that attempting to use NaNs leads to a GIGO situation. This makes the R interface special: it is much more important to check for NAs because naive users _will_ expect them to work, and even not-so-naive users are likely to pass them in accidentally.

ntamas:
> Some high-level interfaces simply cannot pass in a NaN, so the time spent on the check is wasted.

But you can pass in a NaN from C code directly so it would be nice if we could ensure that you cannot crash the C core simply by feeding it with a NaN. GIGO situation is okay IMHO, but a crash is not.

I don't know whether this particular crash is caused by a NaN or something else -- as you mentioned, NA and NaN are not the same, so if NAs are an R-specific thing, then of course we should check for them in the R layer.

szhorvat:
I just tried in Python, and it does not seem that easy to accidentally produce a NaN or even and Inf. E.g., 0.0/0.0 leads to a `ZeroDivisionError` and while numpy does produce NaNs, it shows a clear warning when it happens. In Mathematica, there is no straightforward way to pass in a NaN. In R, 0/0 calmly returns a NaN and NAs are used all the time when working with real-world datasets.

szhorvat:
> GIGO situation is okay IMHO, but a crash is not.

Yes, I agree about that. Incorporating the check everywhere without a performance loss is going to take quite a bit of work though. Just looping over the weight array is not a very good thing if some algorithms may not even access all weights. Putting a naive `isnan` in inner loops is also not a good idea because it's hard to guarantee that it doesn't result in a function call (and therefore a potentially major slowdown). Of course, `x!=x` (or a macro doing the same) would work.

I did wonder why the betweenness algorithm crashes, and if it could indicate something more serious than the presence of NaNs, but in the end it's not surprising: any comparison, i.e. `==`, `>`, `<`, etc., with NaN results in a `false`. The only exception is `!=`. Algorithms that deal with shortest paths need to use inequalities, so it's not that surprising if that if neither `x > y` nor `x <= y` is true then something may go really wrong.

ntamas:
A fairly easy way to produce NaN in Python without any warnings:

```
float('nan')
```

Another one, which can actually happen if you deal with infinities:

```
>>> inf = float('inf')
>>> inf / inf
nan
```

This can happen, for instance, if you try to normalize a vector naively when one of the components of the vector is `inf`. So it's not rare at all.

szhorvat:
> A fairly easy way to produce NaN in Python without any warnings:
> 
> ```
> float('nan')
> ```

I meant accidentally doing it. But point taken—it's perhaps better to check in C.

gaborcsardi:
> Just looping over the weight array is not a very good thing if some algorithms may not even access all weights.

But isn't that what you are suggesting to add to the R interface? :)

vtraag:
OK, so let's agree to ensure that the C core handles NaNs correctly. That does not mean we need to do input checks always on all entries, it just means that is shouldn't crash. The example of the strength is actually a good example. The calculations return NaNs where they are expected to, there is even no need for any check on NaN.

szhorvat:
> > Just looping over the weight array is not a very good thing if some algorithms may not even access all weights.
> 
> But isn't that what you are suggesting to add to the R interface? :)

Yes, but the gist of my argument was that this check is mainly necessary for the R interface only.

Anyway, it seems we came to a decision to try to do it in C. That is not likely to happen for v0.8.x (IMO it should be left for 0.9) but it's desirable to have an R/igraph release based on 0.8. Now the question is:

 - Should this PR be discarded?
 - Or should the checks be temporarily added for functions where it makes sense (i.e. _not_ strengths) just for the releases based on 0.8?

szhorvat:
See https://github.com/igraph/igraph/issues/1560

vtraag:
Yes, I agree, let's leave it for 0.9.0 to do this systematically, since it will take quite a bit of work to check this for all functions.

For the current 0.8.x series, perhaps we should just correct these checks where they actually lead to known (i.e. reported/observed) errors? This is probably still best done in the C core.

vtraag:
I am closing this PR, since we won't solve this problem in the R repository. Any follow up discussion should be done in https://github.com/igraph/igraph/issues/1560 as mentioned earlier by @szhorvat .

Issue #394, opened 2020-05-23
graph_from_data_frame does not simplify graph


SamGG:
Hi,
Thanks for maintaining the important igraph package giving a uniform access to many algorithms.
I think that the graph_from_data_frame function is not explicit concerning the directed argument. When entering a not directed graph with duplicated links, one might expect that the graph will be simplified implicitly. It is not the case, and I think this should be clearly stated in the documentation.
I don't know how an undirected and unsimplified graph is handled by community detection algorithms.
Best.

``` r
library(igraph)
#> 
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#> 
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#> 
#>     union
df = data.frame(from=c(1,2,3,4,1), to=c(2,1,3,5,2), weight=c(1,1,1,1,3))

g <- graph_from_data_frame(df)
g
#> IGRAPH 677ae70 DNW- 5 5 -- 
#> + attr: name (v/c), weight (e/n)
#> + edges from 677ae70 (vertex names):
#> [1] 1->2 2->1 3->3 4->5 1->2
vertex.attributes(g)
#> $name
#> [1] "1" "2" "3" "4" "5"

g <- graph_from_data_frame(df, directed = FALSE)
g
#> IGRAPH 677bf7b UNW- 5 5 -- 
#> + attr: name (v/c), weight (e/n)
#> + edges from 677bf7b (vertex names):
#> [1] 1--2 1--2 3--3 4--5 1--2
edge.attributes(g)
#> $weight
#> [1] 1 1 1 1 3

h <- simplify(g, remove.multiple = FALSE, edge.attr.comb = "sum")
h
#> IGRAPH 677ce17 UNW- 5 4 -- 
#> + attr: name (v/c), weight (e/n)
#> + edges from 677ce17 (vertex names):
#> [1] 1--2 1--2 4--5 1--2
edge.attributes(h)
#> $weight
#> [1] 1 1 1 3

h <- simplify(g, edge.attr.comb = "sum")
h
#> IGRAPH 677d7d4 UNW- 5 2 -- 
#> + attr: name (v/c), weight (e/n)
#> + edges from 677d7d4 (vertex names):
#> [1] 1--2 4--5
edge.attributes(h)
#> $weight
#> [1] 5 1
```

<sup>Created on 2020-05-23 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

[...]

PR #395, opened 2020-05-24
Graph imported from data.frame is not simplified

SamGG:
As discussed in #394, I would like to document that stating the graph as undirected does not imply that reversed links are aggregated automagically. I think that simplication could speed up some calculations. Let me know if you feel this addition informative.
Best.

### DIFF
--- a/R/data_frame.R
+++ b/R/data_frame.R
@@ -49,7 +49,7 @@
 #'
 #' Regardless of the \code{directed} status of the graph, the graph is not
 #' simplified. To remove the loops and/or multiple edges from a graph, use
-#' \code{simplify}. Simplifying the graph can speed up some calculations.
+#' \code{simplify}.
 #'
 #' \code{as_data_frame} converts the igraph graph into one or more data
 #' frames, depending on the \code{what} argument.
--- a/R/data_frame.R
+++ b/R/data_frame.R
@@ -47,6 +47,10 @@
 #' Excel and are imported into R via \code{\link{read.table}},
 #' \code{\link{read.delim}} or \code{\link{read.csv}}.
 #'
+#' Regardless of the \code{directed} status of the graph, the graph is not
+#' simplified. To remove the loops and/or multiple edges from a graph, use
+#' \code{simplify}. Simplifying the graph can speed up some calculations.
+#'
 #' \code{as_data_frame} converts the igraph graph into one or more data
 #' frames, depending on the \code{what} argument.
 #'


szhorvat:
Just a note that the purpose of simplification is *not* to speed up calculations.

When you are computing some graph measure, you need to consider:

 - Does it make sense at all for multigraphs?
 - If it does make sense, does igraph's implementation work for multigraphs?
 - If yes, does it make sense for your application (interpretation of the graph) to retain multi-edges? What is their interpretation? Retaining them does affect many graph measures, such as betweenness centrality.

I am not sure how @gaborcsardi feels about it, but personally I would remove "Simplifying the graph can speed up some calculations." because it might mislead some users.

SamGG:
Being not experienced with graphs or multigraphs, I removed the speed up sentence.
Thanks for your feedback.

SamGG:
Just for information, for Louvain algorithm, the simplified weighted graph leads to the same result than the original multigraph, with computation of 76 sec (including the simplification, 300 000 vertices, 6164839 edges) vs 102 sec (unsimplified, 8 646 579 edges).

mbojan:
I am pretty sure that Louvain alg. ignores multiple edges (unless you "aggregate" the graph by counting edges within each dyad and use the counts as weights) and the slowdown is just because your handling a larger object memory-wise.

As we discussed in #394, and as @szhorvat writes above, simplifying the graph is not a pure efficiency matter. It is not appropriate to assume in general that if there are more than one edge within a dyad that all the extra edges are redundant from the data perspective. One popular usecase is social network analysis: you can have multiple edges between two nodes (persons) corresponding to different types of relations (say i is a friend of j, j is a boss of i, i often seeks advice from j and so on), which you would like to be able to analyze/visualize jointly

IMHO from the software perspective it is better not to assume what particular application is the data structure used for. It has always been in R tradition to empower the user with the responsibility of being explicit how the data should be processed (just like `NA`s require explicit treatment even if you want to compute a simple `mean()`).

vtraag:
> I am pretty sure that Louvain alg. ignores multiple edges

The implementation actually loops over (incident) edges, so it does take into account multiple edges. Similarly, `igraph_modularity` loops over edges, so also considers multiple edges. FInally, `igraph_modularity_matrix` relies on `get_adjacency` which also considers multiple edges and entries will contain the edge multiplicities. Perhaps this can be explicitly stated in the documentation.

Regardless, as you and @szhorvat already stated
> simplifying the graph is not a pure efficiency matter. 
I agree completely.

vtraag:
The original branch by @SamGG is deleted. It might still be useful to document explicitly that all edges are considered. I added this in 6524d0239432199a7b9fa819cc3b7cacc6d05a67.

PR #429, opened 2020-11-04
Fix warning due to factors in graph_from_data_frame

IgnatiusPang:
When converting a data.frame with an POSIXct column to an igraph object
using the function 'graph_from_data_frame', a warning is issued due to
ineffective checking of classes:
> Warning message:
> In if (class(newval) == "factor") { :
>   the condition has length > 1 and only the first element will be used

Th warning is true as a POSIXct object has actually two classes:
"POSIXct" and "POSIXt".

To remove the warning, the class of edge and vertex attributes is now
checked using the infix notation "%in%".

Signed-off-by: Claus Hunsen <hunsen@fim.uni-passau.de> copied here by Ignatius Pang <ignatius.pang@gmail.com?

### DIFF
--- a/R/data_frame.R
+++ b/R/data_frame.R
@@ -156,7 +156,7 @@ graph_from_data_frame <- function(d, directed=TRUE, vertices=NULL) {
     if (ncol(vertices) > 1) {
       for (i in 2:ncol(vertices)) {
         newval <- vertices[,i]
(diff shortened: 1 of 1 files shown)


vtraag:
I am closing this PR in favour of the original PR #250.

PR #433, opened 2020-11-25
Update C core to 0.8.5

vtraag:
This PR updates the C core to the recently released version 0.8.4. In addition, it does the following:
- Automatically builds a correct (development) version number based on tags.
- Removes the `AMD`, `COLAMD` and `CHOLMOD` directories, since they are no longer included in C core version 0.8.4.
- Corrects the documentation for `cluster_leiden`, which had a missing argument
- The `digest` package was included twice in `Suggests`, the duplicate is now removed.
- Made sure all warnings are treated as errors on the CI.

Unfortunately, there is one error for the Kamada-Kawai layout on my local machine:
```
── 1. Failure: Kamada-Kawai layout generator works (@test_layout.kk.R#26)  ─────
`x` not equal to `expected`.
1/1 mismatches
[1] -84.8 - -86.1 == 1.32
```
Perhaps this is something that is specific to my machine, let's see how this does on the CI .


vtraag:
I am merging this PR. Everything builds correctly now on Windows, macOS and linux. The code should be automatically deployed to the `master` branch, in order to support installation using `remotes::install_github("igraph/rigraph@master")`. At the moment, this does not work, but this is presumably just because there are no appropriate permissions when being run from a PR.

PR #435, opened 2020-11-25
Update CONTRIBUTING.md

IgnatiusPang:


gaborcsardi:
Why?

vtraag:
@IgnatiusPang, I don't think that adding the explicit `devtools` namespace clarifies much. The `devtools` is also already loaded, so it is also not needed. I am going to close this now. If you do have specific argument for why this should be clarified, let me know, and we can consider it again!

PR #441, opened 2021-01-13
 Switch CI from Travis to GitHub actions (continued)

vtraag:
This continue PR #440. For some reason the GitHub action did not get triggered in the PR, but did get triggered once included in the `dev` branch. This PR now continues to correct the GitHub action.


vtraag:
CI does not get triggered, but it is not clear why. I will merge again to the `dev` branch, that will hopefully trigger it. 

There were still some problems with some of the checks, they are now corrected. That is, the line endings were changed for Pajek files (in https://github.com/igraph/igraph/commit/4d52c2e3d7732bf7d0c98a7858332a9d7e53d5b6), that test is now corrected. It seems that the Kamada-Kawai layout are not really stable, so we should skip the test also on CI (it is already skipped on CRAN).

vtraag:
Apparently I made a mistake in the `.yml` file. For some reason this error does not show up when run from a PR, but it does show up from the `dev` branch under the "Actions" tab.

PR #455, opened 2021-03-05
Raise warning if membership is not numerical.

vtraag:
This addresses the problem that was identified in PR #307. However, instead of interpreting any vector as a suitable membership vector, this PR simply raises a warning if the vector is not numerical.

It would be nice to have "named" communities, so that arbitrary vectors would also be suitable as membership, and that these names would also be consistently used throughout the code base. However, this is a much more extensive change. By allowing this in only a single function (`modularity`), but not elsewhere, this will presumably raise questions. This way, the input is expected to be consistent for all functions, namely a numerical vector.

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -424,6 +424,10 @@ modularity <- function(x, ...)
 modularity.igraph <- function(x, membership, weights=NULL, ...) {
   # Argument checks
   if (!is_igraph(x)) { stop("Not a graph object") }
+  if (is.null(membership) || !is.numeric(membership))
+  {
+    stop("Membership is not a numerical vector")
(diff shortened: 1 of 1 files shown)


ntamas:
This PR breaks one of the test cases in `bluster` where a factor is passed to `modularity()` instead of a numeric vector. Since factors in R are essentially numeric vectors with an extra `$levels` attribute to store the list of _possible_ values and a `$class` attribute to store that the vector is in fact a factor, I believe there is no harm in allowing factors to be passed in to `modularity()`. I will commit a patch soon that relaxes the check in this function to allow factors.

PR #459, opened 2021-03-17
Fix minor typo in documentation

zeehio:
operatation -> operation

Since this is a typo correction I don't think this patch is copyrightable. But just in case: I'm fine with GPL 2 or later and FDL.


gaborcsardi:
Thanks!

Issue #312, opened 2019-01-14
cluster_label_prop() with initial labels


briqqo:
Hello,

The function cluster_label_prop() lets R crash when using an initial label (ok without initial state): I have a graph g of twitter retweets and 

V(g)$label = -1
V(g)[name=="XXX"] = 1
V(g)[name=="YYY"] = 2 etc.
cluster_label_prop(as.undirected(g), initial = V(g)$label) # -> crash

This does not depend on the size of g.

briqqo:
The crash seems to come from the non-connectivity of the graph. Reducing the algorithm to the main component works fine.

erikvona:
Experienced the same error, it would be nice if we could have an `stopifnot("Graph must be fully connected to use an initial state" = is_connected(graph) | is.null(initial))` if this is a limitation in the algorithm

szhorvat:
@erikvona I cannot reproduce this problem. If you can, can you please post a minimal working example? Ideally something small enough so that you can create the graph with `graph_from_literal`.

erikvona:
I've done some additional testing, it happens when there's an isolated subgraph where all nodes have blank initial labels. This can be a single isolated vertex.

Note that the following example will crash R:

``` r
library(igraph)
g <- sample_gnp(10, 5/10) %du% sample_gnp(9, 5/9)
cluster_label_prop(g, initial = c(sample(1:3, 10, replace = T), rep(-1, 9)))
```

Or a single isolated vertex:

``` r
library(igraph)
g <- sample_gnp(10, 5/10) %du% sample_gnp(9, 5/9)
g <- add_edges(g, c(1, 12))
cluster_label_prop(add_vertices(g, 1), initial = c(sample(1:3, 19, T), -1))
```

I think the desired behavior would be to either throw an error, or identify those isolated subgraphs and replace the undefined labels with random unused ones so the isolated subgraphs are clustered separately.

szhorvat:
Thanks.

Here's a simpler deterministic example:

```
library(igraph)
g <- graph_from_literal(1-2,3)
cluster_label_prop(g, initial = c(1,1,-1))`
```

PR #492, opened 2021-11-20
Avoid crash in cluster_label_prop

erikvona:
If cluster_label_prop is ran with an initial state and the graph is not fully connected, R crashes (#312) . Added code to throw an error in that case to avoid crashing R (a fix in the C code would be nicer but I'm unable to do that and the issue has been open for 2 years).

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -1586,6 +1586,7 @@ cluster_label_prop <- function(graph, weights=NULL, initial=NULL,
                                         fixed=NULL) {
   # Argument checks
   if (!is_igraph(graph)) { stop("Not a graph object") }
+  stopifnot("Graph must be fully connected to use an initial state" = is_connected(graph) | is.null(initial))
   if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
   weights <- E(graph)$weight
   }


szhorvat:
I do not believe that this is a limitation of the algorithm. If it indeed is, it should be handled in the C core, not the R interface. Thanks for bringing this up, we will look into it.

szhorvat:
Thanks again for bringing this up. The problem appears to be in the C core of igraph, and it will be fixed there. Therefore I will close this PR.

ntamas:
I'm inclined to keep this as a stopgap measure for 1.2.9 or so (which I will submit to CRAN in a few days anyway to solve  a build issue on Solaris and another one on Apple Silicon), and then remove it once we upgrade the C core in the R interface to 0.9.x. Any objections?

ntamas:
Hmmm, actually, this may cause breakage in reverse dependencies if someone happens to use label propagation in a downstream project on a graph that is _not_ connected but has no unlabeled vertices (or that has unlabeled vertices if they are replaced with proper labels in the course of the algorithm). So, I cannot merge this as-is without potentially breaking revdeps. Let's just wait until R-igraph 1.3 to fix this.

PR #521, opened 2022-04-05
Add forum URL to DESCRIPTION

szhorvat:


ntamas (approved):

szhorvat:
The intention is to keep users off GitHub and channel them into the forum. I did not find another suitable field. There's `MailingList`, but this is a forum, not a mailing list. Other packages which have support forums seem to use the URL field.

PR #535, opened 2022-05-12
remove @importFrom graphics plot

MichaelChirico:
`graphics::plot` is now a re-export of `base::plot`; this may be removed in the future.

### DIFF
--- a/R/cohesive.blocks.R
+++ b/R/cohesive.blocks.R
@@ -365,7 +365,6 @@ summary.cohesiveBlocks <- function(object, ...) {
 #' @method plot cohesiveBlocks
 #' @export
 #' @importFrom grDevices rainbow
-#' @importFrom graphics plot
  
 plot.cohesiveBlocks <- function(x, y,
                                 colbar=rainbow(max(cohesion(x))+1),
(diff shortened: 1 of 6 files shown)


MichaelChirico:
Hmm on second thought, doing so introduces a build-time dependency on R>=4.0.0. R<4 users would still be able to install the pre-compiled package but I'm not sure it would work...

ntamas:
Actually we haven't really thought about policies to support older R versions yet; igraph happened to work on earlier versions and we test it all the way back to R 3.4 in CI, but it doesn't mean that we want to continue supporting 3.4 forever.

`igraph` is used in the `tidygraph` package, which is part of the Tidyverse, so I guess we probably shouldn't deviate too far from what Tidyverse supports. Right now it is not a significant effort on our end to keep things working on older R versions, so it's unlikely that we will drop support for 3.4 and 3.5 in the near term, but sooner or later it will happen, and this PR serves as a reminder for the necessary adjustments.

MichaelChirico:
yep! feel free to refer back. but I certainly wouldn't bump the dependency requirement for something so trivial unless R itself drops plot from graphics

PR #551, opened 2022-07-17
Let layout_nicely() accept negative weights

szhorvat:
`layout_nicely()` is the default layout for `plot.igraph`. In most cases, this function calls `layout_with_fr()` for small graphs and `layout_with_drl()` for large ones. Neither of these two support non-positive weights, but they did not validate the weight vector before igraph 1.3.3. Now that they do, `plot(graph)` will fail on graphs with non-positive weights.

This PR makes it so:

 - If some weights are not positive, `layout_nicely()` ignores weights with a warning. Thus when using `plot` with default settings, it will not fail.
 - `layout_with_fr()` and `layout_with_drl()` continue to throw errors on non-positive weights. When they are used directly and explicitly (instead of through `layout_nicely`), non-positive weights will not be accepted.
 - It is not documented that weights must be positive for FR and KK. DrL already had this documented.

### DIFF
--- a/R/layout.R
+++ b/R/layout.R
@@ -543,6 +543,9 @@ layout.circle <- function(..., params = list()) {
 #' @aliases layout.auto
 #' @param graph The input graph
 #' @param dim Dimensions, should be 2 or 3.
+#' @param weights A vector giving edge weights. The \code{weight} edge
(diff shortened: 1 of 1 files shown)


ntamas (approved):

ntamas:
Thanks!

PR #576, opened 2022-09-23
Use testthat edition 3

krlmlr:
for parallel testing.

Starts with consistent code formatting (removal of spaces at EOF and lines at EOL), a prerequisite for the following bulk search+replace operations.

Adapts the test code to be compatible with testthat edition 3, which is required for parallel testing. This reduces the time to test the entire package to a half (4 cores, 32s -> ~16s).

Finally, `library()` shouldn't be used in test code, and is not necessary.

A lot of files are touched, each commit addresses a particular issue and can be reviewed individually. I can submit smaller PRs if needed.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -27,3 +27,5 @@
 ^cran-comments\.template.md$
 ^CRAN-RELEASE$
 ^CRAN-SUBMISSION$
+^igraph\.Rproj$
+^\.Rproj\.user$
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@
 igraph.Rcheck/
 .venv/
 .vscode/
+.Rproj.user
--- a/igraph.Rproj
+++ b/igraph.Rproj
@@ -1,0 +1,22 @@
+Version: 1.0
+
+RestoreWorkspace: No
+SaveWorkspace: No
+AlwaysSaveHistory: Default
+
+EnableCodeIndexing: Yes
+UseSpacesForTab: Yes
+NumSpacesForTab: 2
+Encoding: UTF-8
+
+RnwWeave: Sweave
+LaTeX: pdfLaTeX
+
+AutoAppendNewline: Yes
+StripTrailingWhitespace: Yes
+LineEndingConversion: Posix
+
+BuildType: Package
+PackageUseDevtools: Yes
+PackageInstallArgs: --no-multiarch --with-keep.source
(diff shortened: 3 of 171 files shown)


ntamas:
@krlmlr I have merged #577 and now there are several files that are in conflict with `dev`. I tried resolving these with `git merge -Xours dev` but I cannot push new commits to this PR. Can you please either allow us to push commits to this PR or resolve the conflicts so I can proceed with the review?

ntamas (changes requested):
Thanks a lot, everything looks good apart from the two lines that I pointed out. If you manage to fix the merge conflicts and these two issues then I think this can safely be merged.

Re whitespace: should we add a pre-commit hook that ensures that there are no trailing whitespace anywhere? In Python projects I would use https://pre-commit.com/ but I don't know whether there's a similar tool for R projects.

krlmlr:
Thanks, good catch. Resolved. I wonder why my local test runs didn't catch the problem.

ntamas:
The CI failure is weird, I cannot reproduce it locally on my Mac. I'm gonna re-run the failed job and see what happens.

krlmlr:
I’ll take a look too.

ntamas:
All checks passed so I'm going to merge this, thanks!

krlmlr:
Thanks. Some notes:

- Confirming that the `Matrix::sparseMatrix::Matrix` error leads to a parse error that is silently skipped when running tests in parallel. This is the reason why I didn't see the error, but it's great these errors are caught in CI/CD.
- With parallel testing, I'm now seeing sporadic test failures and warnings:
    - in `test-vs-es:230`, almost always
    - in `test_leading.eigenvector.community.R:65:5`, warning or error, rarely: ARPACK solver failed to converge (1001 iterations, 0/1 eigenvectors converged). Can be reproduced independently by running `testthat::test_local("leading.eigenvector.community")` repeatedly
    
    Submitted PRs.

ntamas:
When you mention "parallel testing", does it mean that `testthat` is running multiple _threads_ in the same process, and those threads are calling igraph functions concurrently, or does it use multiple _processes_? Is there anything special I need to do to turn on parallel testing on my machine or is it now enabled by default?

krlmlr:
Tests are run in parallel processes, as R is single-threaded. It is enabled by a setting in `DESCRIPTION` that starts with `Config/testthat/...`. Control the number of processes with the `TESTTHAT_CPUS` environment variable.

Issue #328, opened 2019-03-22
arpack should provide a way to be deterministic


flying-sheep:
In order to get a deterministic result, one either has to set a seed in FORTRAN or provide a starting vector (v0). I assume that would be the `start` parameter. Currently `start` is not implemented though:

https://github.com/igraph/rigraph/blob/5e7a689afde1864c2f36b2a7825db24f2f89b22b/R/centrality.R#L117-L119

I also have no idea how to set a seed in FORTRAN…

There’s some discussion in scipy/scipy#9023

PR #579, opened 2022-09-24
Failure with stochastic test

krlmlr:
Noticed with parallel testing. Should we, for now, set a seed where this failure doesn't occur? Seed 1 seems fine.

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -295,7 +295,7 @@ estimate_closeness <- function(graph, vids=V(graph), mode=c("out", "in", "all", 
 #' @export
 
 arpack_defaults <- list(bmat="I", n=0, which="XX", nev=1, tol=0.0,
-                              ncv=3, ldv=0, ishift=1, maxiter=1000, nb=1,
+                              ncv=3, ldv=0, ishift=1, maxiter=3000, nb=1,
                               mode=1, start=0, sigma=0.0, sigmai=0.0)
 
 #' ARPACK eigenvector calculation
--- a/tests/testthat/test_leading.eigenvector.community.R
+++ b/tests/testthat/test_leading.eigenvector.community.R
@@ -5,7 +5,6 @@ test_that("cluster_leading_eigen works", {
   ## Check-test
 
   set.seed(18)
-  arpack_defaults$maxiter <- 3000
 
   f <- function(membership, community, value, vector, multiplier, extra) {
     M <- sapply(1:length(vector), function(x) {
--- a/tests/testthat/test_leading.eigenvector.community.R
+++ b/tests/testthat/test_leading.eigenvector.community.R
@@ -5,6 +5,7 @@ test_that("cluster_leading_eigen works", {
   ## Check-test
 
   set.seed(18)
+  arpack_defaults$maxiter <- 3000
 
   f <- function(membership, community, value, vector, multiplier, extra) {
     M <- sapply(1:length(vector), function(x) {
--- a/tests/testthat/test_leading.eigenvector.community.R
+++ b/tests/testthat/test_leading.eigenvector.community.R
@@ -4,6 +4,8 @@ test_that("cluster_leading_eigen works", {
 
   ## Check-test
 
+  set.seed(18)
+
   f <- function(membership, community, value, vector, multiplier, extra) {
     M <- sapply(1:length(vector), function(x) {
       v <- rep(0, length(vector))


szhorvat:
Can you please write what the failure is (test code, expected output, actual output) before going ahead with changing the test?

szhorvat:
I think I misunderstood. This "PR" is not a proposed change, but a demonstration of how setting a certain seed triggers a test failure. Is that correct?

szhorvat:
OK, so let's be complete an write down here that the error message is:

```
At core/linalg/arpack.c:803 : ARPACK solver failed to converge (1001 iterations, 0/1 eigenvectors converged).
```

I am not sure where this number of 1000 is coming from. In the C core the default number of iterations is 3000. The R interface is overriding this somewhere. Doing more iterations may resolve #581.

szhorvat:
I'm not going to mess with this any more because changing the default options does not seem to be taking effect and I don't know enough R to understand why. Also, there's an issue with downloading pandoc that causes CI builds to fail.

My point is that the default number of iterations should be consistent between the C core and the R interface. The former has 3000, the latter 1000. Are these numbers well founded? I don't know, possibly not. However, based on my experimentation with the Mathematica interface, increasing iterations is likely to fix #581, but not this issue.

The R interface uses the bundled ARPACK which is very outdated, and does not work as reliably as a recent ARPACK-NG version. One peculiarity is that R/igraph and C/igraph bundle different instances of ARPACK: the former uses the original Fortran code, the latter an f2c'd version. @ntamas : At some point you updated BLAS/LAPACK/ARPACK to the latest versions which could still be translated with f2c. Did the C core get out of sync with the R interface at that time, did it stay on an old version? Even if not, would it be a good idea to update to an even newer version, given that f2c is not used in R/igraph?

ntamas:
> At some point you updated BLAS/LAPACK/ARPACK to the latest versions which could still be translated with f2c. Did the C core get out of sync with the R interface at that time,

It never was in sync. The R interface has always been linked to whatever version of BLAS and LAPACK was shipped with R (one can safely assume that BLAS and LAPACK is always present). Plus, since we can also reasonably assume that the compilation environment of the R interface has a Fortran compiler, we just use the Fortran code of ARPACK instead of an `f2c`-translated version. One of the obstacles to upgrading ARPACK in the C core to a newer version is that `f2c` does not support Fortran 90, and newer versions of ARPACK use Fortran 90 constructs that `f2c` is not comfortable with. But I think it's unreasonable to expect that the version numbers of BLAS / LAPACK / ARPACK could be kept in sync between the C core and the R interface anyway.

One thing that could _possibly_ be done is to update the vendored ARPACK in the R interface to a newer version, assuming that the Fortran compiler in the standard CRAN environment can handle newer ARPACK versions. But I have no idea whether it would solve the issue we have at hand here.

szhorvat:
I'm talking specifically about ARPACK, not BLAS/LAPACK. They are different things that shouldn't be mixed up, even though they're named similarly.

szhorvat:
> But I have no idea whether it would solve the issue we have at hand here.

I don't know about _this_ issue, but there is a substantial difference in the quality of the ARPACK bundled with the C core and the latest ARPACK-NG. Examples of that include:

 - https://github.com/igraph/igraph/issues/1990
 - https://github.com/igraph/igraph/issues/1641

ntamas:
> What I was asking was whether the ARPACK in R/igraph is older than in C/igraph. It really shouldn't be.

Okay, so what should be the course of action now that the original (Fortran) ARPACK homepage is essentially gone? We have ARPACK-NG, it's written in Fortran, but in theory we could vendor that in R-igraph; but then we are out of sync with the C core, which cannot vendor ARPACK-NG due to `f2c`'s inability to convert the source to C. (Or, we would need to add a Fortran compiler as a dependency for igraph, which I'd rather avoid).

IMHO being "in sync" with the C core is not that important; I'm totally fine with vendoring a recent ARPACK-NG in R-igraph and keeping an older version in the C core out of necessity.

szhorvat:
> IMHO being "in sync" with the C core is not that important; I'm totally fine with vendoring a recent ARPACK-NG in R-igraph and keeping an older version in the C core out of necessity.

Agreed. What I meant was not to have an older one in R, as we know that old ones don't perform as well.

Here's the ARPACK page from the Internet Archive:

https://web.archive.org/web/20220619221744/https://www.caam.rice.edu/software/ARPACK/

It was possible to download the sources. All files are dated 2002 or earlier, so the version in R/igrah can't be older than the one in C/igraph, assuming that C/igraph isn't using ARPACK-NG.

szhorvat:
Updating to ARPACK-NG would be nice, but IMO not urgent. Let's first try to understand better what is going on here. After increasing max iterations to 3000, to match the C core, this problem went away.

But the other one, #581, didn't. More concerningly, #581 does not appear to be deterministic, despite setting a seed.

So, I propose:

 - Increase maxiter to 3000 in R/igraph to match the C core
 - Investigate the non-determinism in #581. ARPACK may be using a different RNG internally to initialize state?

szhorvat:
See also #328

ntamas:
Increasing the iteration count seems to have resolved this problem so I'm closing the PR. Feel free to reopen if the issue resurfaces.

PR #581, opened 2022-09-24
More serious (?) failure with stochastic test

krlmlr:
Similar to #579, but here an actual expectation fails. This seems to require a more thorough investigation.

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -295,7 +295,7 @@ estimate_closeness <- function(graph, vids=V(graph), mode=c("out", "in", "all", 
 #' @export
 
 arpack_defaults <- list(bmat="I", n=0, which="XX", nev=1, tol=0.0,
-                              ncv=3, ldv=0, ishift=1, maxiter=1000, nb=1,
+                              ncv=3, ldv=0, ishift=1, maxiter=3000, nb=1,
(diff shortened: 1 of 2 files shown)


ntamas:
Linking igraph/igraph#2207 here as a possible source of nondeterminism

szhorvat:
Just noting that the non-determinism (i.e. non-responsiveness to a seed set in R) was just one of the issues here. The main issue is of course that running this community-detection function twice in a row may produce different results. That calls into question the accuracy of the produced eigenvectors.

ntamas:
I have backported the recent determinism-related ARPACK patches in igraph 0.10.x to the 0.9 branch and this failure now seems to be gone, at least on my machine. I'll commit this and merge the change proposed by @szhorvat re the iteration count and then we'll see how this fares in CI.

ntamas:
CI passed with an updated C igraph core so I'm going to close this PR. Feel free to reopen if you find another issue similar to this.

szhorvat:
Does the CI pass with _this_ modification though? WHat if you merge `dev` into this PR?

ntamas:
I can't push commits to this PR, but it passes the tests locally on my computer, even with the modifications outlined here.

szhorvat:
> I can't push commits to this PR

It's strange that it allowed me to edit from the web interface though.

PR #593, opened 2022-12-14
chore: Remove or replace `\dontrun{}`

krlmlr:
to see how coverage changes.


krlmlr:
Coverage does increase, but just by a tiny bit, and for files that are already covered well enough.

ntamas (approved):

PR #600, opened 2022-12-15
fix: Avoid defining PACKAGE_VERSION at all in Makevars

krlmlr:
It works locally, trying out in CI/CD now. Maybe this will fix r-universe?


krlmlr:
Changing the behavior here. Do we need `PACKAGE_VERSION` elsewhere in the C code?

ntamas:
> Do we need PACKAGE_VERSION elsewhere in the C code?

`PACKAGE_VERSION` is used by `include/igraph_version.h`, and the constants defined there are used by the C-level `igraph_version()` function call. When compiling the C core directly with CMake, there's some magic code in CMake that determines the appropriate version number from the closest git tag to the revision being compiled. I did not want to replicate that in `autoconf`, so right now there's a section in the Makefile that dynamically rewrites `igraph_version.h` with `awk` and `sed` so it includes the same version number as the one used by the R package.

I think that as long as we don't use the `igraph_version()` function from R any more, the current solution is okay. Note the contents of the file named `src/include/igraph_version.h`:

```
#define IGRAPH_VERSION __IGRAPH_QUOTE2(PACKAGE_VERSION)
#define IGRAPH_VERSION_MAJOR @PACKAGE_VERSION_MAJOR@
#define IGRAPH_VERSION_MINOR @PACKAGE_VERSION_MINOR@
#define IGRAPH_VERSION_PATCH @PACKAGE_VERSION_PATCH@
#define IGRAPH_VERSION_PRERELEASE "@PACKAGE_VERSION_PRERELEASE@"
```

which basically means that we need to avoid code that attempts to use these macros; other than that it should be fine.

Issue #466, opened 2021-06-10
Overwriting attributes converts the new value to the class of the old value


arcresu:
When setting an attribute that already exists (using `V()<-` or `set_vertex_attr()` and presumably also with edge attributes), igraph (1.2.6) will convert the new value to match the class of any existing value. For example:

``` r
library(igraph)
#> 
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#> 
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#> 
#>     union

# does some processing and stores the result in a new integer attribute 'kind'
set_kind <- function(gr) {
  # gr <- delete_vertex_attr(gr, "kind")  # workaround
  set_vertex_attr(gr, "kind", value = vcount(gr))
}

# (1) processing a clean graph works as expected
g <- make_full_graph(n=1)
g <- set_kind(g)
class(V(g)$kind)
#> [1] "numeric"

# (2) this graph happens to have an existing attribute 'kind' with a different class;
#     the result now has the wrong class
g <- make_full_graph(n=1) %>% set_vertex_attr("kind", value = "big")
g <- set_kind(g)
class(V(g)$kind)
#> [1] "character"
```

<sup>Created on 2021-06-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

Basically I would expect the parts marked (1) and (2) to do the same thing, but the result actually depends on whatever value the attribute had before. This makes `V() <-` inconsistent with the normal R semantics of the `<-` operator which allow you to replace a variable with a value of a different class just fine.

To make the processing code robust against arbitrary input graphs, it's necessary to explicitly clear every attribute before setting it, but this becomes cumbersome.

ntamas:
FWIW it looks like `character` cannot be converted to `integer` simply by doing an assignment, but it works the other way round:

```r
# (1) converting a character attribute to integer doesn't work
> g <- make_full_graph(n=1) %>% set_vertex_attr("kind", value = "big")
> class(V(g)$kind)
#> [1] "character"
> V(g)$kind <- vcount(g)
> class(V(g)$kind)
#> [1] "character"
> V(g)$kind
#> [1] "1"

# (2) but converting an integer to a character does
> g <- make_full_graph(n=1) %>% set_vertex_attr("kind", value = vcount(g))
> class(V(g)$kind)
#> [1] "integer"
> V(g)$kind <- c("big")
> class(V(g)$kind)
#> [1] "character"
```

[...]

PR #601, opened 2022-12-15
chore: Iterators over all vertices and edges gain an attribute

krlmlr:
For #466.

### DIFF
--- a/R/iterators.R
+++ b/R/iterators.R
@@ -149,6 +149,15 @@ graph_id.igraph.es <- function(x, ...) {
   get_es_graph_id(x) %||% NA_character_
 }
 
+is_complete_iterator <- function(x) {
+  identical(attr(x, "is_all"), TRUE)
+}
+
+set_complete_iterator <- function(x, value = TRUE) {
+  attr(x, "is_all") <- TRUE
+  x
+}
+
 #' Vertices of a graph
 #'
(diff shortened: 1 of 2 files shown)


ntamas (approved):

szhorvat:
Is it possible to export edge and vertex selectors, e.g. when saving the current workspace? If yes, will this break that? What if someone saved some edge/vertex selectors in a previous version, but when re-loading that data, they try to use it with a new igraph version?

As I understand, there is versioning for igraph object. Do we want some sort of versioning (and auto-updating) for other kinds of objects that igraph produces?

ntamas:
I think the current implementation is backwards compatible. It adds an `is_all` attribute to newly created vertex and edge sequence objects, but it only uses `attr(x, "is_all")` to retrieve it, and compares the result with `TRUE`. If `is_all` is not present (for already-serialized vs and es objects), the `attr()` call will return `NULL`, which is not identical to `TRUE` so we are on the safe side.

The attribute is needed only to recognize the situation when the user tries to do something like `V(g)$name <- something`. The goal is  that in this case, if we know that `V(g)` contains all vertices (i.e. it was never subsetted), we should allow changing the type of the attribute during the assignment (will be implemented in a separate PR).

krlmlr:
Thanks for the explanation, this fell through the cracks. I second that.

PR #604, opened 2022-12-15
test: Replace expect_output() with snapshot tests

krlmlr:
These are easier to understand and debug if they fail.


ntamas:
I wonder whether we could simply forcibly set the ID of the graph to the value expected by the snapshot so we could make this deterministic.

krlmlr:
Closing in favor of #605.

PR #605, opened 2022-12-15
feat: New id argument to print(), with new option print.id

krlmlr:
With stable snapshot tests, for #466.

### DIFF
--- a/R/par.R
+++ b/R/par.R
@@ -159,6 +159,10 @@ igraph.pars.callbacks <- list("verbose"=igraph.pars.set.verbose)
 #' @importFrom pkgconfig set_config_in get_config
 
 igraph_options <- function(...) {
+  igraph_i_options(...)
+}
+
+igraph_i_options <- function(..., .in = parent.frame()) {
   if (nargs() == 0) return(get_all_options())
 
   ## Short notation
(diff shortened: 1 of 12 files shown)


krlmlr:
There's light at the end of the rabbit hole.

This is a rather large change, but the commits make sense individually. I need this to troubleshoot failures in the indexing code.

R/par.R
```diff
@@ -186,10 +190,17 @@ igraph_options <- function(...) {
...
   ## Set them
   names(temp) <- paste0("igraph::", names(temp))
-  do.call(set_config_in, c(temp, list(.in = parent.frame())))
+  do.call(set_config_in, c(temp, list(.in = .in)))
   invisible(old)
 }
 
+local_igraph_options <- function(..., .in = parent.frame()) {
+  old <- igraph_options(..., .in = .in)
+  print(old)
```
ntamas, line 199: Is this `print()` statement necessary here or is it only for debugging purposes?

krlmlr: Good catch, removed.

R/par.R
```diff
@@ -185,10 +190,16 @@ igraph_options <- function(...) {
...
   ## Set them
   names(temp) <- paste0("igraph::", names(temp))
-  do.call(set_config_in, c(temp, list(.in = parent.frame())))
+  do.call(set_config_in, c(temp, list(.in = .in)))
   invisible(old)
 }
 
+local_igraph_options <- function(..., .in = parent.frame()) {
+  old <- igraph_options(..., .in = .in)
+  withr::defer(rlang::inject(igraph_options(!!!old)), envir = .in)
```
ntamas, line 199: I guess that we need to move `withr` from the `Suggests:` section of the `DESCRIPTION` file to the `Imports:` section, right?

krlmlr: Not necessarily. This isn't part of an exported API, and we can realistically expect that it's available during testing.

ntamas:
Merging PR upon CI pass, FYI.

PR #606, opened 2022-12-15
chore: Add new docs target to Makefile

krlmlr:
so that we can run `make docs` before committing, e.g., in a pre-commit hook.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ build: pre_build
 	R CMD INSTALL -l "/tmp" --no-configure --no-byte-compile .
 
 docs: pre_build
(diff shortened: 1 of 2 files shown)


ntamas:
I guess that the change to the `man/` folder is unrelated -- shall I merge it nevertheless?

krlmlr:
The `man/` change is intended, it should have been part of https://github.com/igraph/rigraph/commit/bc2414b6f45bef0622f75860ee66bf8dcdc260a0.

krlmlr:
`make docs` could be used in a pre-commit hook.

PR #608, opened 2022-12-15
feat: Support `index = NULL` when setting attributes

krlmlr:
For #466. Two FIXMEs remain, to be addressed in a follow-up PR.

I'd like to propose to make attribute updates a bit stricter as part of this exercise: allow `value` only of length one or the same length as the index. While this is technically a breaking change, most uses of passing `value` of incompatible length are very likely a mistake. @ntamas: What do you think?

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -148,8 +148,9 @@ graph.attributes <- function(graph) {
 #' @param graph The graph.
 #' @param name Name of the attribute to query. If missing, then
 #'   all vertex attributes are returned in a list.
-#' @param index A vertex sequence, to query the attribute only
+#' @param index An optional vertex sequence to query the attribute only
 #'   for these vertices.
+#'   If `NULL`, the default, the attribute is queried for all vertices.
 #' @return The value of the vertex attribute, or the list of
 #'   all vertex attributes, if \code{name} is missing.
 #'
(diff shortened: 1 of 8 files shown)


ntamas:
Not sure; I can easily imagine people making use of the vector recycling feature of R to assign "perodic" vertex or edge attributes to graphs with a certain regular structure.

krlmlr:
I did think about that, but I find it rather exotic, to be honest. Those users, if any, would need to use `rep()` beforehand. Can we get data from our users to drive this decision?

krlmlr:
It doesn't complicate things, but I believe it's the right thing to do.

I have a few ideas how to make revdepchecks a less costly exercise.

ntamas:
Any improvements to the revdepcheck process are welcome. Right now the biggest pain for me is that the revdepcheck process has to be babysitted because the revdepcheck process crashes randomly (with a segfault). Sometimes it's running for hours without problems, but every now and then it crashes, so I cannot just leave it running during the night and be done with it by the morning. I believe this is mostly due to running the revdepchecks over multiple CPU cores in parallel as the crashes never seem to happen when running on a single core only. Also, sometimes there are failures that happen when running the tests in parallel but they disappear when I start running them on a single thread only.

And of course there are always false positives due to network outages etc; they are somewhat cumbersome to deal with as you sometimes need to dig into the test output to understand what's happening and in 99% of the cases it's a waste of time.

PR #612, opened 2022-12-18
chore: Apply styler

krlmlr:
Each commit represents increasing strictness. We can tweak the formatting rules to some extent if needed.

All levels up to and including "line_breaks" are guaranteed by styler to give the same parse tree. I reviewed the last three commits very carefully.

I excluded all files that use `graph_from_literal(...)`, `make_graph(~ ...)` or `edges(c(...))`, because these calls are sometimes formatted specially. I'll do a second pass once we agree on the style.


ntamas:
All is fine here. `line_breaks` with `strict = TRUE` might have been a bit too eager to break stuff into multiple lines, but I'm even okay with that. The biggest benefit would be to have some sort of a pre-commit hook that also re-formats files changed within the commit, and/or to have a plugin for the code editors to automatically apply styling when saving a file. Is this possible?

PR #615, opened 2022-12-26
chore: Prepare for using fledge for managing NEWS.md

krlmlr:
This puts `DESCRIPTION` under version control, and tweaks `NEWS.md`. The version number and changelog will be maintained by {fledge}. It will run in GitHub Actions, once per night (or per week). It will list the most recent changes from the merged PRs and top-level commits to `NEWS.md` and increase the version, whenever there has been a change.

Each such version will get a tag and a top-level entry in `NEWS.md`. Before release, the `NEWS.md` should be edited manually, the autogenerated entries should be merged and groomed as necessary.


szhorvat:
Can you explain what one needs to be aware when adding entries to the changelog after this change?

krlmlr:
This PR doesn't change much. Moving forward, there will be no need to add entries to `NEWS.md` manually, they will be scraped from PR titles and contributors, following "conventional commit" syntax.

ntamas:
One thing that we need to be careful in the future, though, is not to use commit messages that provide no context to the user about what changed. More often than not there are some commits that may be meaningful / relevant to devs but not meaningful to end users. I think that the general rule will be that a draft changelog will be assembled from commit messages, and then we need to go through it before every release manually to remove those that are not relevant to the end user and re-write those that _are_ relevant but have been phrased in a way that does not provide enough context.

szhorvat:
What about changes made on the C side?

ntamas:
I think those should be covered by a "chore: bumped C core to version X.Y.Z" style commit, and if we do bump the C core to fix a bug on the R side, we should probably explain it in the _details_ of the commit message (i.e. the part after the first line). I don't know whether `fledge` will show that by default or not, but it must be part of the commit log review process to replace the generic "bumped C core" message with something better if more context is needed.

krlmlr:
fledge will add a link to the PR, a link to the issue (if applicable), and also mention the contributors (if applicable). The PR body or the body of the commit message is not processed.

szhorvat:
> I think those should be covered by a "chore: bumped C core to version X.Y.Z" style commit

Up til now the R interface did not use released versions of the C core. It used specific commits.

-----

I am not involved deeply enough with the R interface to argue this beyond what I am about to write, but I feel that I need to point out a few problems with this idea ... I use some software that generates so-called "changelogs" from commits. I find these changelogs utterly useless. Here's for example SageMath: https://www.sagemath.org/changelogs/sage-9.7.txt  Take a look, and tell me what version 9.7 brings for users. Should you upgrade? What part of your existing code will break? What new thing will you be able to do?

This approach might bring a minor convenience for developers, but IMO even that is illusory. In order to preserve any amount of usefulness one must be extremely disciplined when preparing commits and writing the commit message. In the end it's more drain on one's time than writing the NEWS file.

For users, however, it is clearly a major step backwards. igraph's users are not developers. They are primarily researchers, some of whom don't even know what is version control, and are not familiar with the technical language used in commits. Even developers would find it hard to glean what the relevant changes are, because changelog generated from commit messages are plainly _bad_, just take a look at that SageMath changelog I linked.

So in the end what is being achieved with this? Who benefits, and what is the benefit? Is there a true net benefit other than the warm fuzzy feelings developers tend to get from automating yet another thing?

I won't comment on this further, but this had to be written down.

krlmlr:
Good points.

- The changelog you linked to contains ticket number and author name at the beginning of each line, I agree that this is suboptimal. Our changelog entries should start with a function name (if applicable) or with a phrase that clarifies if it's worth reading the entire bullet point.
- Entries will be grouped by "conventional commit" code: bug fixes, features, etc., will appear separately.
- I agree that an autogenerated changelog must be curated manually before release to remain useful. This is part of the plan.
- We still can manually edit `NEWS.md` at any time in the process. Edits of existing parts will not conflict with automatically imported entries.

szhorvat:
Let's see how it goes and judge it by the quality of the end product.

PR #616, opened 2022-12-26
Remove igraph version from GML header

krlmlr:
We no longer have `IGRAPH_VERSION` available at compile time, and this is the only instance where it is used.

Patching the C core for now. Moving forward, perhaps `igraph_write_graph_gml()` could accept a version argument that can be provided externally?

### DIFF
--- a/patches/002-workaround-gml-creator.patch
+++ b/patches/002-workaround-gml-creator.patch
@@ -1,0 +1,15 @@
+diff --git a/src/core/io/gml.c b/src/core/io/gml.c
+index 77c7f0541..375523ee6 100644
+--- a/src/core/io/gml.c
++++ b/src/core/io/gml.c
+@@ -611,8 +611,8 @@ int igraph_write_graph_gml(const igraph_t *graph, FILE *outstream,
+     timestr[strlen(timestr) - 1] = '\0'; /* nicely remove \n */
+ 
+     CHECK(fprintf(outstream,
+-                  "Creator \"igraph version %s %s\"\nVersion 1\ngraph\n[\n",
+-                  IGRAPH_VERSION, creator ? creator : timestr));
++                  "Creator \"igraph %s\"\nVersion 1\ngraph\n[\n",
++                  creator ? creator : timestr));
+ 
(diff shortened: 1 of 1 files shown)


ntamas:
> Moving forward, perhaps igraph_write_graph_gml() could accept a version argument that can be provided externally?

This does not seem to fit well with the general usage pattern of a pure C library; i.e. why should we expect the user to provide the version number of the library?

In my opinion, a better way forward would be to let the caller of `igraph_write_graph_gml()` decide what gets written into the "creator" field, falling back to the default "igraph version x.y.z" string if no creator is provided (i.e. it is the null string). The R interface could them simply pass "igraph" without a version number there, or create the string dynamically in the R layer where we can easily query the version number of the library from the package metadata.

szhorvat:
> In my opinion, a better way forward would be to let the caller of `igraph_write_graph_gml()` decide what gets written into the "creator" field, falling back to the default "igraph version x.y.z" string if no creator is provided

That is what 0.10 does.

----

But can you explain what is going on here and why can't `IGRAPH_VERSION`, a standard part of the igraph API, be accessed when compiling the R interface?

If such a patch is necessary on the R side, so be it, but it should not be pushed upstream to the C library. It is not reasonable to ask that `IGRAPH_VERSION` should not be used.

Also, this is not the only place in the C library where `IGRAPH_VERSION` is used.

krlmlr:
Right, the other usage is in `dot.c` ?

The `IGRAPH_VERSION` used here wasn't the same as for the C library, it was overridden by the version of the R package. It should be easy (?) to use the default `IGRAPH_VERSION` of the C library, this would also remove the need for this patch.

ntamas:
@szhorvat The R interface contained an override for the igraph version number; otherwise users would have asked why the version number is completely different from the version number of the R package when they saved a GML file from the R interface. The override was done from the compiler invocation with a macro (i.e. `-DIGRAPH_VERSION=...`), which busted `ccache` every time the version number was bumped on the R side - which happened fairly often as we used the fourth component of the version number to indicate 9000 + the number of commits since the last official release.

szhorvat:
Thanks for the explanation.

PR #617, opened 2022-12-29
fix: Fix builds with some `make` variants

krlmlr:
Introduced recently.


szhorvat:
I confirm that building the package success on my machine with this change.

I have GNU Make 3.81 as included with macOS 10.14.6.

PR #619, opened 2023-01-01
Use *3d functions rather than rgl.* functions

dmurdoch:
`rgl` will be deprecating a number of `rgl.*` functions in an upcoming release.  (See https://github.com/dmurdoch/rgl/blob/deprecated/vignettes/deprecation.Rmd for a discussion of the issues.)  This PR does two things:

 - It substitutes equivalent `*3d` calls for the `rgl.*` calls.
 - It removes workaround code that may have been needed with the `rgl.*` calls, but doesn't appear to be needed now.

### DIFF
--- a/R/plot.R
+++ b/R/plot.R
@@ -822,20 +822,14 @@ rglplot.igraph <- function(x, ...) {
     col = vertex.color
   )
 
-  # add the labels, 'l1' is a stupid workaround of a mysterious rgl bug
+  # add the labels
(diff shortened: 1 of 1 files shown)


ntamas:
LGTM but I'm not familar with `rgl` so I'll leave it to @krlmlr to decide whether this can be merged.

krlmlr (approved):
Thanks, looks good.

PR #620, opened 2023-01-03
chore: Move to single branch setup

krlmlr:
With this, the repo can be simply cloned, and `R CMD INSTALL .` and `devtools::install_github()` install the R package. We need to run `make` only to synchronize the generated files.

Commits can be reviewed individually.

The version number is no longer accessible from the C code. Could we change `configure.ac` so that it is taken from `DESCRIPTION` ?


ntamas:
> Could we change configure.ac so that it is taken from DESCRIPTION ?

Yes, sure.

ntamas:
Note to ourselves: the parser/lexer sources are autogenerated. In order to avoid committing changes to them all the time when `make` is executed on different computers with different versions of `bison` and `flex`, we should probably document which versions are to be used, and explicitly reject unnecessary changes to these files unless a commit is dealing with the parsers.

krlmlr:
Yes, this and the autoconf version. We could agree to let GHA handle it?

krlmlr:
GHA will reject unwanted changes, by the way, for forks. For branches in the main repo, these changes are autocommitted.

ntamas:
> We could agree to let GHA handle it?

Works for me.

krlmlr:
Can we instruct `bison` and `flex` to print their version?

ntamas:
I think `bison --version` and `flex --version` should do the trick.

ntamas:
I don't think so; the best I can think of is to do it in a post-processing step.

krlmlr:
A green build, finally.

I checked, installation from GitHub directly from this branch works. If we agree to move on, the next steps after merging would be:

- remove workflow that updates the `master` branch (not done here yet)
- force-push `dev` to `master`
- set main branch to `master`
- update GitHub installation instructions, remove `@master`
- update r-universe instructions, remove `@master`

After that, we could rename `master` to `main` here and in the other repositories. I can share a code snippet for that.

ntamas:
Looks good to me; I haven't checked the additions (I guess they are the result of adding the vendored source code of igraph) now, but the rest seems okay. Shall we attempt to merge `dev` one last time into this before merging back?

krlmlr:
Done, except for changing the default branch. I don't have access to "Settings" to complete the remaining steps. I now think these should be instead:

- Delete `master` branch on the remote (I have pushed its most recent state to `master-last`)
- Change the default branch from `dev` to `master` (in https://github.com/igraph/rigraph/settings/branches)

This will retarget all open PRs, a good thing, I believe. Can you please take care of this?

ntamas:
I think this is done now - please check and let me know if I missed something.

PR #621, opened 2023-01-03
chore: Track revdepcheck results

krlmlr:
Only three packages failing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 /igraph_*.tar.gz
 /igraph_*.tgz
 /object_files
-/revdep/
 /version_number
 /configure.ac
 igraph.Rcheck/
--- a/revdep/.gitignore
+++ b/revdep/.gitignore
@@ -1,0 +1,1 @@
+/cloud.noindex/
--- a/revdep/README.md
+++ b/revdep/README.md
@@ -1,0 +1,24 @@
+# Revdeps
+
+## Failed to check (9)
+
+|package   |version |error |warning |note |
+|:---------|:-------|:-----|:-------|:----|
+|DRviaSPCN |?       |      |        |     |
+|genekitr  |?       |      |        |     |
+|immcp     |?       |      |        |     |
+|numbat    |?       |      |        |     |
+|Platypus  |?       |      |        |     |
+|NA        |?       |      |        |     |
+|tidySEM   |?       |      |        |     |
+|vivid     |?       |      |        |     |
+|NA        |?       |      |        |     |
+
+## New problems (3)
+
+|package      |version |error  |warning |note |
+|:------------|:-------|:------|:-------|:----|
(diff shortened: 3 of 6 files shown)


ntamas:
`r6causal` imports `causaleffect` so I strongly suspect that the two failures are related.
As for `sfnetwork`, the output does not seem to provide any useful clues :(

ntamas:
Also, is this _without_ the change proposed to avoid recycling of attribute values?

szhorvat:
> As for `sfnetwork`, the output does not seem to provide any useful clues :(

It referenced `distances()` with the Johnson method, which we did change.

```
   10. └─sfnetworks:::st_network_cost.sfnetwork(...)
   11.   ├─base::do.call(igraph::distances, c(args, dots))
   12.   └─igraph (local) `<fn>`(`<sfnetwrk>`, 1, 10, weights = `<[m]>`, mode = "in", algorithm = "johnson")
```

Right now it only supports `mode='out'` with directed graphs. Here they use `mode='in'`. So that's the reason for the failure.

But I have to say that the error message is very confusing:

```
> g<-sample_gnm(10,20,directed=T)
> distances(g,algorithm='johnson')
Error in distances(g, algorithm = "johnson") : 
  Johnson's algorithm works with mode="out" only for directed graphs
```

 - It's unclear that the problem occurred because the default was `mode='all'`. BTW that's a pretty bad default for distances ... I'd expect that in directed graphs, directed distances will be computed by default. But I guess this can't be fix as it would be a "breaking change"
 - When I read the message, it sounded as if the word "only" referred to "directed". In fact, it refers to the mode setting. It should probably say, "Johnson's algorithm only supports mode="out" when using directed graphs."

krlmlr:
Yes, this is the baseline.

szhorvat:
It might make sense to implement `mode='in'` for Johnson on the R side. All it really takes is swapping the from/to arguments an transposing the result matrix. Throughts?

szhorvat:
It's good to note that [the test sfnetwork uses](https://github.com/luukvdmeer/sfnetworks/blob/c18078a72cb93b8b73d12ea45c402f359951593b/tests/testthat/test_paths.R#L237) is somewhat pointless, as it's done on a graph with non-negative edge weights. Johnson's algorithm is only useful with directed graphs that have some negative weights. If the graph is not like this, the C function just calls the Dijkstra implementation anyway.

ntamas:
> All it really takes is swapping the from/to arguments an transposing the result matrix. Thoughts?

Does it? Wouldn't we also need to reverse the edge directions?

szhorvat:
That function only returns distances, not paths. The A-to-B distance with "out" is the same as the B-to-A distance with "in", no?  With "in" we go against the arrows in `A--->B`

krlmlr:
Merging to add infrastructure to track revdepcheck results on GitHub.

PR #624, opened 2023-01-04
Use markdown in documentation

krlmlr:
I did multiple passes over the documentation. Trying to consequently use `()` when referring to functions, and omitting when not. I probably missed something, a review might help (but the documentation is so large...).

I noticed the use of `\sQuote{}` and `\dQuote{}`, and also omitting of quotes altogether, when referring to strings. I propose to use `` `"string"` ``, but this is a separate PR.


krlmlr:
Merging this requires a leap of faith, but also helps improve the docs in the medium term. Better to merge before #625, there are conflicts, but I can handle them easier in that branch.

ntamas:
> I noticed the use of \sQuote{} and \dQuote{}, and also omitting of quotes altogether, when referring to strings. 

I think those were completely arbitrary, there weren't any formal rules, so any improvement on that front is welcome. I'll review this PR and then merge if OK.

ntamas (approved):

ntamas:
Haven't checked all the files one by one but I've spent a fair amount of time reading the new docs in random places and it seemed OK to me. Let's merge and fix issues later if we find any.

PR #625, opened 2023-01-04
fix: `NULL` means empty vertex or edge list again in all functions

### LINKED ISSUE #623: Treatment of `NULL` vertex and edge sets
The interpretation of `NULL` vertex and edge sets has changed in d05ac3140bde37d53365da464e23b22fe89e1b90, to support #466. This leads to failures in the causaleffect package, https://github.com/igraph/rigraph/pull/621/files#diff-bbdda7f28391ee0ec001638e5603609582124345e0d65ebd4747cc592f9bd56dR30-R38.

That package is calling `subgraph.edges(g, NULL)` at some point. Should this be an error (better, but with breaking changes), or do we want to support the current (implicit) behavior that `NULL` means empty vertex/edge set?

This potentially affects all functions that use `as.igraph.vs()` or `as.igraph.es()` .

## Current dev

``` r
g <- igraph::make_ring(10)
igraph:::as.igraph.vs(g, NULL)
#> + 10/10 vertices, from 881e524:
#>  [1]  1  2  3  4  5  6  7  8  9 10
igraph:::as.igraph.es(g, NULL)
#> + 10/10 edges from 881e524:
#>  [1] 1-- 2 2-- 3 3-- 4 4-- 5 5-- 6 6-- 7 7-- 8 8-- 9 9--10 1--10
```

<sup>Created on 2023-01-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## CRAN version

``` r
g <- igraph::make_ring(10)
igraph:::as.igraph.vs(g, NULL)
#> numeric(0)
igraph:::as.igraph.es(g, NULL)
#> numeric(0)
```

<sup>Created on 2023-01-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

krlmlr:
Closes #623.

Reverts #607 and #608.

Confirming that it fixes the causaleffect problem.

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -145,7 +145,7 @@ graph.attributes <- function(graph) {
 #' @param graph The graph.
 #' @param name Name of the attribute to query. If missing, then
 #'   all vertex attributes are returned in a list.
-#' @param index A vertex sequence, to query the attribute only
+#' @param index An optional vertex sequence to query the attribute only
 #'   for these vertices.
 #' @return The value of the vertex attribute, or the list of
 #'   all vertex attributes, if `name` is missing.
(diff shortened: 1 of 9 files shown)


krlmlr:
I have no idea why the tests are so flaky on macOS and with coverage. Perhaps this PR helps stabilize, though I don't understand why.

krlmlr:
Rerunning CI/CD.

PR #626, opened 2023-01-04
WIP: Special case

krlmlr:
Does this give green builds?


krlmlr:
Doesn't look like it does.

PR #628, opened 2023-01-05
distances() with algorithm='johnson' now handles all mode parameter values correctly

### LINKED ISSUE #571: distances() ignores 'mode' when algorithm='johnson'
**Describe the bug**

The `distances()` function ignores the `mode` parameter when `algorithm='johnson'` because `igraph_distances_johnson()` does not support setting the mode: https://github.com/igraph/igraph/issues/2193

This can lead to confusion and invalid results.

I suggest that when `algorithm='johnson'` and `mode != 'all'` for undirected or `mode != 'out'` for directed an error should be thrown.

**To reproduce**

```r
> g<-make_graph(c(1,2,2,1,2,3))
> distances(g)
     [,1] [,2] [,3]
[1,]    0    1    2
[2,]    1    0    1
[3,]    2    1    0
> distances(g, algorithm='dijkstra')
     [,1] [,2] [,3]
[1,]    0    1    2
[2,]    1    0    1
[3,]    2    1    0
> distances(g, algorithm='johnson')
     [,1] [,2] [,3]
[1,]    0    1    2
[2,]    1    0    1
[3,]  Inf  Inf    0
```

Notice that since the default mode is `'all'`, the graph is treated as undirected by both the default algorithm and Dijkstra. But with Johnson it is unexpectedly treated as directed.

**Version information**

 - R/igraph version: 1.3.4

szhorvat:
This PR fixes #571 properly by bringing support for all `mode` values instead of restricting the available modes when `algorithm='johnson'`.

### DIFF
--- a/R/structural.properties.R
+++ b/R/structural.properties.R
@@ -266,9 +266,12 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
 #' unweighted graphs; the Dijkstra algorithm (\sQuote{`dijkstra`}), this
 #' works for graphs with non-negative edge weights; the Bellman-Ford algorithm
 #' (\sQuote{`bellman-ford`}), and Johnson's algorithm
-#' (\sQuote{`"johnson"`}). The latter two algorithms work with arbitrary
+#' (\sQuote{`johnson`}). The latter two algorithms work with arbitrary
 #' edge weights, but (naturally) only for graphs that don't have a negative
-#' cycle.
+#' cycle. Note that a negative-weight edge in an undirected graph implies
+#' such a cycle. Johnson's algorithm performs better than the Bellman-Ford
+#' one when many source (and target) vertices are given, with all-pairs
+#' shortest path length calculations being the typical use case.
 #'
 #' igraph can choose automatically between algorithms, and chooses the most
 #' efficient one that is appropriate for the supplied weights (if any). For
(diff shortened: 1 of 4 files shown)


szhorvat:
I just updated to the tip of `dev` and added some more docs.

szhorvat:
I don't understand why the CI run fails. Building the package succeeds on my machine. Is this because I edited the docs and did not update `.Rd` files? How can I update those files?

szhorvat:
There's an ugliness here I should mention:

The running time of all of the methods implemented here primarily depends on the number of sources, and is approximately independent of the number of targets. Thus the typical use is to calculate shortest path length from _some_ vertices to _all_ other vertices. Since the solution here swaps sources with targets, the performance characteristics are no longer as expected.

szhorvat:
In fact, in practice, Bellman-Ford tends to be faster for sparse graphs than Johnson. The core reason is that Bellman-Ford is even faster than Dijkstra in these cases, and all that Johnson does is allow the use of Dijkstra with negative weights. Example:

```
set.seed(123)
g<-sample_gnm(3000,10000,directed = T)
E(g)$weight<-runif(ecount(g),min=-0.01,max=1)
bench::mark(distances(g,mode='out',algorithm = 'bellman-ford'))
bench::mark(distances(g,mode='out',algorithm = 'johnson'))

E(g)$weight<-runif(ecount(g),min=0.01,max=1)
bench::mark(distances(g,mode='out',algorithm = 'dijkstra'))
```

Here I got:

 - Bellman-Ford: 1.09 s
 - Johnson: 1.63 s
 - Dijkstra: 1.57 s

I'll have to check what's happening with igraph 0.10. I know that this is not how it works with dense graphs, as we have a benchmark for those: https://github.com/igraph/igraph/blob/master/tests/benchmarks/igraph_distances.c

But all this is now in the domain of the C core.

PR #631, opened 2023-01-11
chore: Stabilize vs-es tests

krlmlr:
I'm having a hard time understanding why this particular test fails. I tried to replicate multiple times. Let's skip for now and monitor.


ntamas:
Would you like to merge this now and then keep on watching the `dev` branch, or shall we keep this open on the side and update with `dev` occasionally?

krlmlr:
I seriously wonder why covr is still failing. I'll take another look.

The intention is to merge this so that we have green builds again.

PR #634, opened 2023-01-15
Fix test order

krlmlr:
for faster parallel testing and to avoid having to skip tests.

The vs-es failures occur sporadically with parallel testing, I hope that they disappear if these tests always run first.


krlmlr:
It seems that an empty edge set sometimes prints differently. I propose to treat this as a non-issue until we can replicate it in a situation that matters in practice.

ntamas (approved):

PR #635, opened 2023-01-16
Update instructions

krlmlr:
Follow-up to https://github.com/igraph/rigraph/pull/620#issuecomment-1383069883.


krlmlr:
jammy should have autoconf 2.71, I wonder why the configure script hasn't been updated.

PR #638, opened 2023-01-16
1st config draft

maelle:
This is a draft I'll improve.


maelle:
First draft code

```r
r_files <- fs::dir_ls("R")

collect_families <- function(filename) {
  lines <- brio::read_lines(filename)
  family_lines <- which(grepl("#' @family", lines))

  if (length(family_lines) == 0) {
    return()
  }

  families <- lines[family_lines]
  unique(trimws(gsub("#' @family", "", unique(families))))
}

families <- unlist(purrr::map(r_files, collect_families))

config_pattern <- function(family) {
  c(
    sprintf("- title: %s", snakecase::to_sentence_case(family)),
    "- contents:",
    sprintf('  - has_concept("%s")', family)
  )
}

others <- c(
  sprintf("- title: Others"),
  "- contents:",
  sprintf('  - lacks_concepts(c(%s))', toString(sprintf('"%s"', families)))
)

pkgdown_lines <- c(
  "url: ~",
  "",
  "template:",
  "  bootstrap: 5",
  "",
  "reference:",
  unlist(purrr::map(families, config_pattern)),
  others
)

brio::write_lines(pkgdown_lines, "_pkgdown.yml")

```

maelle:
Way I added more family tags

```r
pkg <- pkgdown::as_pkgdown()
ref <- pkg$topics

# we can't treat topics with more than one source yet
ref <- ref[lengths(ref[["source"]]) == 1,]

treat_source <- function(source, ref) {
  cat(sprintf("Treating %s\n", source))

  if (source == "R/igraph-package.R") {
    # https://github.com/igraph/rigraph/issues/639
    return()
  }

  ref <- ref[ref[["source"]] == source,]

  # Mix of concepts and not: fix by hand¿
  if (nrow(ref[lengths(ref[["concepts"]]) == 0,]) != nrow(ref)) {
    cat(sprintf("Not all exports have concepts in %s\n", source))
    return()
  }

  if (nrow((ref[lengths(ref[["concepts"]]) == 0,])) == 0) {
    return()
  }

  rd_lines <- brio::read_lines(source)

  family_tag <- sprintf("#' @family %s", fs::path_ext_remove(basename(source)))

  exports <- which(grepl("#' @export", rd_lines))

  for (i in seq_along(exports)) {
    rd_lines <- append(
      rd_lines[-exports[i]],
      values = c(family_tag, "#' @export"),
      after = (exports[i] - 1)
    )
    # each round we are adding one line
    # so the exports' position changes
    exports <- which(grepl("#' @export", rd_lines))
  }
  brio::write_lines(rd_lines, source)
}

purrr::walk(unique(ref[["source"]]), treat_source, ref = ref)

```

maelle:
(then `document()` and then running the first code)

NAMESPACE
```diff
@@ -706,7 +706,6 @@ export(path.length.hist)
 export(permute)
 export(permute.vertices)
 export(piecewise.layout)
-export(plot.igraph)
```
maelle, line 709: oops, not sure why running `devtools::document()` would remove those? or was `NAMESPACE` out of date?

_pkgdown.yml
```diff
@@ -0,0 +1,222 @@
+url: ~
+
+template:
+  bootstrap: 5
+
+reference:
+- title: Adjacency
+- contents:
+  - has_concept("adjacency")
```
krlmlr, line 9: Are you sure about the two dashes here and everywhere?

```suggestion
- title: Adjacency
  contents:
  - has_concept("adjacency")
```

NAMESPACE
krlmlr: Interesting (for some definition of).

I wonder if this still happens if we keep the source files that define `plot.igraph()` and `print.igraph()` unchanged.

.gitignore
```diff
@@ -13,3 +13,4 @@ igraph.Rcheck/
 .Rproj.user
 /configure~
 /build/
+docs
```
krlmlr, line 16: Let's be precise here:

```suggestion
/docs/
```

krlmlr (commented):
Thanks, nice idea to use the `@family` tag!

_pkgdown.yml
maelle: yes it works and it comes straight from https://pkgdown.r-lib.org/reference/build_reference.html#reference-index (I agree it's not intuitive)

krlmlr:
Thanks, looks good! Can you please deploy the pkgdown to the gh-pages branch in your fork, and also add a GHA workflow?

krlmlr:
@ntamas suggested using the ordering/hierarchy in the C library docs at https://igraph.org/c/html/latest/. Moving forward, can you please align the pkgdown docs with the C library docs?

I wonder if we should edit https://github.com/igraph/igraph/blob/master/interfaces/functions.yaml:

- Add more sectioning
- Reorder to match the pkgdown docs

The advantage here is that we could later create the `@family` tags and also the ordering/hierarchy for Python and Mathematica docs, from a single source.

ntamas:
We can even start adding section/subsection tags to `functions.yaml` - it does not have a strict syntax and any keys under the function names that are not known to the parser are silently ignored as of now.

Another YAML file could then provide proposed titles for the sections/subsections and the documentation could use those.

ntamas:
@maelle Can you provide us mere mortals with instructions about how to run the whole `pkgdown` pipeline? Is it as easy as `Rscript -e 'pkgdown::build_site()'` or is there more to it?

krlmlr:
Thanks, looks good.

I don't understand the comment regarding the changes in the C library config. My request was about ordering the topics in the pkgdown reference index, to align them with what exists now in the C library docs.

szhorvat:
Regarding the C library docs, they are not perfect, and could be organized better too. IMO we can be flexible here, and don't need to force a one-to-one correspondence.

maelle:
so later this week or next, should I organize the pkgdown config following the current state of the C library docs?

szhorvat:
My point was that while it makes sense to organize it similarly, we don't need to be too rigid about it. The most important things is that the organization should be clear and easy to navigate. If something seems to not make sense, IMO we can also adapt/reorganize the C docs. Just ask.

maelle:
@szhorvat 

>  so it's useful for R users to sometimes refer to the C docs as well.

Beyond the scope of this PR but I think that if it's the case, a few lines could be added to the R docs to state that and roughly explain how to navigate the C docs for usage in R.

szhorvat:
> in `functions.yaml`, could one add nesting?

Note that there is no one-to-one correspondence between R/igraph functions and the contents of that file. Furthermore, this file is not specific to R/igraph—it is intended to eventually be used with the other high-level interfaces as well. The high-level interfaces differ more from each other than one might naively expect. They don't just provide plain "bindings" to the C library. Instead each one tries to integrate well into its host language, which means that necessarily need to be different from each other.

The intention with `functions.yaml` is to automatically generate some low-level _private_ bindings to the C library that can be used as building blocks for a nicely designed high-level interface.

I do not think that we should start mixing up `functions.yaml` with the documentation, or increase the strength of coupling between specifically the R interface and `functions.yaml`.

szhorvat:
@maelle If you want me to send you a file where each of these functions is tagged according to their section in the C/igraph docs, I can do that. Just ask.

ntamas:
Instead of messing with `functions.yaml`, I'd rather have a `docs.yaml` or something like that, with top-level categories and a raw list of functions that belong to those headings underneath.

krlmlr (approved):
Should we merge this first, to get something (and the big differences out of the way), and then improve gradually?

szhorvat:
I hope this helps. These are extracted from the XML files produces for the C/igraph 0.9 documentation. It's similar to what you were requesting for `functions.yaml` / `docs.yaml`, but it has fewer functions (some functions are not properly documented yet, especially in 0.9 — 0.10 is better but different)

[igfuns.json.gz](https://github.com/igraph/rigraph/files/10479941/igfuns.json.gz)

maelle:
Thank you @szhorvat!

Writing the order below that I will use too

```r
order <- c(
Introduction",
Installation",
Tutorial",
Basic data types and interface",
Error handling",
Memory (de)allocation",
Data structure library: vector, matrix, other data types",
Random numbers",
Graph generators",
Games on graphs",
Vertex and edge selectors and sequences, iterators",
Graph, vertex and edge attributes",
Structural properties of graphs",
Graph cycles",
Graph visitors",
Cliques and independent vertex sets",
Graph isomorphism",
Graph coloring",
Graph motifs, dyad census and triad census",
Generating layouts for graph drawing",
Reading and writing graphs from and to files",
Maximum flows, minimum cuts and related measures",
Vertex separators",
Detecting community structure",
Graphlets",
Hierarchical random graphs",
Embedding of graphs",
Graph operators",
Using BLAS, LAPACK and ARPACK for igraph matrices and graphs",
Bipartite, i.e. two-mode graphs",
Advanced igraph programming",
Non-graph related functions ",
Licenses for igraph and this manual"
)
```

(webscraped from the docs website)

szhorvat:
Just noting that these don't apply to R/igraph at all:

```
Basic data types and interface",
Error handling",
Memory (de)allocation",
Data structure library: vector, matrix, other data types",
Random numbers",
Vertex and edge selectors and sequences, iterators",
Using BLAS, LAPACK and ARPACK for igraph matrices and graphs",
Advanced igraph programming",
```

Regarding `Vertex and edge selectors and sequences, iterators`, R/igraph has _vertex and edge sequences_ instead. It has no iterators.

Regarding `Using BLAS, LAPACK and ARPACK for igraph matrices and graphs`, R/igraph exposes an API to ARPACK, but not to BLAS/LAPACK (which is the domain of the core R)

Regarding `Generating layouts for graph drawing`, R/igraph has advanced visualization funtionality, and the title should reflect that. The C core can only compute vertex coordinates ("layouts") and that's all.

maelle:
good catch, indeed I was looking at the wrong docs. will the R package eventually be based on a newer C library version?

szhorvat:
Generally speaking, feel free to use your judgement and rename or split up sections as it seems reasonable. If something's not clear, just ask. E.g. the gigantic structural properties section will definitely need to be split up at some point. https://igraph.org/c/html/latest/igraph-Structural.html

As I understand, you are re-organizing the sections shown in the sidebar, right?

image

If so, feel free to keep many of them separate if that seems to make sense instead of merging everything into a huge "structural properties" section.

szhorvat:
Sorry about the many comments:

One more suggestion I have is to include "Graphlets" under "Cliques". 

"Graphlet" is a term that has been used to mean several different things, and I'm sure that many people will be confused by the usage in igraph. "Graphlet" is more commonly used to refer to _motifs_. What igraph implements is an entirely unrelated niche concept related to cliques.

maelle:
I'm trying to see if I can somehow automated the grouping of R topics into one of the groups. For instance `igraph_add_edges` has the equivalent `add_edges` but `igraph_barabasi_aging_game` is `aging.barabasi.game`. 

At the moment I think the easiest solution is for me to manually place the different "families" of this PR under the categories, using the JSON as reference. For instance I see "games" fit under "Graph generators".

maelle:
> Sorry about the many comments:

Please don't be, that's very useful!

szhorvat:
I'm wondering if some sections could be excluded completely from the index. Nexus used to be an online network repository, but it has been defunct for years, and we don't have the resources to restart it. Since the functions are useless anyway, wouldn't it be better to avoid mentioning them, @ntamas ?

maelle:
Current target order

```r
order <- c(
"Introduction",
"Installation",
"Tutorial",
"Graph generators",
"Games on graphs",
"Graph, vertex and edge attributes",
"Structural properties of graphs",
"Graph cycles",
"Graph visitors",
"Cliques and independent vertex sets",
"Graph isomorphism",
"Graph coloring",
"Graph motifs, dyad census and triad census",
"Generating layouts for graph drawing",
"Reading and writing graphs from and to files",
"Maximum flows, minimum cuts and related measures",
"Vertex separators",
"Detecting community structure",
"Graphlets",
"Hierarchical random graphs",
"Spectral coarse graining",
"Embedding of graphs",
"Graph operators",
"Bipartite, i.e. two-mode graphs",
"Non-graph related functions ",
"Licenses for igraph and this manual"
)
```

szhorvat:
> but `igraph_barabasi_aging_game` is `aging.barabasi.game`.

Actually, `aging.barabasi.game` should be considered deprecated, like most functions with dots in their name. The new name for it is `sample_pa_age`.  I'm sorry about this awful mess we have. As you can see, there's a lot to clean up ...

@maelle Is my understanding correct that the sections you want to have are what I currently see in the right sidebar on https://maelle.github.io/rigraph/ ?  If yes, how were these generated? Each function is tagged with a section already?  I think these categories are not bad, but many need a renaming, and some should be merged.  I could help with this if you want, but I don't know how to best communicate the reorganization. Also, unfortunately I won't have time before the week of Feb 6 ...

maelle:
Do you mean excluding them from both the pkgdown website index and from the R package manual index?
If so in the source we'll need to add

```r
#' @keywords internal
```

If they're just to be removed from the pkgdown website, they need to be listed in a section called internal in the pkgdown index reference.

Lastly, you could also choose to remove them from the package entirely but I don't think that's what you mean. :-)

szhorvat:
> Lastly, you could also choose to remove them from the package entirely

Personally I think that's the best idea, but this is up to @ntamas who has a much better understanding of what kinds of changes CRAN is likely to complain about :-)

maelle:
The sections are for https://maelle.github.io/rigraph/reference/index.html

Currently we're using "families" as a few of the scripts had those, and we added the same family tag to topics that are in the same R script. All topics in blabla.R are in the family "blabla". These families are too numerous and ordered alphabetically, so yes I want to regroup and order them using the JSON you shared. I'll try to make progress later this week.

szhorvat:
If I wrote you a list about which of those sections/familes should be merged, and how they should be renamed, would that help? Then perhaps we can order them at the end?

Or if you prefer, which of those current sections go into which new one listed in https://github.com/igraph/rigraph/pull/638#issuecomment-1400480504

maelle:
> Or if you prefer, which of those current sections go into which new one listed in https://github.com/igraph/rigraph/pull/638#issuecomment-1400480504

Thank you! If you have time to, that'd probably be better than my beginner guessing. The families are the strings called by `has_concept()` in https://github.com/maelle/rigraph/blob/config-draft/_pkgdown.yml.

However you might find the families themselves not very well defined (since we used that "one family per script" strategy) -- in that case please tell us so so we can improve. 

Also note the last section https://maelle.github.io/rigraph/reference/index.html#others is a catch-all for the functions without a family yet.

It's all a lot of work but will really make a difference in the user experience. :sparkles:

ntamas:
> Nexus used to be an online network repository, but it has been defunct for years, and we don't have the resources to restart it. Since the functions are useless anyway, wouldn't it be better to avoid mentioning them, @ntamas ?

I'm all in favour of hiding these functions from the documentation :) As for removing them completely - I think that the ultimate test that decides whether something can go into CRAN or not is the reverse dependency check. So, if we remove these functions, run a revdepcheck and nothing breaks, then I'd say it's pretty safe to remove these functions completely.

ntamas:
I have noticed that `plot.igraph` and `print.igraph` have been removed from the `NAMESPACE` file, which now gives a `NOTE` in the output of `R CMD check`. Was this intentional?

PR #645, opened 2023-01-23
docs: Move vignette to standard location

maelle:
Moving the vignette means pkgdown will pick it up.
Naming it like the package means pkgdown will link to it from the [navbar with the text "Get started"](https://pkgdown.r-lib.org/reference/build_site.html#navigation-bar).

Now, I am not aware of why it was in doc in the first place, so this might be a bad idea.

No preview as the workflow is added in another PR #638, but I checked I could build the website locally.

Cc @krlmlr


DESCRIPTION
```diff
@@ -44,3 +46,4 @@ Config/testthat/parallel: true
 Config/testthat/start-first: vs-es, scan, vs-operators, weakref, watts.strogatz.game
 RoxygenNote: 7.2.3
 Roxygen: list(markdown = TRUE)
+VignetteBuilder: knitr
```
maelle, line 49: I ran `usethis::use_vignette("igraph")` to create the vignette. It added the necessary dependencies (rmarkdown, knitr) to `DESCRIPTION`.

iosonofabio:
Thank you.

No reason why it was there in particular AFAIK, feel free to move. We are working on a Spanish vignette as well, and perhaps further languages. Do you know where those RMD files for translations would go by any chance?

maelle:
> We are working on a Spanish vignette as well, and perhaps further languages. Do you know where those RMD files for translations would go by any chance?

Very good question... No there is no standard yet. I'd suggest adding the language code to the filename.

For info I am (unrelatedly to cynkra) involved in [rOpenSci multilingual publishing project](https://ropensci.org/multilingual-publishing/) where we do not translate vignettes yet though. We use DeepL API to get a first translation draft that human translators then fix. 

In any case it'll make sense to tweak the pkgdown configuration to have several vignettes listed under "Get Started" when you add the vignettes.

PR #646, opened 2023-01-23
docs: link to installation troubleshooting info

maelle:
@szhorvat would it make sense to retire the wiki page and transfer the content a vignette/article instead so that it'd end up on the pkgdown website (and in the pkgdown search index)?

cc @krlmlr

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,12 @@ For installation from source on Windows, you need to have
 Additionally, the three system requirements of `glpk`, `libxml2` and `gmp` are
 not optional, but hard requirements. For versions R >= 4.0 you can install these
 using
+
 ```
 pacman -Sy mingw-w64-{i686,x86_64}-glpk mingw-w64-{i686,x86_64}-libxml2 mingw-w64-{i686,x86_64}-gmp
 ```
+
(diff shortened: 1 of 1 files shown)


szhorvat:
Personally I don't mind about where the troubleshooting info lives, but I think it's important to be able to update the troubleshooting page without having to make a new igraph release. Would that still be possible?

maelle:
> Personally I don't mind about where the troubleshooting info lives, but I think it's important to be able to update the troubleshooting page without having to make a new igraph release. Would that still be possible?

Yes, if the current pkgdown deployment remains the same (well #638 is not merged yet), with the documentation website always generated for each commit on the default branch. This has pros and cons as explained in https://r-pkgs.org/website.html#development-mode 

> "Users who install from CRAN get version 0.2.0. But the pkgdown site is built from the development version of the package."

> "This creates the possibility that users will read about some new feature on the website that is not present in the package version that they have installed with install.packages(). We find that the simplicity of this setup outweighs the downsides, until a package has a broad user base, i.e. lots of users of varying levels of sophistication. It’s probably safe to stay in “release” mode until you actually hear from a confused user."

ntamas:
I'm in favour of moving the installation troubleshooting info to a vignette if that improves discoverability for R users. Until then, I'm going to just merge this PR so at least the link to the wiki appears on the homepage.

PR #661, opened 2023-02-03
CRAN release v1.4.0

ntamas:
- [x] Resolve all tasks in `cran-comments.md`
- [x] Run `fledge::release()`
- [ ] When the package is accepted on CRAN, run `fledge::post_release()`


krlmlr:
origin/main really should never be merged here, look what happened to `DESCRIPTION` .

szhorvat:
If you end up having to do another submission, consider including this fix, made in the C core: https://github.com/igraph/igraph/commit/3a22a3fa29d11b55e0bc41a7bf2a4000d1330857

szhorvat:
> If you end up having to do another submission, consider including this fix, made in the C core: [igraph/igraph@3a22a3f](https://github.com/igraph/igraph/commit/3a22a3fa29d11b55e0bc41a7bf2a4000d1330857)

Added an extra commit to slightly relax the checks.

krlmlr:
I removed the merge and force-pushed.

inst/CITATION
```diff
@@ -1,15 +1,10 @@
-citHeader("To cite 'igraph' in publications use:")
-
-citEntry(entry="article",
-	 title="The igraph software package for complex network research",
-         author=personList(as.person("Gabor Csardi"),
-			   as.person("Tamas Nepusz")),
-	 journal="InterJournal",
+bibentry(bibtype="Article",
+         header="To cite 'igrpah' in publications use:",
```
krlmlr, line 2:
```suggestion
         header="To cite 'igraph' in publications use:",
```

krlmlr:
Oh, these are done automatically.

🤦🤦🤦 

Trying another route. I believe the correct/suggested way is to avoid mentioning C++1x in `DESCRIPTION` .

ntamas:
Is there a way to check whether a CRAN review is already in progress? I kinda lost track of where we are with the submission process after several back-and-forth rounds with automatic CRAN pretest failures on Friday; I think that my last submission passed the pre-tests and it's currently in review, but I'm not 100% sure.

krlmlr:
``` r
foghorn::cran_incoming("igraph")
#> # A tibble: 1 × 5
#>   package version    cran_folder time                size 
#>   <chr>   <pckg_vrs> <chr>       <dttm>              <chr>
#> 1 igraph  1.4.0      waiting     2023-02-03 14:38:00 2.7M
```

<sup>Created on 2023-02-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

krlmlr:
I believe "waiting" means they're waiting for us to respond. But let's just submit a new version, there's a typo anyway.

ntamas:
> I believe "waiting" means they're waiting for us to respond. But let's just submit a new version, there's a typo anyway.

Well, they asked me via email whether the "additional issues" related to Apple M1 have been fixed; I responded with yes and there was no response since.

krlmlr:
It seems safest to just resubmit, to make sure the tag + release on GitHub is consistent with what is on CRAN.

ntamas:
Okay, I won't do this tonight but if you think it's okay to attempt another submission tomorrow in the morning, just let me know or mark it as ready for review and I'll trigger a submission tomorrow.

szhorvat:
Will you include the Infomap fix then?

ntamas:
> Will you include the Infomap fix then (which I backported to the 0.9 C core for R's sake)?

Yes, if we make another submission tomorrow anyway then it costs nothing to bump the C core before we submit.

krlmlr:
I force-pushed and ensured that the auto-merge no longer happens.

krlmlr:
Sorry for the mess, this is a case I hadn't encountered before. Fixing now to avoid further pain downstream.

DESCRIPTION
```diff
@@ -35,8 +35,7 @@ URL: https://igraph.org, https://igraph.discourse.group/
 SystemRequirements:
     gmp (optional),
     libxml2 (optional),
-    glpk (>= 4.57, optional),
-    C++17
```
ntamas: @krlmlr This change breaks the CRAN pretests again with the following `NOTE`:

```
* checking C++ specification ... NOTE
  Specified C++11: please update to current default of C++17
```

I'm going to revert this for the sake of the submission and then we should decide what the long-term solution should be.

szhorvat:
This C++17 requirement is weird. It's not made clear at all in the documentation, which is a bit incoherent at the moment:

https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Using-C_002b_002b-code

I think they might be worried that packages may be using features that were technically removed (or deprecated) in C++17.

I should note that even igraph 0.10 compiles perfectly fine with GCC 4.8 (I test it occasionally).

szhorvat:
Why was the GLPK requirement removed?

DESCRIPTION
krlmlr: Where did we specify C++11?

krlmlr: Oh, got it, in `CXX_STD` .

krlmlr:
I don't fully understand the C++11 and C++17 issues yet. Would it work if we avoid specifying `CXX_STD` entirely?

If we keep `CXX_STD`,  we probably also need `SystemRequirements: C++17` back.

szhorvat:
> On the other hand, specifying C++11 when the code is valid under C++14 or C++17 reduces future portability.

FWIW, the `release/0.9` branch, which R/igraph is based on, compiles with no warning when using a December prerelease of Clang 16 (pickiest compiler) and specifying both C17 and C++17. So no future compatibility issues to worry about for now.

(Specifying C23 does trigger a bunch of warnings, but we're far from that—it's not even a finalized standard yet.)

ntamas:
> This C++17 requirement is weird.

Agreed. The one thing that really bothers me is that it appears in `SystemRequirements` as if a C++17 compiler would be _required_ to compile igraph, which in fact isn't required. Declaring C++17 as a requirement might prevent igraph from compilation on R 3.5 and older because Rtools35 did not inlcude a C++17 compiler on all platforms as far as I know. On the other hand, R 3.5 is ancient so we could just as well drop support for it. Nevertheless, it feels like an unnecessary change that's forced on us by CRAN policies even though it would be completely unnecessary.

> Would it work if we avoid specifying CXX_STD entirely?

Maybe, but the "Writing R Extensions" bible says that "[d]ifferent versions of R have used different default C++ standards, so for maximal portability a package should specify the standard it requires". Which is, again, inconsistent with what we actually had to do because "for maximal portability" we should require C++11 only, not C++17.

What's really annoying is that it's just a NOTE in the CRAN pretest output but still it fails the entire pretest process.

szhorvat:
Well, R 3.5 is 5 years old. At some point it makes sense to move on so newer R features can be relied on. We're bumping the version to 1.4, so it's not unreasonable to drop support for old releases at the same time.

R is free in every sense. I can't think of a valid reason to want to use a _new_ igraph version with an _old_ R. It's not like supporting old Mathematica versions because some users would need to pay to upgrade. If you're stuck on old R, use old igraph too.

krlmlr:
Some users aren't free to choose what version of R they can run their code on. Often it's mandated by IT, and this means they can't get the shiny new stuff we're building.

szhorvat:
@krlmlr :

What would you propose as a policy for updating requirements?

Talking about 3.5 specifically, in your experience is it still used widely enough in 2023 so as to be a concern?

Did more recent versions of R introduce any features that would be beneficial to be used, but we can't while supporting 3.5?

Would it make sense to explicitly indicate the minimum supported R version, like some other packages do? E.g. Matrix requires R 3.5, meaning that we can't go older than that (Matrix is an igraph dependency). https://cran.r-project.org/web/packages/Matrix/index.html

krlmlr:
I canceled the workflow to avoid the auto-commit.

szhorvat:
CRAN checks fail, apparently because linking is now done with the C compiler (clang, gcc), not the C++ compiler (clang++, g++), as it was before and as it should be.

https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/igraph-00check.html

ntamas:
Probably a side effect of the removal of `STDCXX` from the `Makefile` -- which was done _exactly_ because the CRAN pretests complained about it...

Note that the failure happens only with `clang`, `gcc` seems to be okay. Anyhow, it looks like we'll need a 1.4.1 soon...

krlmlr:
Yes to the `post_release()` part. Let's understand the reason for the failures first before opening a new release branch. We can release again on 2023-02-17 without extra NOTEs from CRAN.

szhorvat:
For 1.4.1, can you please update the C core to the latest commit to deal with #669?

szhorvat:
There's another issue here that is likely the result of the removal of `STDCXX`.

When igraph is compiled, the `CXX` setting from `~/.R/Makevars` is no longer respected.  It always uses the same compiler that R itself was compiled with. The `CC` setting _is_ respected, only the `CXX` one isn't.

szhorvat:
> When igraph is compiled, the `CXX` setting from `~/.R/Makevars` is no longer respected. It always uses the same compiler that R itself was compiled with. The `CC` setting _is_ respected, only the `CXX` one isn't.

Just to have a written record of this on GitHub:

It seems that this is not igraph-specific. The same thing happens with other packages, such as `rlang`. It may be due to a change in R. A possible solution is using `CXX11` instead of `CXX`.

krlmlr:
FWIW, I had observed different behavior today with clang-16. Not too important, I believe.

PR #671, opened 2023-02-13
chore: Makevars cleanup

krlmlr:
This adds a dummy C++ file to ensure that a C++ linker is always selected. Also, we don't need to ship `src/Makevars` .

Rationale, from r-exts, https://cran.r-project.org/doc/manuals/r-release/R-exts.html#External-C_002b_002b-code:

> Even fewer external libraries use C++ internally but present a C interface, such as rgeos. These require the C++ runtime library to be linked into the package’s shared object/DLL, and this is best done by including a dummy C++ file in the package sources.

This is essentially what we sent to CRAN, plus the dummy C++ file.


src/dummy.cpp
```diff
@@ -0,0 +1 @@
+// This file is required so that package installation uses a C++ linker.
```
szhorvat, line 1: Can you add a link to https://cran.r-project.org/doc/manuals/r-release/R-exts.html#External-C_002b_002b-code in a comment?

szhorvat:
Since this is following advice directory from the R manual, I guess it's fine. Please just keep that link in a comment in the sources.

This file is not going to be compiled, so there won't be issues with empty compilation units (which CRAN would complain about). No need for hacks like [this](https://github.com/igraph/igraph/blob/master/src/internal/glpk_support.c#L166).

szhorvat (approved):

PR #672, opened 2023-02-13
chore: Remove patches already shipped

krlmlr:
Cleanup.


krlmlr:
I was mistaken, looks like these are still needed.

PR #674, opened 2023-02-13
chore: Avoid submodules for building igraph

krlmlr:
We keep using the submodules, but only as a source of the C files. The process of updating the C core is documented in `src/README.md` .


szhorvat:
Makes sense to me, though @ntamas is much more qualified to comment on this type of thing :-)

szhorvat (approved):

src/README.md
```diff
@@ -0,0 +1,25 @@
+# igraph/src
+
+## Updating the C sources
+
+1. Check out and update the `cigraph` submodule to the desired version.
+    ```sh
+    git submodule init
+    git submodule update
+    # git submodule magic
```
szhorvat, line 9: Was this line meant to be left in?

krlmlr: We could replace it by an example to update to a specific SHA. I'm not familiar with `git submodule` syntax.

szhorvat: Ah, that's what you meant. I'll complement this README file later to make that clear, but feel free to merge anytime.

ntamas (approved):

PR #701, opened 2023-03-13
refactor: remove srand()!

### LINKED ISSUE #693: Remove srand()?
After all, it's been creating a `warning()` only since a lot of versions.

maelle:
Fix #693


szhorvat:
One comment here is that these removals are probably best done for 1.5.0, not 1.4.2.

It needs to be decided:

 - Will there be an 1.4.2 or we go straight to 1.5.0?
 - Should R/igraph move to a two-branch development model (`master` / `develop`) like the other igraph repos?

ntamas:
Personally, I have no plans for a 1.4.2, but you never know -- if an urgent issue pops up that needs to be fixed ASAP, then it should probably go into a version 1.4.2.

The two-branch setup probably makes sense in cases where you have intensive ongoing development work that you do _not_ want to squeeze into the next release if the next release needs to be a hotfix release. Here, I don't think that any such development is currently ongoing so it is almost always safe to release from master. So, I'd say let's keep the current one-branch setup for the time being, merge this PR and commit ourselves that the next release will be 1.5.0 and not 1.4.2.

Objections?

ntamas:
Also, note that `srand()` has been printing a warning for _years_ now and had no effect anyway, so I would say it's not even a breaking change if we decide to do a 1.4.2 as the function has been deprecated for a long while now.

So I think I'll just resolve the conflicts and merge this PR as is.

Issue #699, opened 2023-03-10
Negative vertex IDs are handled unintuitively


szhorvat:
In functions that take a single vertex ID as a parameter (instead of a vertex sequence), negative IDs, which should be invalid, are often handled in a strange way. Take for example `shortest_paths()`:

```
> g<-make_ring(4)
```

This is fine:

```
> shortest_paths(g,1,3)$vpath
[[1]]
+ 3/4 vertices, from 9cc5663:
[1] 1 2 3
```

This is fine:

```
> shortest_paths(g,5,3)$vpath
Error in shortest_paths(g, 5, 3) : 
  At core/paths/unweighted.c:294 : cannot get shortest paths, Invalid vertex id
```

What is happening here?

```
> shortest_paths(g,-1,3)$vpath
[[1]]
+ 2/4 vertices, from 9cc5663:
[1] 2 3
```

It turns out that internally, `shortest_paths` uses `as.igraph.vs()` on the `from` parameter, and effectively indexes into the vertex ID vector. And with R's usual quirkiness, `-1` means "everything except `-1`":

```
> igraph:::as.igraph.vs(g,-1)
[1] 2 3 4
```

Then all but the first element gets ignored. Thus, unintuitively, `-1` ends up meaning `2`.

szhorvat:
This relates to the more general problem that the auto-generated R-to-C glue code does not properly validate scalars. If a scalar is expected, but the vector that is passed has length different from 1, then the code blindly take the first element.

When there are multiple element, this just takes the first one, which is not a great sin, but it can still lead to confusion, as you see here. However, if there are zero elements, then we have an out-of-bounds access.

Issue #153, opened 2016-05-14
neighbors bug


gaborcsardi:
``` r
library(igraph)
graph <- make_ring(10)
neighbors(graph, numeric())
#> + 2/10 vertices:
#> [1]  2 10
```

kdaily:
I see this as well in version 1.0.1 - if the list of vertices supplied is empty, there are still edges returned.

szhorvat:
Isn't this a bug in how arguments are translated to C? The `v` argument of `neighbors` should be precisely one integer, not zero, not more than one. The argument translator should catch cases when this is not true. The same thing causes trouble in #140.

ntamas:
It is most likely a bug in the internal `as.igraph.vs()` function at the end of `iterators.R`. The relevant part of the code looks like this:

```
if (is.logical(v)) {
  res <- as.vector(V(graph))[v]
} else if (is.numeric(v) && any(v<0)){
  res <- as.vector(V(graph))[v]
} else {
  res <- as.numeric(v)
}
```

`numeric()` fails the first two branches so it is treated as `res <- as.numeric(numeric())`, yielding `numeric(0)`, i.e. a zero-length vector. The code that calls `as.igraph.vs()` then takes the first element of the vector and decreases it by 1.

I guess what we would need is an additional check like this:

```
if (length(res) == 0) {
  stop("No vertex name was given")
}
```

A similar error exists in `as.igraph.es` as well.

szhorvat:
Should this even use `as.igraph.vs`?  This function takes a single vertex ID, not a vertex selector.

https://igraph.org/c/doc/igraph-Basic.html#igraph_neighbors

https://github.com/igraph/rigraph/blob/a14fc50abf9bf3ae17b870ccbc21f88d0eefed6e/tools/stimulus/rinterface.c.in#L3687

I would imagine (but I don't know the details) that the line ` vid=REAL(pvid)[0];` interprets the input as a vector of reals and takes the first element. But `numeric(0)` is a zero-length vector, so I assume this accesses memory it shouldn't.

It's similar in #140

ntamas:
[...]

PR #712, opened 2023-03-16
chore: Change stimulus to generate functions with `_impl` suffix

krlmlr:
This allows reenabling autogenerated wrappers, step by step.


szhorvat:
In case you missed it, I just wanted to point out the importance of making sure that the generated glue code would validate scalars, i.e. ensure that they are indeed length-1 vectors. https://github.com/igraph/rigraph/issues/699#issuecomment-1464543700

krlmlr:
I'm doing baby steps aka small wins -- this is meant as a first step, we can address asserting length later (which will also uncover errors in downstream packages...)

R/zzz-deprecate.R
```diff
@@ -586,3 +586,29 @@ deprecated("igraph.shape.noclip", shape_noclip)
 deprecated("igraph.shape.noplot", shape_noplot)
 #' @export create.communities
 deprecated("create.communities", make_clusters)
+
+
+#' @export
+graph.count.isomorphisms.vf2 <- graph.count.isomorphisms.vf2_impl
```
maelle, line 592: @krlmlr could you do these calls in another script? I'm planning to delete zzz-deprecate. Otherwise I'll adapt my script, as you prefer.

szhorvat:
Yes, you're right.

We should open a separate issue so we don't forget.

The important case is checking for a length of zero, as it can lead to a crash, as in #153. Having a length larger than 1 is not a serious issue—it just takes the first value.

R/zzz-deprecate.R
krlmlr: Thanks, done.

Issue #714, opened 2023-03-16
special case the lifecycle transformation of igraph.arpack.default


maelle:
since at the moment it's a list not a function

maelle:
same for igraph.eigen.default

maelle:
it's a bit of a challenge as if I make it a function, it's a breaking change. @krlmlr any suggestion?

maelle:
@krlmlr any idea? I'm a bit stuck as to how to deprecate _data_.

maelle:
I found an idea via https://stackoverflow.com/a/75192818/5489251

https://github.com/allisonhorst/palmerpenguins/pull/25

I'll draft something

maelle:
but, actually, igraph.arpack.default isn't exported?!

maelle:
I mean eigen_defaults isn't.

maelle:
And since these two lists are not data, the delayedAssign idea doesn't work. :thinking:

maelle:
maybe I'm making this more complicated than it should, although the two deprecated names were exported, they're maybe not used a lot?

ntamas:
They are not used frequently; they represent default parameters to the underlying ARPACK eigensolver and most users won't mess around with the default settings. Nevertheless, `eigen_defaults` should be the new name that is exported and `igraph.eigen.default` is the one that should be deprecated. If `igraph.arpack.defaults` already exists as `arpack_defaults`, then the latter one should be the preferred one, similarly to `eigen_defaults`.

krlmlr:
We can't easily make it assignable.

krlmlr:
The tidyverse has very detailed opinions on defaults, in particular <https://design.tidyverse.org/def-user.html> (incomplete) and the sections before that.

Issue #713, opened 2023-03-16
why is as_phylo.igraphHRG not exported


maelle:
I'm working on an automated script for #708 and noticed this.

If the function isn't exported, the lifecycle stuff isn't so important.

ntamas:
As far as I know this function simply adds support for igraph's `igraphHRG` objects to be used in conjunction with the `as_phylo()` generic function so you can convert them to `phylo` objects from the `ape` package. Although I'm a bit confused now because `ape` seems to use `as.phylo` and not `as_phylo`... maybe we've accidentally renamed a dotted function that we shouldn't have?

krlmlr:
There never was an `as.phylo()` in the codebase.

We need delayed method registration to switch to `as.phylo()` because ape is only suggested, not imported.

maelle:
But does this mean we can simply remove `asPhylo.igraphHRG()` (that was deprecated to `as_phylo.igraphHRG()` in `zzz-deprecate.R`)?

ntamas:
Hmmm, it may mean that `as.phylo()` in igraph has fallen victim to the mass renaming of dotted methods to underscored methods in the past, and we should reverse the renaming.

maelle:
ok, so in community.R? could that be done independently from #716? But what to do about `as_phylo.igraphHRG()` in that PR?

krlmlr:
We should implement `as.phylo()` methods and forward our `as_phylo()` to `as.phylo()` (with a deprecation warning). I seriously doubt that anyone has implemented `as_phylo()` methods: https://github.com/search?q=org%3Acran%20as_phylo.&type=code .

krlmlr:
For implementing an S3 method for a suggested package, drop https://github.com/r-lib/rlang/blob/main/R/standalone-s3-register.R into the package and use the functions provided there.

krlmlr:
A separate small PR seems a good way here.

PR #716, opened 2023-03-16
docs: use lifecycle instead of zzz-deprecate.R

maelle:


inst/deprecate.R
```diff
@@ -0,0 +1,131 @@
...
+get_title <- function(fn_name) {
+  rd_href <- pkgdown:::get_rd_from_help("igraph", fn_name)
+  pkgdown:::extract_title(rd_href)
+}
+
+# treat calls ----
+treat_call <- function(old, new) {
+
+  if (old %in% c("igraph.eigen.default", "igraph.arpack.default", "graph.famous")) {
+    return()
```
maelle, line 92: they will be treated separately, manually, after we agree on the output of the automatic part. I opened issues to not forget.

krlmlr:
I have another PR in the works, can we please merge that first? This looks like repeated application of a script?

inst/deprecate.R
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  pkg_defs <- parse_package_defs()
+  template <- paste(readLines(here::here("inst", "deprecate-template.txt")), collapse = "\n")
+  relevant_row <- pkg_defs[pkg_defs[["name"]] == new,]
+  if (nrow(relevant_row) > 1) {
+    relevant_row <- relevant_row[!grepl("aaa-auto", relevant_row[["script_name"]]),]
```
maelle, line 99: for quite a few functions there are two definitions

maelle:
> I have another PR in the works, can we please merge that first? This looks like repeated application of a script?

I can rebase this PR once you have merged your PR? 

I might still do a commit that shows what running the script does, but I could undo it, then rebase, then run it. Does that sound good?

maelle:
Example (the script hasn't finished running but my time is up): https://github.com/maelle/rigraph/pull/2

krlmlr:
Merged my PR now, but looks like there are no conflicts?

maelle:
> Merged my PR now, but looks like there are no conflicts? 

Right but there would be, once I run the script :grin:

maelle:
Notes of the functions that are not "properly" documented. For instance `edge_disjoint_paths()` has no own manual page. It's only in https://r.igraph.org/reference/edge_connectivity.html?q=edge_disjoint_paths#ref-examples because of aliases. https://github.com/igraph/rigraph/blob/1cee91414ce2d9b02c9362dcf44b5f3f8146b7b7/R/flow.R#L281

So for the old version of that function, I can't inherit the parameter definition since there is no parameter definition. I can't use an alias for the function because if I do so, it gets no manual page of its own so no lifecycle badge and explanation.

I will explore these two ideas (not this week)
- do more work in my script to extract argument definition via aliases (thanks to `pkgdown::as_pkgdown()`).
- use uninformative parameter documentation such as `#' @param graph see ?edge_disjoint_paths`

ntamas:
In this particular case, `edge_disjoint_paths()`, `adhesion()` and `edge_connectivity()` are different names for the same graph concept: how many edge-disjoint paths there are between a given pair of vertices (or, if the pair is not specified, between _any_ pair of vertices in the input graph)? `edge_disjoint_paths()` is a bit of a misnomer because I would expect such a function to return the _paths_ themselves and not only the _number_ of paths.

Personally, if I were to design the API today from scratch, I'd only keep `edge_connectivity()` (and maybe `adhesion()` as an alias - this name is not used very commonly in graph theory but I believe it's used by social scientists, who frequently have their own names for common graph theory concepts). `edge_disjoint_paths()` may be used some time later in the future to implement a function that returns the paths themselves, but we don't have such a function in the C core yet.

(To add insult to injury, the C core has separate functions: `igraph_edge_disjoint_paths()` returns the _number_ of edge disjoint paths between a given pair of vertices, while `igraph_edge_connectivity()` returns it for all source-target pairs. The R function decides to call one or another based on whether `source` and `target` are given or not).

ntamas:
I think that this deprecation should point to `edge_connectivity()` (so, function B).

maelle:
preview https://github.com/maelle/rigraph/pull/4/files

I still need to fix #714

maelle:
interesting bug: I can't run this code if I run `load_all()` first, as it prevents `pkgdown:::get_rd_from_help("igraph", fn_name)` from working :exploding_head:

krlmlr:
Do I understand correctly that we need to run the `inst/deprecate.R` script after merging this?

Should the script and the template live in `inst/` instead of `tools/` because the script isn't necessary for the installed package?

maelle:
No, I will run it before we merge. I even think we might want to delete it but I can move it in the meantime. If all goes well, this introduces a bunch of soft deprecations. A future version of the script would convert all of them to warnings at the next stage for instance.

In any case #713 and #714 will need to be fixed first (they're special cases that I moved out of this "automatic" PR).

krlmlr:
Blocked by upstream issues.

maelle:
This is blocked by https://github.com/igraph/rigraph/issues/714 (then I'll need to re-run the code to check what needs to be tweaked)

krlmlr:
Should be unblocked once the PRs are merged.

maelle:
Now blocked by #899 (will do a PR next week), since at least one function that will be renamed is a function that is "deprecated to" in this PR. :sweat_smile:

krlmlr:
Running the script now. We can tackle the ivs stuff later?

krlmlr:
What does "check there are no stray aliases" mean? I'll review the documentation changes in this PR closely, anything else needed?

From my perspective, this is the last PR before I can finally run the revdepchecks for the 1.6.0 release candidate.

krlmlr:
@maelle: can you please fix the check failures? I'll run the revdepchecks now, and another time after this is merged. This will give a better signal too.

Do we need to redistribute the code now in `aaa-NOT-auto.R` ? Too bad to call it `zzz-deprecate.R`, like the old script?

Please help me understand why you think the ivs (maxcohesion?) things should come first. Can't we add the deprecation later, manually? Eventually, we'll need to be able to do it manually anyway.

maelle:
I will start a fresh PR from main then run the script after 

- running R CMD check on this branch to catch errors
- changing the version in the lifecycle text to 1.6.0.

maelle:
"Rd files with duplicated alias" all over the place.

maelle:
apart from the alias situation, tests will need to be updated.

maelle:
a function is deprecated to `make_famous_graph()` that is not exported.

maelle:
>   Note: possible error in 'shapes(shape = shape, ': unused arguments (clip = clip, plot = plot, parameters = parameters) 
  Note: possible error in 'triangles(graph = graph, ': unused argument (vids = vids) 
  Note: possible error in 'sample_pref(nodes = nodes, ': unused argument (type.dist.matrix = type.dist.matrix) 
  Note: possible error in 'curve_multiple(graph = graph, ': unused argument (x = x) 
  Note: possible error in 'distance_table(graph = graph, ': unused arguments (weights = weights, unconnected = unconnected, details = details) 
  Note: possible error in 'cohesive_blocks(blocks = blocks, ': unused argument (blocks = blocks) 
  Note: possible error in 'sample_last_cit(n = n, ': unused arguments (types = types, attr = attr) 
  Note: possible error in 'membership(communities = communities, ': unused arguments (no = no, steps = steps) 
  Note: possible error in 'betweenness(graph = graph, ': unused argument (e = e) 
  Note: possible error in 'sample_traits_callaway(nodes = nodes, ': unused argument (k = k) 
  Note: possible error in 'cohesive_blocks(blocks = blocks, ': unused arguments (blocks = blocks, file = file, project.file = project.file) 
  Note: possible error in 'as_adj_list(graph = graph, ': unused argument (.x = .x) 
  Note: possible error in 'distance_table(graph = graph, ': unused arguments (from = from, to = to, mode = mode, weights = weights) 
  Note: possible error in 'distance_table(graph = graph, ': unused arguments (from = from, to = to, mode = mode, weights = weights, output = output, predecessors = predecessors, inbound.edges = inbound.edges, algorithm = algorithm, x = x, x = x) 
  Note: possible error in 'graph_from_adj_list(adjlist = adjlist, ': unused argument (x = x) 
  Note: possible error in 'is_bipartite(types = types, ': unused arguments (types = types, edges = edges, directed = directed) 
  Note: possible error in 'as_data_frame(d = d, ': unused arguments (d = d, directed = directed, vertices = vertices) 
  Note: possible error in 'connect(graph = graph, ': unused arguments (nodes = nodes, mindist = mindist) 
  Note: possible error in 'graphlet_basis(graph = graph, ': unused arguments (cliques = cliques, niter = niter, Mu = Mu) 
  Note: possible error in 'graph_from_graphnel(graphNEL = graphNEL, ': unused arguments (n = n, l = l) 
  Note: possible error in 'shapes(coords = coords, ': unused arguments (coords = coords, el = el, params = params, end = end) 
  Note: possible error in 'shapes(coords = coords, ': unused arguments (coords = coords, v = v, params = params) 
  Note: possible error in 'as_graphnel(graph = graph, ': unused arguments (x = x, x = x, y = y) 
  Note: possible error in 'subgraph(graph = graph, ': unused argument (impl = impl) 
  Note: possible error in 'cohesive_blocks(blocks = blocks)': unused argument (blocks = blocks) 
  Note: possible error in 'cliques(graph = graph, ': unused arguments (subset = subset, file = file, x = x) 
  Note: possible error in 'cliques(graph = graph, ': unused argument (subset = subset) 
  Note: possible error in 'is_matching(graph = graph, ': unused arguments (weights = weights, eps = eps) 
  Note: possible error in 'distance_table(graph = graph, ': unused arguments (v = v, to = to, mode = mode, weights = weights, algorithm = algorithm)

maelle:
this is wrong

```
average.path.length <- function(graph , weights = NULL , directed = TRUE , unconnected = TRUE , details = FALSE) { # nocov start
   lifecycle::deprecate_soft("1.6.0", "average.path.length()", "distance_table()")
   distance_table(graph = graph, weights = weights, directed = directed, unconnected = unconnected, details = details)
} # nocov end
```

maelle:
probably a regex I wrote wrong :eyes:

maelle:
in shapes.Rd, aliases are used for things that are not aliases. :thinking:

maelle:
For things such as  `edge_connectivity()` I'll update zzz-deprecate.R manually, and I'll stop using aliases as they don't make any sense for this use case.

krlmlr:
The methods that this commit removes are never called because the `"dendPlot"` generic doesn't exist. The `igraph::dendPlot()` function has `UseMethod("plot_dendrogram")`, this is what is used for dispatch.

maelle:
> Please help me understand why you think the ivs (maxcohesion?) things should come first. Can't we add the deprecation later, manually? Eventually, we'll need to be able to do it manually anyway.

Yes, ok.

PR #721, opened 2023-03-18
fix: `as.hclust(hrg.fit(g))` works again

krlmlr:
For #720.


krlmlr:
@Antonov548: can you please propose a change that uses `.Call()` instead of `.C()` so that we can later integrate in #720?

krlmlr:
This probably needs documentation and an example too, CC @maelle.

Nobody ever complained (on the issue tracker) that this doesn't work.

ntamas (approved):
Looks good to me. I leave the merging up to you.

krlmlr:
@maelle: Can you please handle examples and documentation at some point?

PR #723, opened 2023-03-18
chore: Move implementation from `rinterface.c.in` to `rinterface_extra.c`

krlmlr:
I like the line stats of this PR.

PR #725, opened 2023-03-18
chore: Convert all calls to the `igraph_*()` API to one-liners

krlmlr:
To simplify adding error handling code everywhere.

Stimulus generates one-liners only.


krlmlr:
Waiting for the equivalent of #723 which I had to revert for compatibility with #720.

PR #729, opened 2023-03-19
chore: Define and use `IGRAPH_R_CHECK()`

### LINKED ISSUE #726: Define and use new `IGRAPH_R_CHECK()` macro for all API calls
After #725.

This macro would call `Rf_error()` if an error is detected.

Code that creates a regular expression matching all API function calls:

```r
yaml <- yaml::read_yaml("cigraph/interfaces/functions.yaml")
function_names <- names(yaml)

function_name_rx <- paste(rex::escape(function_names), collapse = "|")
rx <- paste0(" (", function_name_rx, ")[(].*[)];$")

clipr::write_clip(rx)
```

Example (shortened):

```
  (igraph_empty|igraph_add_edges|igraph_add_vertices|...)[(].*[)];$
```

krlmlr:
Closes #726.

### DIFF
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -27,3 +27,5 @@
 #define R_IGRAPH_VERSION_VAR ".__igraph_version__."
 
 SEXP R_igraph_add_env(SEXP graph);
+
+#define IGRAPH_R_CHECK(x) do { x; } while (0)
--- a/src/rinterface.c
+++ b/src/rinterface.c
@@ -137,7 +137,7 @@ SEXP R_igraph_empty(SEXP n, SEXP directed) {
   c_n=INTEGER(n)[0];
   c_directed=LOGICAL(directed)[0];
                                         /* Call igraph */
-  igraph_empty(&c_graph, c_n, c_directed);
+  IGRAPH_R_CHECK(igraph_empty(&c_graph, c_n, c_directed));
 
                                         /* Convert output */
   IGRAPH_FINALLY(igraph_destroy, &c_graph);
@@ -187,7 +187,7 @@ SEXP R_igraph_full_citation(SEXP n, SEXP directed) {
   c_n=INTEGER(n)[0];
   c_directed=LOGICAL(directed)[0];
                                         /* Call igraph */
-  igraph_full_citation(&c_graph, c_n, c_directed);
+  IGRAPH_R_CHECK(igraph_full_citation(&c_graph, c_n, c_directed));
 
                                         /* Convert output */
   IGRAPH_FINALLY(igraph_destroy, &c_graph);
@@ -217,7 +217,7 @@ SEXP R_igraph_extended_chordal_ring(SEXP nodes, SEXP W, SEXP directed) {
   R_SEXP_to_matrix(W, &c_W);
   c_directed=LOGICAL(directed)[0];
                                         /* Call igraph */
-  igraph_extended_chordal_ring(&c_graph, c_nodes, &c_W, c_directed);
+  IGRAPH_R_CHECK(igraph_extended_chordal_ring(&c_graph, c_nodes, &c_W, c_directed));
 
                                         /* Convert output */
   IGRAPH_FINALLY(igraph_destroy, &c_graph);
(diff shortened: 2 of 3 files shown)


krlmlr:
@Antonov548: Can you please continue here and commit directly to this branch? The goal is to have the new macro check for an igraph error, and call `Rf_error()` in this case. We can then get rid of the igraph error handler.

krlmlr:
We also need to patch Stimulus to generate `IGRAPH_R_CHECK()` calls.

src/rinterface.h
```diff
@@ -27,3 +27,10 @@
 #define R_IGRAPH_VERSION_VAR ".__igraph_version__."
 
 SEXP R_igraph_add_env(SEXP graph);
+
+#define IGRAPH_R_CHECK(func) \
+    do { \
+        if (0) {igraph_error_type_t __c = func; } \
```
krlmlr, line 33: What do we need the `if (0)` for? Can you please add a comment to explain?

krlmlr (commented):
Thanks, looks great!

szhorvat:
Fatal error are entirely separate from normal error and use a different handler / different mechanism. They are used in two cases: unrecoverable errors and assertion failures. Essentially, these are error that cannot or should not be handled.

Normally, the process would crash under such an error. But there is a handler that makes it possible to be a bit more polite to the user, and at least try to show a nice message and return to R top level. However, generally speaking, it is not safe for the user to continue in such a situation. The message should say something to the effect: "Save your work NOW and restart R. Also, this is a bug, please report steps to repro it."

When such a a serious error occurs, resources will not be freed and the internal state may have become inconsistent, so continuing is dangerous. Longjumping to the top level is just a courtesy and a last minute attempt to try to let the user save their work.

src/rinterface.h
szhorvat: Remove this from here. In the C core this looks different and is essentially for a static compiler check, will explain when I get back next week.

ntamas:
> The R_igraph_warning_handler is dangerous because Rf_warning() aka warning() can also longjmp (with options(warn = 2) set in the R code)

I'm having similar problems in the Python interface where the user can also decide to turn igraph warnings into errors and then I end up throwing exceptions in places where Python does not check for them.

> How can we handle the reason in R_igraph_error_handler()?

I guess that it should be copied to a pre-allocated error message buffer (in case igraph frees its own buffer that holds the reason by the time we want to print it), and then it should be printed later in the call to `Rferror()`.

ntamas:
> Can we then use stimulus to add macro IGRAPH_R_CHECK for all function that are returns igraph_error_t in rinterface?

Yes; `stimulus` defines the code snippets that it uses in `types-RC.yaml` so you can modify those to add the required `IGRAPH_R_CHECK()` calls.

Note that there are several helper functions such as `R_SEXP_to_igraph_copy()` that _call_ igraph functions but do not return an `igraph_error_t` on their own _yet_ (because they also assume that R will longjmp if there is an error). These will have to be modified as well.

Antonov548:
@ntamas I have added buffer for `warnings` message and `warnings` are showing only after call stack is unwinding. Since warning also can produce `longjump`. But with this approach only first warning message is saved and showed at once.

Should we have some dynamic array for warnings or proceed with current approach?

Also does `fatal_handler` should have same behavior as `error_handler`? (unwind stack - show error).

ntamas:
I think it's okay to keep one warning only. We don't use warnings extensively in the C core and most of the time you'll only see one warning anyway.

As for fatal vs normal errors: I think @szhorvat explained the difference somewhere but I can't find where. The difference is that fatal errors indicate a condition where igraph's internal data structures are mostly in an inconsistent state and the safest course of action for the user would be to save whatever they were doing, quit R and start it again. After a fatal error, we make no guarantees about igraph's future behaviour. Normal errors are totally "safe" in the sense that igraph's data structures are still consistent and the user can continue with their work.

Antonov548:
@ntamas Should it be fine then to leave for fatal error just `error` which will terminate session immediately with long jump? And for normal error's have this feature that print's error only after call stack is unwounded.

krlmlr:
Thanks! I believe we can treat fatal errors the same way as errors, with a slightly different error message from the R side. @Antonov548: Can you please open a new PR for that? Do we need an issue?

PR #741, opened 2023-03-30
refactor!: Remove `igraph.eigen.default()` and `eigen_defaults`, introduce internal `eigen_defaults()` as a function

maelle:
Easy case from the two lists in #714


R/zzz-deprecate.R
```diff
@@ -321,8 +321,6 @@ deprecated("hub.score", hub_score)
 deprecated("igraph.arpack.default", arpack_defaults)
 #' @export igraph.console
 deprecated("igraph.console", console)
-#' @export igraph.eigen.default
```
maelle, line 324: it's problematic that this was exported. I'm not sure it was used though?!

krlmlr:
Thanks!

PR #744, opened 2023-03-30
chore: fatal error with `IGRAPH_R_CHECK()`

Antonov548:


szhorvat:
Why does this need to be changed? A fatal error handler is expected precisely to do either a longjump or abort the process (which we don't do for R).

A fatal error can't be handled by definition. It is explicitly illegal for the fatal error handler to return.

szhorvat:
OK, let me be explicit then: This is incorrect and must not be merged. The current version of the fatal error handler is fine.

The purpose of fatal error "handling" is to avert a crash at the last minute, and try tobehave just a little bit better. But we still can't even guarantee not messing up the process state, so the user is advised to restart R after a fatal error occurs.

Consider e.g. this:

```
IGRAPH_ASSERT(ptr != NULL);
fun(*ptr);
```

The `IGRAPH_ASSERT` can trigger a fatal error. What if the fatal error handler returns? Then we get a crash on the second line, defeating the purpose of the fatal error handling mechanism

szhorvat:
I'm still quite sick and won't be commenting much. I'm not even sure I can make it for the meeting tomorrow :-(

PR #750, opened 2023-03-31
fix: Use longjmp by default in functions called from the top level

krlmlr:
This restores the behavior of the current CRAN version. It's also somewhat safe to longjmp if we're not very deep into the call stack.

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -2435,14 +2435,15 @@ void R_igraph_error_handler(const char *reason, const char *file,
    * IGRAPH_FINALLY_FREE() because 'reason' might be allocated on the heap and
    * IGRAPH_FINALLY_FREE() can then clean it up. */
 
-  if (R_igraph_errors_count == 0 || R_igraph_in_r_check != 0) {
+  if (R_igraph_errors_count == 0 || R_igraph_in_r_check == 0) {
     snprintf(R_igraph_error_reason, sizeof(R_igraph_error_reason),
       "At %s:%i : %s%s %s", file, line, reason,
       maybe_add_punctuation(reason, ","),
       igraph_strerror(igraph_errno));
     R_igraph_error_reason[sizeof(R_igraph_error_reason) - 1] = 0;
 
-    if (R_igraph_in_r_check != 0) {
+    if (R_igraph_in_r_check == 0) {
+      IGRAPH_FINALLY_FREE();
       R_igraph_error();
     }
   }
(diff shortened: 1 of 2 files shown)


krlmlr:
@Antonov548: Why does this fail with oldrel-4? Install with:

```bash
rig add 3.5
rig system make-links
rig system setup-user-lib 3.5
rig system add-pak --pak-version devel 3.5
```

Run with `R-3.5` instead of `R` .

src/rinterface_extra.c
```diff
@@ -2429,12 +2435,16 @@ void R_igraph_error_handler(const char *reason, const char *file,
    * IGRAPH_FINALLY_FREE() because 'reason' might be allocated on the heap and
    * IGRAPH_FINALLY_FREE() can then clean it up. */
 
-  if (R_igraph_errors_count == 0) {
+  if (R_igraph_errors_count == 0 || R_igraph_in_r_check != 0) {
```
Antonov548, line 2438: @krlmlr shouldn't be here `if (R_igraph_errors_count == 0 || R_igraph_in_r_check == 0) {` ?

`0` means that error was accrued not in context of `R_IGRAPH_CHECK` and we should print error anyway.

krlmlr: Good catch!

krlmlr:
Thanks!

krlmlr:
@ntamas: Your feedback is still appreciated.

ntamas:
> What's a good test for an allocation error, an igraph call that allocates that much memory that any OS will immediately say "nope"?

Hmmm... users mostly stumble upon out-of-memory errors when trying to calculate the shortest path matrix of a huge graph. So, generate a graph with 1 million vertices or so (I usually use `sample_grg()` as a testbed; 100 vertices with a radius of 0.2 or 10000 vertices with a radius of 0.02 or 1000000 vertices with a radius of 0.002), then try to invoke `distances()` on it. Most likely it will throw an OOM right when it tries to allocate the result matrix, although on Linux the OS tends to overcommit to the memory requirement in the hope that the memory won't be needed after all, so you might have to wait a bit until it really crashes.

ntamas:
> Your feedback is still appreciated.

Looks OK to me.

src/rinterface_extra.c
```diff
@@ -2380,6 +2380,12 @@ static int R_igraph_errors_count = 0;
 static char R_igraph_warning_reason[4096];
 static int R_igraph_warnings_count = 0;
 
+static int R_igraph_in_r_check = 0;
```
szhorvat, line 2383: Nitpick: since this seems to be affectively a boolean, why not use a C99 `bool` here? If you do, `#include <stdbool>`.

szhorvat:
This should be undone eventually though, as it's really an unsafe hack. It'd be good to add comments about such hacks, not to get confused about them in the future.

PR #751, opened 2023-03-31
feat: Avoid longjmp for interrupts

### LINKED ISSUE #746: Make sure that interruption works without a long jump
Do we need to return `IGRAPH_INTERRUPTED` instead of `R_interrupts_pending = 1` and `R_CheckUserInterrupt()` ?

krlmlr:
in preparation for igraph/C 0.10.

Closes #746.

### DIFF
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -30,10 +30,12 @@ SEXP R_igraph_add_env(SEXP graph);
 
 void R_igraph_error();
 void R_igraph_warning();
+void R_igraph_interrupt();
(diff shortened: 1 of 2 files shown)


krlmlr:
@ntamas: Please let me know if any of this is murky. I tested interactively, the new behavior is much better than the existing behavior. We can still forward this condition to R as an interrupt if it later turns out to be useful/important.

ntamas:
Seems okay to me, thanks a lot!

PR #754, opened 2023-03-31
chore: Use true vertex names for printing vertex sets. If a vertex set captures a relationship between vertices (e.g., the `father` component of `bfs()`), the vertex set is printed as a named vector

### LINKED ISSUE #227: bfs might have a bug when using named vertices
Hi I do not understand why the output is as such. There is a possible bug in bfs (or I might not understand correctly)

tree<-graph_from_literal(3-+5:6,6-+2,3-+4)
obj<-graph.bfs(tree,root=3, neimode="out",father=TRUE,order=TRUE,unreachable = FALSE)
Observe that the output is obj$order = 6 2 and obj$father= 3 5 6 2 4. But I expect it to be obj$order = 3 5 6 4 2 and obj$father= NA 3 3 3 2

### LINKED ISSUE #228: bfs might have a bug
_From @tintinthong on September 20, 2017 17:54_

Hi I do not understand why the output is as such. There is a possible bug in bfs (or I might not understand correctly)

```
tree<-graph_from_literal(3-+5:6,6-+2,3-+4)
obj<-graph.bfs(tree,root=3, neimode="out",father=TRUE,order=TRUE,unreachable = FALSE)
```
Observe that the output is obj$order = 6 2 <NA> <NA> <NA> and obj$father= 3 5 6 2 4. But I expect it to be obj$order = 3 5 6 4 2 and obj$father= NA 3 3 3 2

_Copied from original issue: igraph/igraph#1032_

krlmlr:
Closes #227. Closes #228.

### DIFF
--- a/R/iterators.R
+++ b/R/iterators.R
@@ -1262,13 +1262,18 @@ print.igraph.vs <- function(x,
                             id = igraph_opt("print.id"),
                             ...) {
   graph <- get_vs_graph(x)
+  if (!is.null(graph)) {
+    vertices <- V(graph)
+  } else {
+    vertices <- NULL
+  }
   len <- length(x)
   gid <- graph_id(x)
 
   title <- "+ " %+% chr(len) %+% "/" %+%
-    (if (is.null(graph)) "?" else chr(vcount(graph))) %+%
+    (if (is.null(vertices)) "?" else chr(length(vertices))) %+%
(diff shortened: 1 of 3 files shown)


ntamas:
Looks good to me!

ntamas (approved):

PR #193, opened 2017-05-15
Factor support

thomasp85:
Fixes #34

In order to avoid conversion of factors attribute elements are
initialised prior to assignment if the attribute does not exist or if
the assignment is overwriting all elements within that attribute

PR #758, opened 2023-04-02
feat!: Accept numeric edge IDs

### LINKED ISSUE #223: Better cooperation of as_data_frame() and graph_from_data_frame() ?
It would be great to have the following functionality back:

```r
dlist <- as_data_frame(g, what="both")
graph_from_data_frame(dlist$edges, vertices=dlist$vertices)
```

In other words, using (parts of) output of `as_data_frame` in `graph_from_data_frame` directly. I recall it was possible before recent API changes. Above will work if `g` has vertex `name`s, but it won't if it does not. See the following reprex:

``` r
library(igraph)
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
g <- make_graph(~ A, B -- C, C -- D)
V(g)$a <- letters[1:4]
g1 <- delete_vertex_attr(g, "name")

# 1
dlist <- as_data_frame(g, what="both")
# Works if `g` has names
r <- graph_from_data_frame(dlist$edges, vertices=dlist$vertices, directed=is_directed(g))
identical_graphs(g, r) # Why?
#> [1] FALSE
identical(dlist, as_data_frame(r, what="both"))
#> [1] TRUE

# 2
dlist1 <- as_data_frame(g1, what="both")
# Won't work if `g` does not have names
graph_from_data_frame(dlist1$edges, vertices=dlist1$vertices, directed=is_directed(g1))
#> Error in graph_from_data_frame(dlist1$edges, vertices = dlist1$vertices, : Some vertex names in edge list are not listed in vertex data frame
```

In particular:

1. I think the main reason that this connection is lot is that `as_data_frame( . , what="vertices")` does not include the vertex IDs as the first column anymore.
2. I don't understand why `g` and `r` are not `identical_graphs` in `# 1`. Both edge and vertex data frames are identical. Am I missing something?

krlmlr:
Closes #223.

This needs a fix of the tests, and a revdepcheck to assess the impact.

### DIFF
--- a/R/data_frame.R
+++ b/R/data_frame.R
@@ -71,16 +71,17 @@
 #'
 #' @aliases graph_from_data_frame graph.data.frame as_data_frame get.data.frame
 #' @param d A data frame containing a symbolic edge list in the first two
-#'   columns. Additional columns are considered as edge attributes.  Since
+#'   columns, as vertex names or vertex IDs.
+#    Additional columns are considered as edge attributes.  Since
 #'   version 0.7 this argument is coerced to a data frame with
-#'   `as.data.frame`.
+#'   [as.data.frame()].
 #' @param directed Logical scalar, whether or not to create a directed graph.
 #' @param vertices A data frame with vertex metadata, or `NULL`. See
 #'   details below. Since version 0.7 this argument is coerced to a data frame
-#'   with `as.data.frame`, if not `NULL`.
+#'   with [as.data.frame()], if not `NULL`.
 #' @return An igraph graph object for `graph_from_data_frame()`, and either a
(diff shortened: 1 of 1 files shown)


krlmlr:
@thomasp85: Are you still interested in #193? I noticed I'm making an effort here to convert factors to characters, perhaps we should be doing things the other way round? What behavior would you like to see -- setting and retrieving attributes of type factor?

krlmlr:
Running revdepchecks now.

krlmlr:
The impact is severe, but I'll also run with the current main branch to compare.

krlmlr:
Actually, rerunning one more time with this branch and the broken packages.

PR #762, opened 2023-04-03
refactor!: kill igraph.arpack.default and arpack_defaults lists, add arpack_defaults() as a function

### LINKED ISSUE #714: special case the lifecycle transformation of igraph.arpack.default

maelle:
Fix #714 

@krlmlr actually I see [usage of the old name in reverse dependencies](https://github.com/search?q=org%3Acran+arpack_defaults&type=code), should I
- cancel the transformation of arpack_defaults to a function
- or make PRs to those packages

Whilst replacing occurrences of the old name, I noticed the opportunity to use `modifyList()`, so this PR does that too but in a different commit. :innocent:


krlmlr:
We could allow a function in the places where `arpack_default` is used?

maelle:
> We could allow a function in the places where arpack_default is used?

Could you please elaborate? 

In my [latest commit](https://github.com/igraph/rigraph/pull/762/commits/ab5fb8882756b8839881e05485bf3f8c97372f2c) I added a maybe too complex way to deprecate the usage of `arpack_default` for the `options` argument
- the argument is evaluated;
- if it's called "arpack_default" and is a function (it is a function from igraph, and remains a function unless the user creates an object named so), we replace it with `arpack_defaults()` **after showing a deprecation message**.

R/centrality.R
```diff
@@ -499,27 +500,38 @@ arpack_defaults <- list(
...
                    env = parent.frame(), complex = !sym) {
+
+  eval_try <- rlang::eval_tidy(options)
+  options_value <- rlang::call_args(rlang::current_call())[["options"]]
+  if (is(eval_try, "function") && as.character(options_value) == "arpack_defaults") {
+    lifecycle::deprecate_soft(
+      "1.5.0",
+      I("arpack_defaults"),
+      "arpack_defaults()",
+      details = "So a function, not an object."
```
krlmlr, line 513: I don't understand this wording.

R/centrality.R
```diff
@@ -499,27 +500,38 @@ arpack_defaults <- list(
 #' }
 #' @family arpack
 #' @export
-arpack <- function(func, extra = NULL, sym = FALSE, options = arpack_defaults,
+arpack <- function(func, extra = NULL, sym = FALSE, options = arpack_defaults(),
                    env = parent.frame(), complex = !sym) {
+
+  eval_try <- rlang::eval_tidy(options)
```
krlmlr, line 506: What do we need `eval_tidy` for? Can't we just force, and if the user has passed a function, we call it?

maelle: We only need to catch the case where the argument is `arpack_defaults` and a function. I'm using `rlang::call_args(rlang::current_call())[["options"]]` on the following line hence the rlang here, especially as rlang is a dependency.

Now, maybe this whole "hack" is a bad idea.

krlmlr (approved):
Looks good! Can you please merge when the checks pass?

maelle:
actually no, I need to expand the hack to all functions here.

maelle:
actually with ensure_igraph() on the main branch, I'll just close and start again.

PR #763, opened 2023-04-03
refactor!: Define methods for `ape::as.phylo()` instead of own `as_phylo()` generic, remove deprecated `asPhylo()`

### LINKED ISSUE #713: why is as_phylo.igraphHRG not exported

maelle:
Fix #713 

Not really breaking since it's still a soft deprecation.


NAMESPACE
```diff
@@ -139,8 +141,8 @@ export(articulation.points)
 export(articulation_points)
 export(as.directed)
 export(as.igraph)
+export(as.phylo)
 export(as.undirected)
-export(asPhylo)
```
maelle, line 143: this one I really killed.

maelle: Which seems fine given the occurrences of the same name in https://github.com/search?q=org%3Acran+asPhylo+&type=code are not from this igraph method.

R/community.R
```diff
@@ -696,15 +696,15 @@ as.hclust.communities <- function(x, hang = -1, use.modularity = FALSE,
 #' @rdname communities
 #' @family community
 #' @export
-as_phylo <- function(x, ...) {
-  UseMethod("as_phylo")
+as.phylo <- function(x, ...) {
```
krlmlr, line 699: There is already `ape::as.phylo()`, do we really need to do anything here?

R/community.R
```diff
@@ -696,15 +696,15 @@ as.hclust.communities <- function(x, hang = -1, use.modularity = FALSE,
...
+  UseMethod("as.phylo")
 }
 
 #' @rdname communities
-#' @method as_phylo communities
+#' @method as.phylo communities
```
krlmlr, line 704: Do we need to use `s3_register()` in `.onLoad()` here?

maelle: I'm a bit confused (and not very experienced with S3). That's what you suggested and what I added in `R/aaa-utils.R`, was that wrong?

R/community.R
maelle: Should this just be deleted? As noted in the other comment I'm now confused.

R/aaa-utils.R
```diff
@@ -0,0 +1,3 @@
+.onLoad <- function(...) {
```
krlmlr, line 1: Should we do it the rlang way? Should probably live in `R/zzz.R` then.

https://rlang.r-lib.org/reference/on_load.html?q=onload#comparison-with-onload-

R/community.R
krlmlr: Yes, I guess we don't need it here.

R/community.R
krlmlr: We don't need `#' @method`, I believe.

krlmlr (commented):
Thanks, looks good!

krlmlr (approved):
Looks good, except our `as.phylo()` generic. Can you please fix and merge when CI passes?

krlmlr:
Why do we need `UseMethod("ape::as.phylo")` ?

What happens if you remove the `as.phylo()` export, without replacement?

NAMESPACE
```diff
@@ -141,7 +140,7 @@ export(articulation.points)
 export(articulation_points)
 export(as.directed)
 export(as.igraph)
-export(as.phylo)
+export(as.phylo.communities)
```
krlmlr, line 143: This needs to go to.

krlmlr:
R CMD check errors:

```
  as_phylo.igraphHRG: no visible global function definition for
    ‘as.phylo’
  dendPlotPhylo: no visible global function definition for ‘as.phylo’
  hrgPlotPhylo: no visible global function definition for ‘as.phylo’
  Undefined global functions or variables:
    as.phylo
```

NAMESPACE
krlmlr: - Remove the appropriate `#' @export` tag
- Ensure that `as.phylo.communities()` is registered on load (if ape is loaded) or as soon as ape is loaded, by calling `s3_register()` in `.onLoad()` or via `rlang::run_on_load()`

NAMESPACE
```diff
@@ -30,6 +30,7 @@ S3method(as.matrix,igraph)
 S3method(as_ids,igraph.es)
 S3method(as_ids,igraph.vs)
 S3method(as_phylo,communities)
+S3method(as_phylo,igraphHRG)
```
krlmlr, line 33: I wonder why these entries still exist. Should they also be methods for the `ape::as.phylo()` generic and registered via `s3_register()` .

krlmlr (commented):
For the CRAN failures, you need to consistently call `ape::as.phylo()` instead of `as.phylo()` .

krlmlr:
Do you need more input here?

maelle:
I'm sorry, I'm confused once again. I thought we needed to remove `as.phylo.communities` but it's only defined in this package so let me add it back...

maelle:
@krlmlr I'm still not entirely sure about the changes, even if R CMD check seems happy, so would appreciate a review.

NAMESPACE
krlmlr: I still think this needs to go. All the `as.phylo()` methods should be registered dynamically when/if ape is loaded.

R/as_phylo.R
```diff
@@ -0,0 +1,42 @@
...
+#' @family community
+#' @export
+#' @param x object to be coerced
+#' @inheritParams ape::as.phylo
+as_phylo <- function(x, ...) {
+  lifecycle::deprecate_soft(
+    "1.5.0",
+    "ape::as.phylo()"
+  )
+  UseMethod("ape::as.phylo")
```
krlmlr, line 17: I'm not sure that this is going to work. This should check that ape is installed and forward to `ape::as.phylo()`, with a deprecation message. This is no longer a generic, we don't need `UseMethod()` here.

Or, alternatively, with or without deprecation message:

R/as_phylo.R
```diff
@@ -0,0 +1,42 @@
...
+#' @rdname as_phylo
+#' @inheritParams ape::as.phylo
+#' @export
+as_phylo.igraphHRG <- function(x, ...) {
+  lifecycle::deprecate_soft(
+    "1.5.0",
+    "as.phylo()"
+  )
+  ape::as.phylo(x, ...)
+}
```
krlmlr, line 41: We don't need those if the generic is gone.

R/as_phylo.R
```diff
@@ -0,0 +1,42 @@
...
+rlang::on_load(s3_register("ape::as.phylo", "igraphHRG"))
```
krlmlr, line 42: Should this live closer to the declaration of the method?

R/community.R
```diff
@@ -96,7 +96,7 @@
 #' `as.hclust` is similar to [as.dendrogram()], but converts a
 #' hierarchical community structure to a `hclust` object.
 #'
-#' `as_phylo()` converts a hierarchical community structure to a `phylo`
+#' `as.phylo()` converts a hierarchical community structure to a `phylo`
```
krlmlr, line 99:
```suggestion
#' [ape::as.phylo()] converts a hierarchical community structure to a `phylo`
```

R/community.R
```diff
@@ -96,7 +96,7 @@
 #' `as.hclust` is similar to [as.dendrogram()], but converts a
```
krlmlr, line 96:
```suggestion
#' [stats::as.hclust()] is similar to [as.dendrogram()], but converts a
```

R/as_phylo.R
krlmlr, line 17:
```suggestion
  check_installed("ape", "for using `as.phylo()`")
  UseMethod("as.phylo")
```

NAMESPACE
maelle: But if they don't exist at all how would that not break old  code?

R/as_phylo.R
maelle: You mean `as.phylo.igraphHRG()` ?

krlmlr (changes requested):
I can take a look here.

krlmlr (approved):

PR #768, opened 2023-04-05
chore: igraph 0.10

### LINKED ISSUE #753: Use igraph/C 0.10
In 32-bit mode, without changes to the internal data structure.

- [x] Build system adapted, https://github.com/search?q=org%3Acran+cmake+path%3A**%2FDESCRIPTION&type=code
- [x] Compiles without error on your machine
- [x] `pkgload::load_all()` works on your machine
- [x] Compiles without warnings on your machine
- [x] Compiles without error on my machine
- [x] `pkgload::load_all()` works on my machine
- [x] Compiles without warnings on my machine
- [x] Compiles without error on CI/CD
- [x] `make_ring(3)` works
- [ ] All tests pass

Needs #752.

Antonov548:
- [x] Remove `scg` related functions
- [x] FIx stimulus to add reference (`&`) to input arguments (related to missed types in `types-RC.yaml`)
- [x] Use `int` for `igraph_bool_t` - changed in vendor/cigraph
- [x] Fix stimulus to check return type of function
- [ ] Remove `multi` in `get.edge.ids` file `interface.R`
- [ ] Change R interface for `R_igraph_weighted_adjacency` - weights is not string anymore
- [ ] Configure `config.h.in`

Closes #753.


krlmlr:
Thanks. How do I build or install the package now?

Antonov548:
> Thanks. How do I build or install the package now?

I was writing you email. 

I have removed 0.9 files. There is now `build_install.sh` script. And after this you can try `R CMD INSTALL`

CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Minimum CMake that we require is 3.16 because we use --ignore-eol when
+# comparing unit test results with expected outcomes (added in 3.14) and we
+# also use SKIP_REGULAR_EXPRESSION to handle skipped tests properly
+cmake_minimum_required(VERSION 3.16)
```
szhorvat, line 4: 0.10 requires 3.18

Antonov548: Yes, I will update it.

krlmlr:
Thanks. We still need all sources under `src/` or perhaps elsewhere so that we can use `devtools::install_github()` . I placed them in `src/vendor/cigraph` for now, that would be the idiomatic place. But we could also have it in `./cigraph` .

I'm now getting errors from CMake when running `sh build_install.sh` .

Can you work with that? Please revert if needed.

Eventually, `./configure` should contain what `build_install.sh` contains now, and only that.

szhorvat:
> There is now `build_install.sh` script. And after this you can try `R CMD INSTALL`

Just so I know what to expect, is this supposed to work at the moment? (It doesn't here.)

Some things that appear to be missing:

 - `USING_R` must be defined when compiling the C core
 - The boolean type must be adjusted to be an `int`.

Antonov548:
> * `USING_R` must be defined when compiling the C core

I have added `USING_R` definition. It is also require R headers. I have used cmake script which we used before to find R headers.

Antonov548:
@szhorvat Seems there is now some missed types for stimulus in `types-RC.yaml`. Should we add all new types as I did in last commit?

szhorvat:
A note on any decision to rely on CMake:

R is special in that no binary packages are distributed for Linux. Thus it is common for inexperienced users to compile packages themselves. Many older, but still commonly used Linux distros/versions include a CMake that is older than 3.18 (e.g. Ubuntu 20.04). There are not too difficult ways to install a recent CMake on these distros (e.g. with snap), but don't expect inexperienced users to be able to do this, and be prepared for a flood of complaints ...

krlmlr:
Thanks for the note. Things have changed, now there's Posit (RStudio) Package Manager that serves binaries for all but the most exotic Linux variants. We'll see.

krlmlr:
Example: https://packagemanager.posit.co/client/#/repos/2/packages/igraph

Antonov548:
@krlmlr all compile errors in `rinterface.c` are fixed. Now it's time to fix `_extra.c`. There is some error which I don't know how to fix and require additional discussion. 

I will try to fix at the beginning a trivial errors such as types incompatibility and etc.

krlmlr:
Thanks. Happy to review an intermediate state if you push.

Antonov548:
@krlmlr I have pushed some fixes in `rinterface_extra.c` there is still some of them.

The problem is that in some function `vector_t` replaces to `vector_int_t`, but some only some arguments and in some not replaces at all, so it's about going one by one function and fix.

szhorvat:
@Antonov548 Do you have instructions on how to compile this branch (what to do after `sh build_install.sh`)?

szhorvat:
> @szhorvat Seems there is now some missed types for stimulus in `types-RC.yaml`. Should we add all new types as I did in last commit?

@ntamas would be the right person to ask.

configure.win
```diff
@@ -0,0 +1,19 @@
...
+# Set version of igraph
+# FIXME: Ensure this is written when vendoring
+if ! [ -f ./vendor/cigraph/IGRAPH_VERSION ]; then
+    echo "0.10.0" >> ./vendor/cigraph/IGRAPH_VERSION
+fi
+
+# Build igraph with cmake
+cmake -H. -Bbuild -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install -DCMAKE_SHARED_LIBRARY_PREFIX_CXX=""
+
+make -C build install
```
szhorvat, line 16: Same comment as before: do not call `make` directly. Use `cmake.`

ntamas:
Re the `igraph_vector_int_copy()` deprecation warning: the replacement is `igraph_vector_int_init_copy()` and this applies to all `*_vector_*` functions. The reason is that we wanted to make it explicit in the name of the function that this variant _initializes_ the first argument _and_ then copies another vector to it. (Contrast it with `igraph_vector_int_update()`, which takes an _already initialized_ vector and then updates it from another one.

During the transition from igraph 0.9 to igraph 0.10, we also introduced a few additional types in `types.yaml` of the C core. The corresponding generated code snippets should be added to `types-RC.yaml` and `types-RR.yaml` on the R side.

Antonov548:
> `types-RR.yaml` on the R side

Thanks for advice. I have first compliable C interface, but with few comments. 

How to deal with `igraph_get_adjacency` ? It also require `weights` now, should it be also passed from R?

Now I will try to add all missed types in RR.

image

szhorvat:
> How to deal with `igraph_get_adjacency` ? It also require `weights` now, should it be also passed from R?

Let's spell out the question here.

Currently, `as_adjacency_matrix()` takes an `attr` parameter. This must be the name of an edge attribute, whose values will be used in the matrix.

Other functions that work with weights take a `weights` parameter, which is a vector of weights. This achieves the same purpose in a different manner.

This was regularized in the C core, and now `igraph_get_adjacency()` does not access attributes directly, instead it takes a weight vector, just like other functions. In general, the principle for the C core is not to hard-code attribute names, but pass weight-like data as vectors. You will see a change like this in a few other functions as well.

The motivation for changing this in the C core is entirely unrelated to R, and same change does not need to be made in R.  The decision comes down to when you want to make breaking changes, and how many breaking changes you're willing to make. Changing this would obviously break a lot of code out there, IMO it would make more sense for igraph 2.0.  And in igraph 2.0, instead of choosing between the `weights` and `attr` approach, we could support _both_ in _all_ functions that work with weights. python-igraph allows passing either a list of weights, or an attribute name, for any weight argument. This is more convenient and more flexible, but I don't know how to make it work in R, as R does not have scalar values, so it's impossible to distinguish between an attribute name and a length-1 character vector.

szhorvat:
Another related point here is that sparse adjacency matrices are implemented in pure R at the moment, as C/igraph 0.9 did not fully support them. C/igraph 0.10 does, so it's an option to use that functionality. If you choose to do that, you will need to implement a conversion to/from `igraph_sparsemat_t` and R's sparse matrices. Look at `igraph_sparsemat_getelements()` for that, as this function returns the raw CSC format, which R might also be using.  Sorry about the bad docs, we'll need to improve those. Feel free to complain whenever the C docs are not clear enough.

krlmlr:
I still can't compile the package from this branch.

Can you please commit all files that are necessary for compilation, even if they are autogenerated? We will sort it out later.

Antonov548:
> I still can't compile the package from this branch.

Okay, I have pushed changes with `config.h`. But `config.h` in theory is individual for enviorement. I have created `config.h` for gitpod - https://github.com/cynkra/gitpod . 

On gitpod you should be able to see compilation errors which is left.

Also I have created pull request to `stimulus` with related changes to this pull request https://github.com/igraph/stimulus/pull/3

configure
```diff
@@ -1,6439 +1,19 @@
...
-echo "*** Compiler settings used:"
-echo "    CC=${CC}"
-echo "    LD=${LD}"
-echo "    CFLAGS=${CFLAGS}"
-echo "    CPPFLAGS=${CPPFLAGS}"
-echo "    CXX=${CXX}"
-echo "    CXXFLAGS=${CXXFLAGS}"
-echo "    LDFLAGS=${LDFLAGS}"
-echo "    LIBS=${LIBS}"
+make -C build install
```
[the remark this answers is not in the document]
krlmlr: The new code doesn't work for me. Can we also use `cmake -G "Unix Makefiles"` so that the `make` call keeps working in all cases?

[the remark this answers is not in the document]
szhorvat: I don't think that's a good idea. "Unix Makefiles" is not suitable for all situations. For example, MSYS2, which RTools is based on, requires "MSYS Makefiles".

The CMake call should be fixed instead.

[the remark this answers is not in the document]
szhorvat: @krlmlr Can you share the output / error you get?

configure
```diff
@@ -1,6439 +1,20 @@
...
-  { printf "%s\n" "$as_me:${as_lineno-$LINENO}: WARNING: unrecognized options: $ac_unrecognized_opts" >&5
-printf "%s\n" "$as_me: WARNING: unrecognized options: $ac_unrecognized_opts" >&2;}
+# Set version of igraph
+# FIXME: Ensure this is written when vendoring
+if ! [ -f ./vendor/cigraph/IGRAPH_VERSION ]; then
+    echo "0.10.0" >> ./vendor/cigraph/IGRAPH_VERSION
 fi
 
+# Build igraph with cmake
+cmake -H. -Bbuild -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install -DCMAKE_SHARED_LIBRARY_PREFIX_CXX=""
```
szhorvat, line 14:
```suggestion
cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install -DCMAKE_SHARED_LIBRARY_PREFIX_CXX=""
```

The `-H` flag prints a help message. Why is it used here? Did you mean to use `-S` to indicate the source directory?

Why is `CMAKE_SHARED_LIBRARY_PREFIX` set here? I see a warning:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_SHARED_LIBRARY_PREFIX_CXX
```

configure
```diff
@@ -1,6439 +1,20 @@
...
-echo "    CC=${CC}"
-echo "    LD=${LD}"
-echo "    CFLAGS=${CFLAGS}"
-echo "    CPPFLAGS=${CPPFLAGS}"
-echo "    CXX=${CXX}"
-echo "    CXXFLAGS=${CXXFLAGS}"
-echo "    LDFLAGS=${LDFLAGS}"
-echo "    LIBS=${LIBS}"
+cmake --build build
+cmake --install build
```
szhorvat, line 17:
```suggestion
cmake --build build --config Release
cmake --install build --config Release
```

I believe this is necessary to make it work with multi-configuration generators. After adding this, the build also works with the Xcode generator on my system, but otherwise it doesn't.

krlmlr:
I tweaked the code in the https://github.com/igraph/rigraph/tree/igraph-0.10-krlmlr branch, by removing code that doesn't compile or gives errors otherwise. Now the C code compiles, and can also be linked on macOS (with a few extra flags I need on that platform, 06e4e7c42eada2df17e75eb667a41a24d8c1baf2, not part of that branch yet).

On Linux (GitPod), I'm seeing:

```
$ R CMD INSTALL .
...
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for ‘igraph’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/workspace/gitpod/R/x86_64-pc-linux-gnu-library/4.2/00LOCK-igraph/00new/igraph/libs/igraph.so':
  /workspace/gitpod/R/x86_64-pc-linux-gnu-library/4.2/00LOCK-igraph/00new/igraph/libs/igraph.so: undefined symbol: igraph_lazy_adjlist_destroy
```

I tried to change `set(CMAKE_C_VISIBILITY_PRESET hidden)` to `default` or `protected` to no avail. Can you reproduce? How to fix?

I'd really like to have, as soon as possible, a version where we can use `R CMD INSTALL` . This will help assess the work that still needs to be done. Can you please integrate my changes into this branch and fix the remaining linking errors on Ubuntu?

A next step would be to review the tests, one by one. For complicated failures, we can use `skip()` in the test code for now, but we should get the basics right first:

```
✗ TESTTHAT_PARALLEL=FALSE R -q -e 'testthat::test_local(reporter = "location")'
> testthat::test_local(reporter = "location")
Start test: we can create vertex/edge seqs
  'test-vs-es.R:3:3' [success]
  'test-vs-es.R:4:3' [success]
  'test-vs-es.R:7:3' [success]
  'test-vs-es.R:8:3' [error]
End test: we can create vertex/edge seqs

Start test: vs/es keeps names
  'test-vs-es.R:20:3' [success]
  'test-vs-es.R:23:3' [success]
  'test-vs-es.R:25:3' [error]
End test: vs/es keeps names

Start test: vs/es refers to the graph
  'test-vs-es.R:39:3' [success]
  'test-vs-es.R:40:3' [success]
End test: vs/es refers to the graph

Start test: vs/es refers to the original graph
R(95492,0x203d3c140) malloc: *** error for object 0x104e21fd0: pointer being freed was not allocated
R(95492,0x203d3c140) malloc: *** set a breakpoint in malloc_error_break to debug
[1]    95492 abort      TESTTHAT_PARALLEL=FALSE R -q -e 'testthat::test_local(reporter = "location")'
```

Antonov548:
> Can you please integrate my changes into this branch and fix the remaining linking errors on Ubuntu?

I have integrated changes:
- fixed R bad generated code
- updated names of `deprecated` functions
- etc.

For now I have error related to `-fPIC`
@szhorvat Does it mean that we can use only shared library of core `igraph`?

image

szhorvat:
> CC: or should we use `CMAKE_POSITION_INDEPENDENT_CODE` in rigraph?

Yes, exactly.

Generally, PIC is necessary for shared libraries, which the final product will be here for R/igraph. However, by default, CMake does not use PIC when compiling into a static library, which is what we do for C/igraph here. If that static library will be linked into a shared library in the end (and not just an executable) then use `-DCMAKE_POSITION_INDEPENDENT_CODE=ON`.

szhorvat:
There are some RNG-related fixes needed that are present partially (but not completely) on Kirill's branch. Do you want me to push these?

After that I get:

```
Error: package or namespace load failed for ‘igraph’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Library/Frameworks/R.framework/Versions/4.2/Resources/library/00LOCK-rigraph/00new/igraph/libs/igraph.so':
  dlopen(/Library/Frameworks/R.framework/Versions/4.2/Resources/library/00LOCK-rigraph/00new/igraph/libs/igraph.so, 6): Symbol not found: _R_igraph_incidence
  Referenced from: /Library/Frameworks/R.framework/Versions/4.2/Resources/library/00LOCK-rigraph/00new/igraph/libs/igraph.so
  Expected in: flat namespace
 in /Library/Frameworks/R.framework/Versions/4.2/Resources/library/00LOCK-rigraph/00new/igraph/libs/igraph.so
Error: loading failed
```

krlmlr:
Please push anything that improves the state of the code, even partial fixes.

Can you search for `R_igraph_incidence` in the R code, comment it (with a FIXME) and rerun `R -q -e 'cpp11::cpp_register()'` ?

Antonov548:
> Can you search for `R_igraph_incidence` in the R code, comment it (with a FIXME) and rerun `R -q -e 'cpp11::cpp_register()'` ?

I have added `R_igraph_incidence` to `_extra` interface for now.
igraph is compiled (on my PC, I will try it on gitpod now), but it failed to load: 
image

szhorvat:
> Please push anything that improves the state of the code

Sure, just didn't want to step on your toes.

Antonov548:
I'm only didn't push this changes: https://github.com/igraph/rigraph/commit/34f380d6fafbfa48c626177caddcd0e8633083ea#r109166731 

Not sure for what it is.

szhorvat:
I have to stop looking at this today. At this point I see `Symbol not found: _dnaupd_`. I just wanted to let you know that this is from ARPACK, in case this helps.

At this point I highly recommend that you explicitly set all `IGRAPH_USE_INTERNAL...` settings when compiling C/igraph, and not leave these up for CMake to guess. Otherwise we'll all see different behaviours on our different systems. 

A quickfix for this is to use the vendored f2c-translated ARPACK. CRAN may not be happy with that, but it can be fixed later.

szhorvat:
> I'm only didn't push this changes: [34f380d#r109166731](https://github.com/igraph/rigraph/commit/34f380d6fafbfa48c626177caddcd0e8633083ea#r109166731)
> 
> Not sure for what it is.

That's done now, see my previous commit.

Antonov548:
> I have to stop looking at this today

Thanks for help. 
I added some of them already - https://github.com/igraph/rigraph/pull/768#issuecomment-1510377970

Maybe on my PC it found local arpack, seems it's really required to add them all. On my machine btw igraph compiled after your changes:
image

szhorvat:
Currently I quickly set `cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install -DIGRAPH_USE_INTERNAL_ARPACK=ON -DIGRAPH_USE_INTERNAL_GMP=ON -DIGRAPH_OPENMP_SUPPORT=OFF` in `configure`, and on my macOS 10.14 machine it now compiles and links.

It can even be loaded.

Of course it doesn't actually work, but some progress :-)

```
> library(igraph)

Attaching package: ‘igraph’

The following objects are masked from ‘package:stats’:

    decompose, spectrum

The following object is masked from ‘package:base’:

    union

> g<-make_graph(c(1,2,2,3))
Error in (function (edges, n = max(edges), directed = TRUE)  : 
  INTEGER() can only be applied to a 'integer', not a 'double'
> 

```

szhorvat:
> Maybe on my PC it found local arpack

Much more likely, it did not find any ARPACK at all, so it automatically chose the vendored ARPACK. My point is precisely to not let it make these decisions automatically, at least for the moment, as the linking step is not yet adapted to CMake's automatic dependency detection.

Antonov548:
> Of course it doesn't actually work, but some progress :-)

Nice, already something :)
RR-interface wasn't touched, I think that's why.

krlmlr:
Confirming that the current branch compiles and loads on my machine too. Superb, thanks for pushing through!

I use:

```patch
diff --git a/configure b/configure
index 9364713f7..e5e2dcd5c 100755
--- a/configure
+++ b/configure
@@ -11,9 +11,8 @@ if ! [ -f ./vendor/cigraph/IGRAPH_VERSION ]; then
 fi

 # Build igraph with cmake
-cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install
+cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install  -DIGRAPH_USE_INTERNAL_BLAS=ON -DIGRAPH_USE_INTERNAL_LAPACK=ON -DIGRAPH_USE_INTERNAL_ARPACK=ON -DIGRAPH_USE_INTERNAL_GLPK=ON -DIGRAPH_USE_INTERNAL_GMP=ON -DIGRAPH_OPENMP_SUPPORT=OFF

-cmake --build build --config Release
-cmake --install build --config Release
+make -C build install

 cp -r ./install/include/igraph ./include
```

because `cmake --build` still doesn't work for me.

I'm on macOS aarch FWIW.

I will work on splitting this PR into more manageable chunks. I'll start with including igraph/C 0.10 side by side with 0.9, this will dramatically minimize the size of this PR. Please proceed with fixing tests in this branch.

szhorvat:
`-DIGRAPH_USE_INTERNAL_BLAS=ON -DIGRAPH_USE_INTERNAL_LAPACK=ON `

I believe these should NOT be used, as we need to link to R's own BLAS/LAPACK. All the rest of the settings you show look good for me, at least for an initial version.

krlmlr:
You can do:

```sh
TESTTHAT_PARALLEL=FALSE R -q -e 'testthat::test_local(reporter = "location")'
```

or, e.g.,

```r
TESTTHAT_PARALLEL=FALSE R -q -e 'testthat::test_local(filter = "^edgenames$", reporter = "location")'
```

to only run the "edgenames" test file.

szhorvat:
> because `cmake --build` still doesn't work for me.

Can you share the output in a pastebin?

szhorvat:
> I believe these should NOT be used (or rather should be set to OFF), as we need to link to R's own BLAS/LAPACK.

I'm sorry, I really have to stop commenting.

I was wrong here. Setting this to `OFF` won't achieve linking to R's BLAS/LAPACK. Instead, CMake will look for other BLAS implementations (which are shipped by default on macOS) and if it doesn't find any, it will error.

You're right, just set it to `ON` _for now_, and we'll need to figure out how to link to R's version at a later point.

krlmlr:
Output of `R CMD INSTALL .` in a clean working copy:

[cc.log](https://github.com/igraph/rigraph/files/11242376/cc.log)

Antonov548:
I can't get why build is failing on CI. There is no includes which should be provided from Makevars? Does CI use some another Makevars?

szhorvat:
@krlmlr Can you please run `cmake --help` and paste here the last section of the output, which shows the generators available for you, as well as the default, marked with a `*`? On my system I see the following, assuming I unset the `CMAKE_GENERATOR` envvar:

```
The following generators are available on this platform (* marks default):
* Unix Makefiles               = Generates standard UNIX makefiles.
  Ninja                        = Generates build.ninja files.
  Ninja Multi-Config           = Generates build-<Config>.ninja files.
  Watcom WMake                 = Generates Watcom WMake makefiles.
  Xcode                        = Generate Xcode project files.
  CodeBlocks - Ninja           = Generates CodeBlocks project files.
  CodeBlocks - Unix Makefiles  = Generates CodeBlocks project files.
  CodeLite - Ninja             = Generates CodeLite project files.
  CodeLite - Unix Makefiles    = Generates CodeLite project files.
  Eclipse CDT4 - Ninja         = Generates Eclipse CDT 4.0 project files.
  Eclipse CDT4 - Unix Makefiles= Generates Eclipse CDT 4.0 project files.
  Kate - Ninja                 = Generates Kate project files.
  Kate - Unix Makefiles        = Generates Kate project files.
  Sublime Text 2 - Ninja       = Generates Sublime Text 2 project files.
  Sublime Text 2 - Unix Makefiles
                               = Generates Sublime Text 2 project files.
```

If you are in fact using the "Unix Makefiles" generator, did you set any special flags to increase verbosity?  The output I usually see with "Unix Makefiles" includes a progress indicator, with percentage, and in basic verbose mode, the compilation command.

I've never seem raw `make` and `touch` calls being echoes, without compilation commands being shown.

image

krlmlr:
Good catch, the following works on my machine too:

```patch
diff --git a/configure b/configure
index 9364713f7..51788b810 100755
--- a/configure
+++ b/configure
@@ -13,7 +13,7 @@ fi
 # Build igraph with cmake
 cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install

-cmake --build build --config Release
-cmake --install build --config Release
+MAKEFLAGS= cmake --build build --config Release
+MAKEFLAGS= cmake --install build --config Release

 cp -r ./install/include/igraph ./include
```

I have `MAKEFLAGS=-j 8 -O` in my R config, this also gets applied for the `configure` script.

krlmlr:
I don't mind postponing fixing CI/CD until later.

krlmlr:
The `-O` harms `cmake --build` . The following works:

```patch
diff --git a/configure b/configure
index 9364713f7..9c75ea263 100755
--- a/configure
+++ b/configure
@@ -13,7 +13,7 @@ fi
 # Build igraph with cmake
 cmake -S . -B build -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install

-cmake --build build --config Release
-cmake --install build --config Release
+MAKEFLAGS="-j 8" cmake --build build --config Release
+MAKEFLAGS="-j 8 -O" cmake --install build --config Release

 cp -r ./install/include/igraph ./include
```

ntamas:
> You're right, just set it to ON for now, and we'll need to figure out how to link to R's version at a later point.

I think that CMake looks at certain default, system-wide directories when it's looking for libraries, but you can prepend your own directories to this list with _some_ CMake setting (I don't remember off the top of my head and I'm not at my desk now). The easiest would probably be to prepend R's preferred library and include folders to CMake's list and let the auto-detection do its job.

szhorvat:
That's a bit complicated due to the non-standard library names used by R. The following work on my system though, and could be used as a starting point for a proper solution.

```
LDFLAGS=-L/Library/Frameworks/R.framework/Libraries/ cmake .. -DBLAS_LIBRARIES=libRblas.dylib -DLAPACK_LIBRARIES=libRlapack.dylib -DBLA_VENDOR=Generic -DIGRAPH_USE_INTERNAL_BLAS=OFF -DIGRAPH_USE_INTERNAL_LAPACK=OFF
```

 - `-DBLAS_LIBRARIES=libRblas.dylib -DLAPACK_LIBRARIES=libRlapack.dylib` These set the library names. I'm not sure if an absolute path is better / more robust than using `LDFLAGS` with simple names.
 - `-DBLA_VENDOR=Generic` May not be needed, attempt to prevent the detection of other BLAS implementations that may be present on the system
 - `-DIGRAPH_USE_INTERNAL_BLAS=OFF -DIGRAPH_USE_INTERNAL_LAPACK=OFF` These are to ensure that we get an error if something goes wrong, and CMake won't just pick up an alternative BLAS.

szhorvat:
> I can't get why build is failing on CI. There is no includes which should be provided from Makevars? Does CI use some another Makevars?

I can reproduce that failure if I remove the system-wide install of igraph I had.

The code that caused the failure seems completely unnecessary. I'm removing it in the below commit.

ntamas:
Maybe we should consider using https://github.com/xtensor-stack/xtensor-r/blob/master/cmake/FindR.cmake to find R, and then we would get the full path to R's BLAS and LAPACK libraries for free?

krlmlr:
I minimized the diff with the main branch so that there is a tiny chance that we can actually review what's happening here.

Is this PR doing too much? Should we keep the data format unchanged for now, continue operating with double vectors in the R interface, and use 64-bit integers in the C core? It seems that we're copying data anyway -- we could copy and raise an error in case of out-of-bounds conditions.

szhorvat:
Why give up now? Good progress has been made. The CI runs fail because they do `make igraph`. That fails locally too, so it should be debuggable. Somehow it doesn't add some of the `-I` compiler options that are present when running simple `R CMD INSTALL .`.

Copying is only necessary in one direction: from igraph to R. In the other direction it's possible to create views into existing vectors, and work with those. Copying in both directions would likely bring a noticeable performance impact.

Antonov548:
@szhorvat I found a reason of some failing tests. I believe a lot of tests are failing to this reason. How we should deal with `cache` property in `igraph_t`? Should be also copied to R structure? 

The problem is that we don't copy cache -> then read graph -> don't read cache -> try delete cache, but it's empty pointer.

ntamas:
Hmmm yes, the cache is a new thing in igraph 0.10; its purpose is to store certain boolean flags about the graph if we already know what their value is. For instance, if we know that the graph has no loop edges (because we have determined it earlier), we can store it as a flag and subsequent calls to the same graph can re-use the cached value without having to check for loop edges again. Essentially the cache is a layer of two boolean vectors: one that tells us for each potentially cacheable boolean feature whether it is currently cached or not, and the other vector tells us what the cached value is (undefined if the property is not cached).

Now, we can do it in C because we know that all manipulations to the graph structure go through our own functions where we manage and maintain the cache. In R, it's not that clear as one can easily see the internals of the graph with a call to `unclass()` and then mess around with the internal vectors without updating the cache. On the other hand, we can argue that the internals of the graph are private, and anyone who calls `unclass()` on a graph has committed a primordial sin. _If_ we can say that, then we can safely copy the cache to the R side and back.

Antonov548:
> primordial sin

Thanks for the explanation. I will try to copy them.
Should I also prepare pull request with `SINS.md`? :)

krlmlr:
The cache could also be an external pointer. It will be lost during `serialize()`, but then it's a cache.

Before we spend too much time here, I'd like to investigate storing the graph object as a raw vector. This can be done in the main branch, independent of this branch. I'll open a new issue.

ntamas:
Circling back to the detection of BLAS: does the standardized R compilation environment (the one used on CRAN)  contain `pkg-config` and is it configured to find R's BLAS / LAPACK first? If so, we could also use CMake's `BLA_PREFER_PKGCONFIG` input variable of the `FindBLAS.cmake` to ask CMake to prefer the BLAS instance returned by `pkg-config`.

PR #778, opened 2023-04-12
Fix linking on Windows: gfortran needs quadmath.

kalibera:
-lquadmath should be specified directly, because it is needed by -lgfortran. So far the linking worked, because Rlapack.dll accidentally re-exported quadmath symbols, but this will be fixed in R-devel shortly and building of igraph would then break without this fix.


ntamas:
I remember certain issues from the past, reported by people who used Microsoft R Open on WIndows and stumbled upon this issue. Thanks for the fix!

PR #779, opened 2023-04-13
fix: Fix tests for dev waldo

hadley:
Which now correctly distinguishes between NA and NaN. I couldn't compile locally so this is a bit of guess, and you might want to verify with dev waldo yourself.


krlmlr:
Thanks! Would you like to share the compilation failures you're seeing?

hadley:
I don't think they're that interesting; I just didn't have the necessary headers installed.

PR #791, opened 2023-04-23
Update CONTRIBUTING.md

ngmaclaren:
Draft updates to CONTRIBUTING.md. It looks like GitHub is flagging changes in most lines: this is because I set an automatic line width that seems to be just a little different from what was there before. Primary substantive changes are to `## Development and Compilation` and the two list sections. I also added some links.


krlmlr:
Thanks!

PR #792, opened 2023-04-25
chore: get/set for igraph data members

Antonov548:


src/rinterface_extra.c
```diff
@@ -3511,12 +3553,15 @@ int R_SEXP_to_igraph(SEXP graph, igraph_t *res) {
...
-  igraph_vector_init_copy(&res->from, REAL(VECTOR_ELT(graph, igraph_t_idx_from)),
-                   GET_LENGTH(VECTOR_ELT(graph, igraph_t_idx_from)));
-  igraph_vector_init_copy(&res->to, REAL(VECTOR_ELT(graph, igraph_t_idx_to)),
-                   GET_LENGTH(VECTOR_ELT(graph, igraph_t_idx_to)));
+  res->n=R_igraph_get_n(graph);
+  res->directed=R_igraph_get_directed(graph);
+
+  igraph_vector_t from;
+  R_igraph_get_from(graph, &from);
+  igraph_vector_copy(&res->from, &from);
```
ntamas, line 3561: `igraph_vector_copy()` is deprecated in 0.10; use [igraph_vector_update()](https://igraph.org/c/doc/igraph-Data-structures.html#igraph_vector_update) instead.

ntamas (changes requested):

src/rinterface_extra.c
ntamas: I think there's another instance of this function call in the PR, which should also be changed.

szhorvat: Most of the code in this PR will need non-trivial changes for 0.10 anyway, e.g.: 1. all the memcpy() won't be feasible anymore 2. instead of creating views (which should be marked with a comment here), we'll need to copy.

Antonov548: @ntamas The idea as I understood was refactoring this for 0.9 and then as soon 0.9 will have independent from integer size structure merge it to 0.10 pull request? Or I miss something?

Antonov548:
Seems url for R package on Windows is not working anymore.

ntamas (approved):

Issue #787, opened 2023-04-21
Use native igraph object as ground truth


krlmlr:
## Converting from R to C

- No work needs to be done, the native igraph object is stored as an external pointer in the `igraph_t_idx_env` part of the data structure.
- When the R graph object is destroyed, it will be garbage-collected eventually. This triggers the finalizer of the external pointer object, which can be made to call destructors in igraph/C.

## Converting from C to R

- We keep the native `igraph_t` object and store it as an external pointer in the `igraph_t_idx_env` part of the data structure.
- We also keep all the information needed to recreate the native `igraph_t` object in the `igraph_t_idx_n`, `igraph_t_idx_directed`, `igraph_t_idx_from` and `igraph_t_idx_to` components
- The native `igraph_t` object may be lost, e.g., during serialization. In this case, we recreate the native `igraph_t` object from the `igraph_t_idx_n`, `igraph_t_idx_directed`, `igraph_t_idx_from` and `igraph_t_idx_to` components

## Attributes

All attributes are still handled only from R for now. ~~But because the native `igraph_t` object may cache R-level attributes, we need to be especially careful about this.~~

## Implementation steps

- [x] Ensure that all indices into `unclass(g)` use `igraph_t_idx_*`, no magic numbers (#784)
- [ ] Fix problems in #789
    - This PR contains the results from checking all reverse dependencies against 6222e73a6. This is the commit where all indexes in the internal list have been shifted. It seems like our code (or perhaps even other packages) are relying on those indexes. We need to understand what causes the failures, and fix them.
- [x] Implement accessors (getters + setters) for `igraph_t_idx_n`, `igraph_t_idx_directed`, `igraph_t_idx_from` and `igraph_t_idx_to` in C
- [x] Don't use `igraph_t_idx_*` from R, at all (we can keep using `igraph_attr_idx_*`)
    - create R wrappers for accessors from previous step if needed
- [x] Store native `igraph_t` as external pointer in `igraph_t_idx_env`
    - the native `igraph_t` is an external pointer stored in `unclass(graph)[[igraph_t_idx_env]]$igraph`
- [x] Rewrite C-level accessors to refer to native `igraph_t`
    - fallback to recreating native `igraph_t` from `igraph_t_idx_n`, `igraph_t_idx_directed`, `igraph_t_idx_from` and `igraph_t_idx_to` if the external pointer is gone
    - use `unserialize(serialize(graph, NULL))` to test
[...]

PR #794, opened 2023-05-01
chore: Store graph pointer in environment

Antonov548:

Store native igraph_t as external pointer in igraph_t_idx_env
the native igraph_t is an external pointer stored in unclass(graph)[[igraph_t_idx_env]]$igraph

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -113,14 +113,14 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - {os: windows-latest, r: 'release'}
+          #- {os: windows-latest, r: 'release'}
           - {os: macOS-latest,   r: 'release'}
-          - {os: ubuntu-20.04,   r: 'release',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
-          - {os: ubuntu-20.04,   r: 'devel',    rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
-          - {os: ubuntu-20.04,   r: 'oldrel-1', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
-          - {os: ubuntu-20.04,   r: 'oldrel-2', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
-          - {os: ubuntu-20.04,   r: 'oldrel-3', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
-          - {os: ubuntu-20.04,   r: 'oldrel-4', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'release',  rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'devel',    rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'oldrel-1', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'oldrel-2', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'oldrel-3', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          #- {os: ubuntu-20.04,   r: 'oldrel-4', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
     env:
       RSPM: ${{ matrix.config.rspm }}
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
@@ -148,6 +148,9 @@ jobs:
         with:
           extra-packages: any::rcmdcheck
           needs: check
+
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
 
       - name: Check R package
         uses: r-lib/actions/check-r-package@v2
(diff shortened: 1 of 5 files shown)


Antonov548:
@ntamas I have added copy of graph as external pointer to env. But after coping graph attributes become different from original graph. There is failing tests now. What could be a root cause?

ntamas:
Vague idea: `igraph_copy()` actually calls the attribute handler functions registered by `igraph_set_attribute_table()` in `R_igraph_init_handlers()`. Could it be the case that somehow it's essentially working on an "incomplete" graph data structure (since the R object representing the C graph is just being constructed) and that's why the tests fail? Yet another reason why we should try avoiding `igraph_copy()` here.

Antonov548:
> So, `R_igraph_to_SEXP()` now always makes a _copy_ of `g` and then destroys `g` itself, instead of simply holding on to a pointer to `g` and _not_ destroying it. Maybe doing so would solve the problem with this PR?

I thought to have it as next step and just try to copy at first. But now I see, the problem is `igraph_i_attribute_table` which increase ref count in attributes after the copy operation.

Is there any function to have `move` logic from one graph to another?

ntamas:
No there isn't, unfortunately.

szhorvat:
"`move` logic" is really a C++ concept, meant to deal with issues stemming from encapsulation. It doesn't make sense in this context. "Moving" is trivial in C:

If you have `igraph_t g1, g2`, then "moving `g2` to `g1`" means `g1 = g2` and remembering that `g2` must not be touched anymore.

----

I should say that I find it a bit hard to follow this refactoring when it's broken into so many PRs which aren't truly independent.

Antonov548:
> "`move` logic" is really a C++ concept, meant to deal with issues stemming from encapsulation. It doesn't make sense in this context. "Moving" is usually trivial in C:

Thanks. Yes, I agree. I just was wondering if there any specific function in igraph core for this.

I will try to make bigger and consistency pull request.

ntamas:
`R_igraph_attribute_table` is a table of callback functions that the C core of igraph calls whenever it wants to do something with attributes. For instance, if the C core deletes a bunch of vertices from the graph, it uses one of these callbacks to inform the higher level interface that some vertices have been removed so the higher level interface can update its own internal data structures related to attribute storage to remain consistent with the graph itself. See [here](https://igraph.org/c/doc/igraph-Attributes.html#igraph_attribute_table_t) for more information.

ntamas:
All in all, I'm not sure whether this is correct, but I can't convince myself that this is incorrect either. `igraph_i_attribute_destroy()` essentially calls the "destroy" callback function from `R_igraph_attribute_table`, which points to `R_igraph_attribute_destroy()`. Its purpose is to undo whatever `R_igraph_attribute_init()` was doing, and something feels odd here. I guess it has to do something with copying the internals of an `igraph_t` (i.e. the "move" operation).

Internally, the `igraph_t` struct contains a `void*` pointer named `attr`. `attr` is there to allow the callback functions in `R_igraph_attribute_table` to access the attribute storage area of the graph object without having a handle to the full `SEXP` object that represents the graph on the R side. `attr` is prepared in `R_igraph_attribute_init()`. By copying the `igraph_t` object the way you do now, you also copy the `attr` pointer. Then you destroy it with `igraph_i_attribute_destroy()` (which calls `R_igraph_attribute_destroy()`), and it seems that it _happens_ to work now but I don't know how fragile this will be.

Is there any chance to avoid even the "move" operation in `R_igraph_to_SEXP` or is that strictly needed?

Antonov548:
> Is there any chance to avoid even the "move" operation in `R_igraph_to_SEXP` or is that strictly needed?

Thanks for this detail explanation. But I didn't understand part with avoid "move". I changed that now it has "move" of graph:
https://github.com/igraph/rigraph/blob/0c17db914927b703dca6c5d3b67c1756f7ac2948/src/rinterface_extra.c#L2918-L2919

If avoid "move" how then environment can save igraph object?

I also think that it's fragile solution and looks like not like a proper way to do this. I think to implement properly I need to check implementation of `R_igraph_attribute_destroy()` and `R_igraph_attribute_init()`.

Antonov548:
One failed test, which is related to old-graph structure. For now I don't understand the root cause.

Now I think deleting `attributes` it's correct way of managing them. 
Also here is function `R_igraph_get_pointer` which returns the native `igraph_t` from enviorement. 

And `R_igraph_get_n` use this function to get `n` for `igraph_t`.

ntamas:
I wanted to test this PR on my end on macOS, but I'm getting errors like this during compilation:

```
core/isomorphism/bliss/graph.hh:187:49: warning: 'function<void (unsigned int, const unsigned int *)>' is deprecated: Using std::function in C++03 is not supported anymore. Please upgrade to C++11 or later, or use a different type [-Wdeprecated-declarations]
```

How did you work around this? Am I missing something?

szhorvat:
Sounds like C++11 is not enabled. Bliss requires C++11 (or later).

ntamas:
Okay, it was my fault. I have an `~/.R/Makevars` file that I used to add `ccache` support. Apparently the flags forcing C++11 were erased by the presence of this file.

krlmlr:
- `R_igraph_set_n()` is fine, it's used only when going from igraph object to R
- See inline comment regarding happy path
- `R_igraph_env_graph_pointer()` should perhaps be `R_igraph_set_pointer()`, and move this function or `R_igraph_get_pointer()` so that they are closer together?
- Next step, as described in #787: `R_igraph_get_n()` and the other getters should check if there is an external pointer, and recreate the igraph object from the R vectors if there's not. Once recreated, it should be stored in the environment, just like after `R_igraph_to_SEXP()`

Since all tests pass, I'm merging. Can you please proceed?

krlmlr:
- `R_igraph_set_n()` is fine, it's used only when going from igraph object to R
- See inline comment regarding happy path
- `R_igraph_env_graph_pointer()` should perhaps be `R_igraph_set_pointer()`, and move this function or `R_igraph_get_pointer()` so that they are closer together?
- Next step, as described in #787: `R_igraph_get_n()` and the other getters should check if there is an external pointer, and recreate the igraph object from the R vectors if there's not. Once recreated, it should be stored in the environment, just like after `R_igraph_to_SEXP()`

Since all tests pass, I'm merging. Can you please proceed?

PR #796, opened 2023-05-10
chore: igraph pointer accessors 

Antonov548:


krlmlr (approved):
Thanks. Good to go, once the checks are green.

Antonov548:
@krlmlr I have question regarding the recreating igraph pointer. 
https://github.com/igraph/rigraph/blob/07310eeb217cf3fc54a843f6cd0cb98631b0b6c8/src/rinterface_extra.c#L2892-L2897

There is now restoring of igraph if it was unbound or null, but if environment is lost then should be recrate environment? 
That's why in getters we need to check if pointer is available, and if not just get value from the list structure, only for case when environment is lost and we can't restore pointer in it.

Antonov548:
Where can I get examples which are failing now?

krlmlr:
We use the R list structure only if the native `igraph_t` object is unavailable. So, each getter should:

- check if the native `igraph_t` is valid (i.e., the environment exists and has a valid external pointer)
    - if not, recreate it from the R list structure
- return the requested value

Does that answer the question?

Antonov548:
> Does that answer the question?

Yes, thanks. Then it what currently there. 

---
I will check what's wrong with examples.

krlmlr:
Thanks!

PR #799, opened 2023-05-13
chore: restore igraph pointer

Antonov548:
Avoid populating igraph_t_idx_oi...igraph_t_idx_is in R


Antonov548:
`R_igraph_get_pointer` restore pointer to igraph from R list structure in case pointer is lost. To restore pointer - `R_igraph_restore_pointer` is called.

The problem now with old graph structure.  Should `R_igraph_get_pointer` create new environment and restore pointer if R structure hasn't environment,?

krlmlr (approved):
Thanks. The code looks good. Why do the checks fail?

Antonov548:
> Thanks. The code looks good. Why do the checks fail?

As I mentioned in comment, the problem with test check of old data type. In old data type there is no environment, so pointer couldn't be restored in environment. Maybe `update_graph` function should be updated?

krlmlr:
Missed that. Perhaps `R_igraph_restore_pointer()` should be split:

- a function that recreates the native igraph object
- a function that adds the native igraph object to the environment

The second part would be called only if the environment exists. The problem is that we aren't really allowed to alter the structure of the object in-place.

I'm not sure what the practical consequences are. We could issue a warning in this case, and offer a way to manually upgrade the structure of the graph R object.

krlmlr (approved):
Looks good now, thanks! Please ping me if the next steps are unclear.

krlmlr:
Please merge when good.

PR #804, opened 2023-05-21
chore: The `from` and `to` components of the R representation of an igraph objects use ALTREP to avoid unnecessary materialization

Antonov548:
Implement ALTREP for igraph_t_idx_from and igraph_t_idx_to (requires R 4.0)


Antonov548:
Seems need to have two different altrep classes for `from` and `to` since it's accessing different members in graph. Or change altrep to store pointer to `from` or `to` member in graph, but not to enviorement.

krlmlr:
Two classes seem fine for me. I'll review when the tests are through.

krlmlr:
I was wondering how this even compiles with oldrel-4... Turns out we only need R >= 3.6, perhaps even R >= 3.5 would be sufficient.

Can you please add verbosity, triggered by an option, so that we can test materialization? Similarly to https://github.com/duckdb/duckdb/blob/85989198104c893f302e2fff6fae6bba65156af8/tools/rpkg/src/reltoaltrep.cpp#L65-L68.

Antonov548:
> I was wondering how this even compiles with oldrel-4... Turns out we only need R >= 3.6, perhaps even R >= 3.5 would be sufficient.

Yes, it interesting. I have added `igraph.verbose` option.

Antonov548:
I think we can add this option ot test which is now failing: https://github.com/igraph/rigraph/pull/806

src/rinterface_extra.c
```diff
@@ -2547,6 +2548,66 @@ int R_igraph_status_handler(const char *message, void *data) {
...
+  SEXP xp=Rf_findVar(Rf_install("igraph"), R_altrep_data1(vec));
+  igraph_t *g=(igraph_t*)(R_ExternalPtrAddr(xp));
+  return igraph_ecount(g);
+}
+
+static void *R_igraph_altrep_from(SEXP vec, Rboolean writeable) {
+  SEXP data=R_altrep_data2(vec);
+  if (data == R_NilValue) {
+    SEXP option=Rf_GetOption(Rf_install("igraph.verbose"), R_BaseEnv);
+    if (option != R_NilValue && !Rf_isNull(option) && LOGICAL_ELT(option, 0) == 1) {
```
krlmlr, line 2561:
```suggestion
    if (option != R_NilValue && TYPEOF(option) == LGLSXP && Rf_xlength(option) == 1 && LOGICAL_ELT(option, 0) == 1) {
```

here and elsewhere.

src/rinterface_extra.c
```diff
@@ -2547,6 +2548,66 @@ int R_igraph_status_handler(const char *message, void *data) {
...
+      Rprintf("Materializing 'from' vector.\n");
```
krlmlr, line 2562: `R_ShowMessage()` is slightly better here.

krlmlr (commented):
Thanks. As discussed, can you please extend the logging to creation and destruction of igraph objects triggered from the R code?

krlmlr:
@Antonov548: can you please extend the logging to creation and destruction of igraph objects triggered from the R code?

Antonov548:
> @Antonov548: can you please extend the logging to creation and destruction of igraph objects triggered from the R code?

Add logs for such events of pointer: make, free, restore.

Also found that rigraph already has function to make logs with option `verbose`.

```R
library(igraph)

igraph_options(verbose = TRUE)

g <- make_ring(3, directed = TRUE)
gs <- unserialize(serialize(g, NULL))

```

krlmlr:
Thanks, great! We can work on enhancing the messages by printing which object is saved/restored, but this is good for now.

``` r
library(igraph)

igraph_options(verbose = TRUE)

g <- make_ring(3, directed = TRUE)
#> Make graph external pointer.
#> Materializing 'from' vector.
#> Materializing 'to' vector.
gs <- unserialize(serialize(g, NULL))
gs
#> Restore graph external pointer.
#> Make graph external pointer.
#> IGRAPH 7eb8913 D--- 3 3 -- Ring graph
#> + attr: name (g/c), mutual (g/l), circular (g/l)
#> + edges from 7eb8913:
#> [1] 1->2 2->3 3->1
```

<sup>Created on 2023-05-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

krlmlr:
Thanks!

PR #805, opened 2023-05-21
fix: Fix direction of edges when restoring

krlmlr:
Should the order be from->to?

PR #806, opened 2023-05-21
Re-add failing test

krlmlr:
This reverts commit 84fb5143c5e7a14af929985a2697b7363ecc31e4.

Follow-up to #805.


krlmlr:
@Antonov548: Very weird. I tested on GitPod in Ubuntu 20.04 and also in a Docker container in 22.04, to no avail.

Can you perhaps put together a Dockerfile that shows the problem?

Some plotting tests now also start to fail in the main branch.

Antonov548:
> ery weird. I tested on GitPod in Ubuntu 20.04 and also in a Docker container in 22.04, to no avail.
> 
> Can you perhaps put together a Dockerfile that shows the problem?
> 
> Some plotting tests now also start to fail in the main branch.

It also depending on the order of running tests or on amount of tests which is running. Because if I run this tests only be one I can't reproduce issue it at all.

Antonov548:
I can confirm that problem is in `igraph_opt`. As I mentioned it's only reproducible when all tests are running which setups this options in different order.

In function `print.igraph` option `print.full` which is sometimes `FALSE` for the test which is failing. As result it doesn't print edges. Maybe we need to expicit specify option `print.full` in this test.

krlmlr:
Interesting. I've opened #809 to observe what happens to that option in the snapshot test. Let's keep this open until we fully understand the problem.

krlmlr:
Now in #809.

Issue #807, opened 2023-05-21
Setting attribute to vertex sequence leads to failed assertion


krlmlr:
**Describe the bug**

A specific usage pattern leads to a failed assertion in the C core.

**To reproduce**

``` r
options(conflicts.policy = list(warn = FALSE))
library(igraph)

gg <- make_ring(3, directed = TRUE)
V(gg)$id <- V(gg)
subgraph(gg, 1)
#> Error in induced_subgraph(graph, vids): At core/core/vector.pmt:483 : Assertion failed: v->stor_begin != NULL. This is an unexpected igraph error; please report this as a bug, along with the steps to reproduce it.
#> Please restart your R session to avoid crashes or other surprising behavior.
```

<sup>Created on 2023-05-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

**Version information**

 - R/igraph version: 342586e4728dd0ef04246053abcd0f69177fe5f1
 - R version: 4.3.0
 - Operating system: macOS

krlmlr:
Introduced in efc8beccd496f275663826ed507dd76728a75952.

krlmlr:
This particular commit has been added to trigger the creation of the native igraph object. Maybe we need to be more careful with what we allow as attributes. Do we really need to support vertex/edge sequences as attributes?

krlmlr:
## CRAN igraph

``` r
options(conflicts.policy = list(warn = FALSE))
library(igraph)

gg <- make_ring(2, directed = TRUE)
V(gg)$id <- V(gg)
dput(gg)
#> structure(list(2, TRUE, c(0, 1), c(1, 0), c(0, 1), c(1, 0), c(0, 
#> 1, 2), c(0, 1, 2), list(c(1, 0, 1), list(name = "Ring graph", 
#>     mutual = FALSE, circular = TRUE), list(id = structure(1:2, class = "igraph.vs", is_all = TRUE, env = <weak reference>, graph = "17d54ab3-1bea-4723-8d5f-2e86ada40fe5")), 
#>     list()), <environment>), class = "igraph")
dput(as.list(unclass(gg)[[10]]))
#> list(me = structure(list(2, TRUE, c(0, 1), c(1, 0), c(0, 1), 
#>     c(1, 0), c(0, 1, 2), c(0, 1, 2), list(c(1, 0, 1), list(name = "Ring graph", 
#>         mutual = FALSE, circular = TRUE), list(), list()), <environment>), class = "igraph"), 
[...]

PR #808, opened 2023-05-21
feat: Avoid vertex and edge sequences in attributes

krlmlr:
For #807.


krlmlr:
This is branched off of v1.4.2, running revdepchecks now.

krlmlr:
No problems at all!

https://github.com/igraph/rigraph/pull/808/files#diff-bbdda7f28391ee0ec001638e5603609582124345e0d65ebd4747cc592f9bd56dR1

I suggest we include this into the upcoming 1.4.3 release, to avoid problems like #807.

ntamas (approved):

szhorvat:
> I suggest we include this into the upcoming 1.4.3 release, to avoid problems like #807.

Since 1.4.3 is published, how about a 1.4.4? If you do go ahead with a 1.4.4, can you please update the C core to the latest commit on this branch, to include an additional fix? The commit message explains the fix. https://github.com/igraph/igraph/tree/release/0.9

szhorvat:
Ah, I see these commits _were_ already included. I was confused by looking at the LTO checks on CRAN, but those are outdated, still showing information not even for 14.2, but for 1.4.1?  Why would that be so?

https://www.stats.ox.ac.uk/pub/bdr/LTO/igraph.out

Notice this warning:

```
core/community/leading_eigenvector.c:156:12: warning: ‘igraph_i_community_leading_eigenvector2’ defined but not used [-Wunused-function]
```

This is for the 1.4.1 sources, as this warning was already eliminated for 1.4.2.

krlmlr:
CRAN keeps asking if LTO issues have been fixed, I keep telling them that we fix in igraph > 1.4.99 because of the C library update. Nothing to worry about for now.

szhorvat:
These are not real issues. They are invalid warnings from GCC. As far as I'm concerned, this is a GCC bug.

krlmlr:
I've seen CRAN in one instance accept LTO warnings as GCC bugs, we'll see.

krlmlr:
Need to double-check if this is really needed or if perhaps there might be other ways to resolve this.

szhorvat:
With igraph 0.10, which has been significantly cleaned up, it will be much easier to show that these are likely GCC problem.

krlmlr:
After reviewing `` `[.igraph.vs`() ``, I am now confident that this is a good solution.

@mbojan: We do support arbitrary vectors, but vertex and edge sets create an unhealthy back reference that our code isn't prepared to deal with.

mbojan:
Thanks @krlmlr , that definitely makes sense to me. I already tried the version from the PR 👍

PR #820, opened 2023-05-27
fix: Rename `merges` to `cluster_merges` in output of `cluster_leading_eigen()`

### LINKED ISSUE #797: cluster_leading_eigen() output is incompatible with dendrogram construction
**Describe the bug**

This was reported at https://stackoverflow.com/q/76226275/695132

`cluster_leading_eigen()` produces a `merges` matrix in which the indices refer to clusters, not vertices. However, functions that convert to dendrograms still assume that these are vertex IDs, leading to wrong output.

The reason for this sort of output from `cluster_leading_eigen()` is that the algorithm generates splits (not merges) and produces a dendrogram that is incomplete from below (not incomplete from above, like in the case when we use other algorithms on disconnected graphs).

It seems to me that since the interpretation of the merges matrix returned by `cluster_leading_eigen()` is different than in the case of other community detection functions, it would be useful not to call it `merges` (also in the C core) and not to store it in `$merges`.

**To reproduce**

```
g <- make_graph(c(1,5, 1,6, 6,5, 5,4, 2,4, 4,3, 2,3), directed=F)
cl <- cluster_leading_eigen(g)
plot_dendrogram(cl)
```

image

image

Notice the incorrect dendrogram.

For reference, these are the produced clusters:

```
> groups(cl)
$`1`
[1] 1 5 6

$`2`
[1] 2 3 4
```

This is the produced merges matrix:

```
> merges(cl)
     [,1] [,2]
[1,]    1    2
```

Here `1` and `2` are community indices, not vertex indices.

**Additional information**

To further illustrate what it means for the dendrogram to be incomplete from below, here's the visualization I use in Mathematica:

image

It's essentially a dendrogram of communities, not of individual vertices.

**Version information**

```
R version 4.2.2 (2022-10-31)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.2/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:

other attached packages:
[1] igraph_1.4.99.9005

 [1] compiler_4.2.2  magrittr_2.0.3  cli_3.6.1       parallel_4.2.2 
 [5] tools_4.2.2     rstudioapi_0.14 Rcpp_1.0.10     nlme_3.1-162   
 [9] ape_5.7-1       grid_4.2.2      digest_0.6.31   rlang_1.1.0    
[13] pkgconfig_2.0.3 lattice_0.21-8 
```

krlmlr:
Closes #797.

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -1690,7 +1690,7 @@ cluster_leading_eigen <- function(graph, steps = -1, weights = NULL,
   res$algorithm <- "leading eigenvector"
   res$vcount <- vcount(graph)
   res$membership <- res$membership + 1
-  res$merges <- res$merges + 1
+  res$cluster_merges <- res$cluster_merges + 1
   res$history <- res$history + 1
   class(res) <- "communities"
   res
--- a/R/community.R
+++ b/R/community.R
@@ -1690,7 +1690,7 @@ cluster_leading_eigen <- function(graph, steps = -1, weights = NULL,
   res$algorithm <- "leading eigenvector"
   res$vcount <- vcount(graph)
   res$membership <- res$membership + 1
-  res$cluster_merges <- res$cluster_merges + 1
+  res$cluster.merges <- res$cluster.merges + 1
   res$history <- res$history + 1
   class(res) <- "communities"
   res
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -8431,7 +8431,7 @@ SEXP R_igraph_community_leading_eigenvector(SEXP graph, SEXP steps,
   SET_VECTOR_ELT(result, 4, eigenvalues);
   SET_VECTOR_ELT(result, 5, eigenvectors);
   SET_VECTOR_ELT(result, 6, history);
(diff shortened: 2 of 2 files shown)


szhorvat:
While function names now avoid dots, "key names" (what are the names that come after `$` called in R?) still typically use them. What's conventional in R? Is `cluster.merges` or `cluster_merges` better here?

krlmlr:
Element names? Names of a vector?

Let's stick with the existing convention, consistency is key. If written from scratch, we could use underscores, dots are fine too. Spaces or other non-syntactic characters would make things difficult.

PR #822, opened 2023-05-27
test: Merge tests

krlmlr:
@ntamas: The tests now fail. Do the make sense to you at all?


ntamas:
No, it doesn't. The tests are correct; in particular, dyad census should give a warning with undirected graphs (and now it does not give a arning); also, running a dyad census on a large empty graph should print an overflow warning because there are N*(N-1) empty dyads in an empty graph if the graph has N vertices. If the warnings are not printed, then this is incorrect behaviour.

ntamas:
Okay, now it's correct. `dyad_census()` was changed somewhere between igraph 0.9 and 0.10 so it indicates overflow by returning -1 so that's why the change is needed in the unit test.

ntamas (approved):

PR #827, opened 2023-05-29
docs: fix use of deprecated functions in examples

### LINKED ISSUE #825: Ensure that no deprecated functions are used in examples
Many examples use deprecated function names. For example, I see many uses of `barabasi.game()`. These should be updated.

@maelle Is there an easy way to detect all instances of these?

maelle:
Fix #825

@szhorvat This was semi-automatic: I automatically found calls to deprecated functions (only those from `R7zzz-deprecate.R` though) then replaced them manually.

@krlmlr Should this script stay in the package?


maelle:
The test failure actually appears on main too. :thinking:

szhorvat:
Very nice that this could be partly automated!

I didn't realise that we marked neighborhood() as a deprecated alias of ego(). "Ego" is a very social science centric description of the concept and IMO should not be the default name in a general purpose network network analysis process. Perhaps it would be better to keep both names as non-deprecated? What do you think @ntamas ?

For this PR, let's keep neighborhood() as is (not replace by ego()) and merge.

R/scan.R
szhorvat, line 1: So just revert this single modification and then merge

szhorvat (approved):

ntamas:
> What do you think @ntamas ?

This was a decision made by Gábor a long time ago and I kinda disagree with it. I'd un-deprecate `neighborhood()` but keep `ego()` as a permanent alias.

PR #832, opened 2023-06-05
fix!: Clean error message when trying to access a graph that has been saved with a newer or an older version of the package, `graph_version()` returns an object of class `"package_version"`

### LINKED ISSUE #830: check_version() and warn_version() claim that the graph has an *older* format even when in fact it's *newer*
`check_version()` and `warn_version()` in `versions.R` issue messages claiming that the graph format is too old whenever its format version does not match the current format version. This message is shown even if the format version is _newer_, not _older_.

This will become a problem if we bump the format version and keep the same system.

See also: https://github.com/igraph/rigraph/issues/829#issuecomment-1571586827

krlmlr:
Closes #830

### DIFF
--- a/R/basic.R
+++ b/R/basic.R
@@ -40,6 +40,7 @@ is_igraph <- function(graph) {
     return(FALSE)
   }
 
+  warn_version(graph)
   TRUE
 }
 
--- a/R/versions.R
+++ b/R/versions.R
@@ -24,6 +24,8 @@
 
 pkg_graph_version <- "0.8.0"
 
+pkg_graph_version_obj <- as.package_version(pkg_graph_version)
+
 #' igraph data structure versions
 #'
 #' igraph's internal data representation changes sometimes between
@@ -46,11 +48,11 @@ pkg_graph_version <- "0.8.0"
 #' @export
 graph_version <- function(graph) {
   if (missing(graph)) {
-    pkg_graph_version
-  } else {
-    stopifnot(is_igraph(graph))
-    .Call(R_igraph_graph_version, graph)
+    return(pkg_graph_version_obj)
   }
+
+  stopifnot(is_igraph(graph))
+  as.package_version(.Call(R_igraph_graph_version, graph))
 }
 
 #' igraph data structure versions
--- a/R/versions.R
+++ b/R/versions.R
@@ -51,7 +51,9 @@ graph_version <- function(graph) {
     return(pkg_graph_version_obj)
   }
 
-  stopifnot(is_igraph(graph))
+  # Don't call is_igraph() here to avoid recursion
+  stopifnot(inherits(graph, "igraph"))
+
   as.package_version(.Call(R_igraph_graph_version, graph))
 }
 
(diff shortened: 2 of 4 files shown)


szhorvat:
This changes the type of the object returned by the public and documented `graph_version()` function. Moreover, the kind of object returned compares differently than the strings that were previously used:

```
> as.package_version('0.10.0') < as.package_version('0.8.0')
[1] FALSE
> '0.10.0' < '0.8.0'
[1] TRUE
```

So this is a breaking change.

----

Of course, the string-based versioning that we used previously is problematic. But if change it, is it not better to use a plain integer, which is easy to compare either in R or C?  At the moment there is both R and C code that deals with format versions.  The format version isn't really exposed much to users, it's use is primarily technical.  Thus I'm not sure that there's any benefit is using multi-part versioning.

What multi-part versioning could achieve is a version bump that keeps the format interpretable by previous igraph. While not inconceivable, that is an unusual situation.  What information could be added to the format that can be ignored by older versions without loss?  The UUID that was added is one example (though its purpose is not 100% clear to me).  Another example would be a cache, similar to the one we use in the C core. A cache can be extended in the future, so even a three-part versioning might be useful for it.  But do we want to save the cache when saving graphs?  Well, is it possible to exclude parts of R object when serializing them to a file?  Perhaps saving it would be mandatory.

Sorry about thinking aloud while writing.

Yes, perhaps you're right and there's potential value in a multi-part format version after all. The disadvantage is the difficulty in handling from C.  Maybe it's good to fix at least the number of parts?

R/basic.R
```diff
@@ -36,7 +36,12 @@
 #' is_igraph(g)
 #' is_igraph(numeric(10))
 is_igraph <- function(graph) {
-  "igraph" %in% class(graph)
+  if (!inherits(graph, "igraph")) {
+    return(FALSE)
+  }
+
+  warn_version(graph)
+  TRUE
```
szhorvat, line 44: `is_igraph()` is called by `ensure_igraph()`, which in turn is called by (almost?) every public function.

So this way we get a version check in every function.

But if there's a version mismatch, the message will say:

> This graph was created by an old(er) igraph version.
> Call upgrade_graph() on it to use with the current igraph version
> **For now we convert it on the fly...**

Is the statement on the last line (bolded) accurate?

krlmlr: I've added tests in https://github.com/igraph/rigraph/pull/832/files#diff-58f022fa113e96e607f6d9f3816676cd2350b79cbc4546a8861543925cc670b3, but this is out of scope for this PR. These tests should trigger when we bump the version to 1.5.0 in #835.

krlmlr:
I'm not in love with the three-part version, the version "0.8.0" is meaningless and has only been introduced for igraph 1.0.0 (the version that gained the environment component). The `graph_version()` function is only used by us (and by the two Leiden packages that are breaking anyway), we could make that return an integer if 0.8.0 is found in the object.

krlmlr:
I backported the tests to the released version of igraph. This PR now shows how the behavior has improved over igraph 1.4.3.

Issue #819, opened 2023-05-24
cutoff=0 should be treated literally for betweenness and closeness calculations


szhorvat:
We promised that from igraph 1.4.0, `cutoff=0` would be treated literally in `betweenness()`, `edge_betweenness()` and `closeness()`, just as it already is in `harmonic_centrality()`.  But we did not make this change.  Let's update the warnings and docs to mention igraph 1.5.0 and do it for that version.

There is currently code like this:

```
  if (cutoff == 0) {
    warning("`cutoff' == 0 will be treated literally from igraph 1.4. If you want unrestricted betweenness, set it to -1")
    cutoff <- -1
  }
```

Docs say:

> The maximum path length to consider when calculating the betweenness. If zero or negative then there is no such limit.

> If cutoff is zero or negative, then the function calculates the exact betweenness scores. Using zero as a cutoff is deprecated and future versions (from 1.4.0) will treat zero cutoff literally (i.e. no paths considered at all). If you want no cutoff, use a negative number.

Fix this too.

Issue #818, opened 2023-05-24
Remove default value for 'loops' argument of centr_degree_tmax()


szhorvat:
We promised to remove the default value of the `loops` argument of `centr_degree_tmax()`, but did not actually do it.

```
  # Compatibility check with pre-igraph 1.3.0
  if (missing(loops)) {
    warning("centr_degree_tmax() will require an explicit value for its 'loops' argument from igraph 1.4.0. Assuming FALSE now.")
  }
```

Docs:

```
#' @param loops Logical scalar, whether to consider loops edges when
#'   calculating the degree. Currently the default value is `FALSE`,
#'   but this argument will be required from igraph 1.4.0.
```

I suggest updating the warning to mention 1.5.0 and doing it for that version.

szhorvat:
@krlmlr @maelle Can you update the milestone on this one and at least change the wording of the warning so it doesn't say that this was changed in 1.4.0 if it's still present in 2.0?

Issue #817, opened 2023-05-24
sample_pref() should reject non-symmetric matrices


szhorvat:
`sample_pref()` has the following warning:

```
warning("Undirected graphs require symmetric preference matrices, symmetrizing matrix. igraph 1.4.0 will reject non-symmetric matrices for undirected graphs.")
```

There is R code to symmetrize the matrix if it's not already symmetric.

This code and the warning should be removed, and the C core should be allowed to throw an error for non-symmetric matrices. This should have been done for 1.4.0. I suggest doing it immediately.

Issue #816, opened 2023-05-24
Remove 'fixed' parameter of layout_with_drl()


szhorvat:
This parameter has never been functional, and was promised to be removed in igraph 1.4.0, but this did not happen.

There's a warning:

```
warning("The `fixed` argument of `layout_with_drl` has no effect and will be removed in igraph 1.4.0.")
```

Docs:

> Logical vector, it can be used to fix some vertices. Unfortunately this has never been implemented in the C core of the igraph library and thus it never worked. The argument is now deprecated and will be removed in igraph 1.4.0.

I suggest immediate removal.

PR #842, opened 2023-06-15
chore!: Remove some behaviour / parameters that were deprecated in 1.3

szhorvat:
Fixes the following deprecated behaviours / parameters:

 - #814 
 - #819
 - #816
 - #817

I judge these to be very low risk, as these have all triggered warnings in the past, and some of the parameters were actually non-functional.

I would suggest also including #815 and #818 but those are slightly higher risk. However, warnings are already in place for these as well.

Requires revdepchecks.


krlmlr:
Reviewing this soon after the 1.5.0 release.

szhorvat:
Some of the docs text will need to be adjusted, as it mentions that the change was made in "1.5".

krlmlr:
Thanks. I'll pick it up from here and create individual PRs for your review.

maelle:
I see this affects the same functions as #852.
To deprecate arguments shouldn't we follow https://lifecycle.r-lib.org/articles/communicate.html#arguments? That's what you mean by individual PRs @krlmlr? I could do that from next week.

krlmlr:
Thanks, that would be great.

krlmlr:
Thanks. Running revdepchecks now.

krlmlr:
Thanks! No further action is needed here.

PR #846, opened 2023-06-15
fix: `types-RC.yaml`

Antonov548:


src/rinterface.c
```diff
@@ -7449,11 +7449,7 @@ SEXP R_igraph_from_prufer(SEXP prufer) {
 
   SEXP r_result;
                                         /* Convert input */
-  if (!Rf_isNull(prufer)) {
-    R_SEXP_to_vector_int_copy(prufer, &c_prufer);
-  } else {
-    IGRAPH_R_CHECK(igraph_vector_int_init(&c_prufer, 0));
-  }
+  R_SEXP_to_vector_int_copy(prufer, &c_prufer);
```
krlmlr, line 7452: Is this change intended?

Antonov548: Yes, this also needed

Antonov548: Input argument in this function couldn't be `NULL`

krlmlr:
Reapplied.

We would gain more confidence merging this if we had the SAN builds running in CI/CD.

PR #849, opened 2023-06-15
CRAN release v1.5.0

krlmlr:
- [x] Resolve all tasks in `cran-comments.md`
- [x] Run `fledge::release()`
- [ ] When the package is accepted on CRAN, run `fledge::post_release()`


szhorvat:
Before submitting this, please take a look at the chatroom. Tests are failing when run locally on my machine (tried two different and separated configurations), and I can't quite figure out what the difference is compared to CI.

PR #854, opened 2023-06-18
chore: ci sanitizer

Antonov548:
- add sanitizer step to ci based on https://github.com/krlmlr/rigraph-san docker image


Antonov548:
Tested in local repository. Step fails if memory leak appear. Also shows call stack of memory leak location.

Antonov548:
> What does a failing check look like? Can you please introduce a memory leak and open a PR so that we can review?

I checked it in my fork - https://github.com/Antonov548/rigraph/actions/runs/5303043517/jobs/9598406798

Antonov548:
https://github.com/igraph/rigraph/actions/runs/5310558329

PR #858, opened 2023-06-19
fix: avoid using of non-standard Fortran function ETIME

### LINKED ISSUE #857: Avoid etime in Fortran code
From CRAN:

etime is a GNU extension, used for two purposes:

i) To find the user and system CPU times.  For their sum (which is what one usually wants) there is (from F95) a standard replacement CPU_TIME.

igraph leidenbase svd all seem to use a LAPACK-derived function SECOND which returns user CPU time.  However, current LAPACK has several versions one of which uses CPU_TIME: see INSTALL/second_INT_CPU_TIME.f

ii) to measure elapsed time.  For that there is (from F90) a standard replacement SYSTEM_CLOCK.  This is what quantreg seems to want for GTIMER, but EXTETIME seems unused.  (It is not clear to me that the timer is actually used so maybe could simply return 0.0.)

gfortran has for many years supported F95 as well as GNU extensions by default so there is no problem using F90/F95 functions.  OTOH, flang (the compiler being considered for Windows-on-ARM) does not have etime and as there are standard equivalents is unlikely ever to do so.

Please review your usage and if possible change to a standard-compliant form.

szhorvat:
The timing function now always returns zero, since we don't need timings for igraph. Fixes #857

### DIFF
--- a/vendor/arpack/second.f
+++ b/vendor/arpack/second.f
@@ -11,22 +11,11 @@
 *  =======
 *
 *  SECOND returns the user time for a process in igraphseconds.
-*  This version gets the time from the system function ETIME.
+*  This version was changed in igraph to always return zero, since
+*  we do not need the timings. The original version used the
+*  non-standard ETIME, which is unsupported by flang.
 *
-*     .. Local Scalars ..
-      REAL               T1
-*     ..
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks!

PR #861, opened 2023-06-28
chore: Prepare move to igraph 0.10

krlmlr:
Extracted from #840.

@Antonov548: I wonder if we can do "small bets" here again: refactor code before actually moving to igraph 0.10. Would that approach work in principle?

I tried adding more commits from that PR, but compilation a40lready broke.

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -2723,6 +2723,17 @@ SEXP R_igraph_0orvector_to_SEXP(const igraph_vector_t *v) {
   return result;
 }
 
+SEXP R_igraph_0orvector_int_to_SEXP(const igraph_vector_int_t *v) {
+  SEXP result;
+  if (v) {
+    PROTECT(result=R_igraph_vector_int_to_SEXP(v));
+  } else {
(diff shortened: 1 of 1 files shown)


krlmlr:
Oh, the tests fail now. That's too bad.

Antonov548:
> Oh, the tests fail now. That's too bad.

I can check why tests are failed in this case. But I think it could be because R functions didn't change accordingly.

I would love to move by small pull request. But sometimes it's hard to seperate such small parts from the big list of commpilation errors. And I think it can slow down the process a lot.

Antonov548:
> Yes, I see this. Let's keep going with the big PR then.

Also I would like to run some static analyzer after fix coompilation errors to get rid of not used functions and declaration in `rinterface_extra.c`.

szhorvat:
Once you set up CMake, static analysers will be much easier to use. For example, CMake can generate Xcode project files, and you can use Xcode's Project -> Analyze feature, which presents the results in a very easy to understand format.

For the C core, we are currently using https://scan.coverity.com/, which is also very good.

PR #869, opened 2023-07-02
WIP: fix: Avoid storing attributes in global preserve list

krlmlr:
The current implementation seems to call `R_PreserveObject()` and `R_ReleaseObject()` way too often. This is R's global "preserve" list of objects that are never freed.

With the new implementation, we only add to the global list directly after initializing the attributes. As soon as we return the constructed object as an `SEXP` to R, the attributes are removed from the global preserve list.

The only problem I see is if we initialize the attributes and then exit with an error. The attribute vector will remain in the global preserve list forever, a memory leak. In fact, we should not use the global preserve list for that, but maintain or own list that we free unconditionally when returning to R.

This is only possible because we now check all igraph function calls and don't longjmp.


krlmlr:
Closing in favor of #870.

PR #870, opened 2023-07-02
fix: Change ownership rules of attribute objects

### LINKED ISSUE #866: Understand attribute hacks
What is the function of `R_igraph_attribute_protected` and `R_igraph_attribute_protected_size` ?

Introduced in response to https://github.com/igraph/igraph/issues/508.

Do we even need to call `R_PreserveObject()` and `R_ReleaseObject()` on the list that contains all attributes? It seems necessary currently, e.g., #865 only passed tests after [2a6e845](https://github.com/igraph/rigraph/pull/865/commits/2a6e845f04a38319555d554c744fe235a45af45c) has been added.

Need a better understanding of object lifetime, and perhaps more documentation.

krlmlr:
Add a global preserve list to which attribute objects can be added while no corresponding R graph object exists. This list is cleared before the next call to an igraph function.

This will only leak memory if a single igraph function call initializes attributes many times without actually using them, and only before a second igraph function call is made.

The commits can be reviewed individually.

Closes #866.

### DIFF
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -35,10 +35,10 @@ void R_igraph_interrupt(void);
 
 #define IGRAPH_R_CHECK(func) \
     do { \
+        R_igraph_attribute_clean_preserve_list(); \
         R_igraph_set_in_r_check(true); \
         igraph_error_type_t __c = func; \
         R_igraph_set_in_r_check(false); \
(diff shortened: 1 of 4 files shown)


Antonov548 (approved):

krlmlr:
Thanks! This was quick, we can still do a detailed review, commit by commit, after the fact.

PR #871, opened 2023-07-04
fix: `free_graph` use `igraph_free` to delete graph

Antonov548:
`R_igraph_to_SEXP` move graph data in external pointer -> graph data shouldn't be deleted after `R_igraph_to_SEXP` call


Antonov548:
@krlmlr the thing we missed: If we use pointer we shouldn't `free` graph in every function which creates new graph.

krlmlr:
Did you mean to close this?

Antonov548:
> Did you mean to close this?

Yes, I closed it. I get confused.

I think we should use `igraph_free` instead of deleting every field. But also I found that we deleting attributes after every conversion of graph with `IGRAPH_I_DESTROY`. I thought that's why it's necessary to assign `->attr`. But after  removing all this tests are still failing.

So, I will try to understand a bit more in details what happens and then create pull request.

PR #887, opened 2023-08-19
[draft] - check build system for igraph without cmake

Antonov548:


krlmlr:
Included in #840 now.

PR #894, opened 2023-09-18
docs: rm use of aliases

### LINKED ISSUE #892: Some functions missing from the function reference
The following functions are not listed in the function reference, and their signatures are not shown in their documentation pages. Do you have time to fix this @maelle ?

 - [ ] max_cliques
 - [ ] count_max_cliques
 - [ ] clique_num
 - [ ] clique_size_counts
 - [ ] largest_weighted_cliques
 - [ ] weighted_clique_num

Is it easy to get a list of non-deprecated function that are not in the function reference, so we can review them and decide if they should be included?

### LINKED ISSUE #958: largest_ivs
Similar to #892

maelle:
Fix #892 
Fix #958

@szhorvat @krlmlr I went through all the occurrences of aliases to find some other functions with the same issue (use of aliases instead of rdname, therefore absence of the signature of some functions in the manual pages), and related issues (useless but harmless use of aliases, that I removed; usage of family together with rdname tags which lead to duplicate seealso as seen in #948)

I removed `max_weighted_cliques()` and `count_max_weighted_cliques()` as they were defined as aliases but didn't exist.

TODOs

I lack info on: There is no documentation for the `...` argument of `clique_size_counts()`. What should it be?
The code is

```r
  if (maximal) {
    maximal_cliques_hist_impl(graph, min, max, ...)
  } else {
    clique_size_hist_impl(graph, min, max, ...)
  }
```


szhorvat:
> how did you create the list?

Manually, hence the mistaken inclusion of `max_cliques`.

szhorvat:
Could you please give me a reference for `@aliases` so I can better understand what it does?

maelle:
`@aliases` means two functions are synonyms. https://roxygen2.r-lib.org/articles/index-crossref.html?q=alias#aliases
Example https://github.com/tidyverse/tibble/blob/4de5c153ca5411fe2a02804a8cafb0edd9c664dc/R/tbl_df.R#L64

Now, if the goal is to document several functions in the same topic, the approach should be different otherwise the function signatures of all doesn't appear, as you noticed. https://r-pkgs.org/man.html#sec-man-multiple-functions

But should all those functions even be documented in the same page, or should each have its own manual page, that would link to the others (thanks to the `@family` tag)?

szhorvat:
> Now, if the goal is to document several functions in the same topic

Yes, this is what is being done here.

I guess `@aliases` doesn't work in this case as different functions documented on the same page should have different aliases, but there's only one `@aliases` declaration?



> But should all those functions even be documented in the same page, or should each have its own manual page, that would link to the others (thanks to the `@family` tag)?

It was done like this historically. I don't have a strong opinion on this.

It makes sense as there are groups of closely related functions.

`cliques`, `max_cliques` and `largest_cliques` all return cliques, and all take similar options which would ideally be documented once. Otherwise we need to duplicate the same text.

It's also nice to have a short text on what a "clique" is and we want to write this up once, for one doc page, not multiple.

How do you recommend we handle this @maelle ?

maelle:
> It's also nice to have a short text on what a "clique" is and we want to write this up once, for one doc page, not multiple.

Even if they were different manual pages, you could re-use the text without copy-pasting it, thanks to documentation inheritance. https://roxygen2.r-lib.org/reference/tags-reuse.html One can use `@inherit topic1 return` for instance to inherit the return value documentation from topic1.

> How do you recommend we handle this @maelle ?

I recommend using `@rdname` instead of `@alias`. But if any manual page is too long, or the functions have different arguments, I recommend using `@family` and documentation inheritance.

Based on that information, what topics would you group on the same page vs just in the same family?

szhorvat:
> Based on that information, what topics would you group on the same page vs just in the same family?

I thought I already replied here, sorry.

I suggest having a single page for:

 - Cliques, i.e. anything with `clique` in its name (and deprecated aliases); _excluding_ weighted cliques.
 - Weighted cliques, anything with `weighted_clique` in its name.
 - Independent vertex sets, anything with `ivs` in its name (and deprecated aliases).

See also #899 which could be fixed at the same time.

szhorvat:
@maelle Do you need more feedback from me on this?

maelle:
to be fixed up after #948 is merged as I see my changes include duplicate `@family` tags. :sweat_smile:

krlmlr:
@maelle: This might be ready to resume now, but let's perhaps close the other big PRs first?

maelle:
>  let's perhaps close the other big PRs first?

Which ones exactly do you mean?

maelle:
clique.number() is deprecated.

maelle:
`max_weighted_cliques()` is defined as an alias but does not exist. Same with  `count_max_weighted_cliques()`. Removing them in the next commit.

maelle:
@szhorvat @krlmlr I went through all the occurrences of aliases to find some other functions with the same issue (use of aliases instead of rdname, therefore absence of the signature of some functions in the manual pages), and related issues (useless but harmless use of aliases, that I removed; usage of family together with rdname tags which lead to duplicate seealso as seen in #948)

maelle:
@szhorvat @krlmlr I updated the PR description to clarify what I did in this branch, and to explain where I still need help (a small thing). Thank you!

ntamas:
I think this is just an example of defensive programming. If there will ever be more arguments, these would be transparently forwarded; other than that I don't think there are any arguments to think about right now. Feel free to remove the ellipsis if needed.

maelle:
So what's the preferred option, removing the ellipsis or documenting them as "used for future extensions"? Do we assume no one was using other argument so we could kill it without proper deprecation?

krlmlr:
`clique_size_counts()` gives an error now if called with extra argument. It's fine to remove the ellipsis I suspect.

krlmlr:
@szhorvat: Please set the "mergequeue" label when good.

PR #901, opened 2023-09-26
Update Stimulus to 0.17

krlmlr:
This doesn't look quite right to me. Can we tweak so that those optional C variables are initialized and `Rf_isNull()` is unnecessary in the call?


krlmlr:
Superseded by #997.

PR #904, opened 2023-10-20
feat: Warn if non-symmetric undirected adjacency matrix

krlmlr:
.

### DIFF
--- a/R/adjacency.R
+++ b/R/adjacency.R
@@ -22,13 +22,18 @@
 ##
 ## -----------------------------------------------------------------
 
-graph.adjacency.dense <- function(adjmatrix,
-                                  mode = c(
-                                    "directed", "undirected", "max",
-                                    "min", "upper", "lower", "plus"
-                                  ),
-                                  weighted = NULL, diag = TRUE) {
-  mode <- igraph.match.arg(mode)
+graph.adjacency.dense <- function(adjmatrix, mode, weighted = NULL, diag = TRUE) {
+  if (mode == "undirected") {
+    if (!isSymmetric(adjmatrix)) {
+      lifecycle::deprecate_soft(
+        "1.5.2",
+        "igraph::graph_from_adjacency_matrix(adjmatrix = 'must be symmetric if mode = \"undirected\"')",
+        details = 'Use mode = "max" to achieve the original behavior.'
+      )
+      mode <- "max"
+    }
+  }
+
   mode <- switch(mode,
     "directed" = 0,
     "undirected" = 1,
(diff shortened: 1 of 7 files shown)


szhorvat:
I'm confused. Why do the revdepchecks not show any new warnings?

I checked that e.g. manynets does indeed pass a non-symmetric matrix, which is why their tests fail with our 0.10 branch, see https://github.com/igraph/rigraph/blob/3f62aa557d188c9c41e7f2325ef43e8d87998e21/revdep/problems.md#manynet  Why does that not show up here in this PR?

szhorvat:
Why is there no warning with this?

```
> library(igraph)

Attaching package: ‘igraph’

The following objects are masked from ‘package:stats’:

    decompose, spectrum

The following object is masked from ‘package:base’:

    union

> m <- cbind(c(0,1), c(0,0))
> m
     [,1] [,2]
[1,]    0    0
[2,]    1    0
> graph_from_adjacency_matrix(m, mode='undirected')
IGRAPH c55e31d U--- 2 1 -- 
+ edge from c55e31d:
[1] 1--2
```

szhorvat:
@krlmlr I tracked down what is going wrong with the `pcalg` package in the revdepchecks.  The error was:

```
Error in UseMethod("isSymmetric") : 
  no applicable method for 'isSymmetric' applied to an object of class "c('double', 'numeric')"
Calls: jointIda ... graph.adjacency -> graph.adjacency.dense -> isSymmetric
```

This seemed strange: what object types are there that are accepted by `graph_from_adjacency_matrix` yet are not matrix-like enough to work with `isSymmetric`?

It turns out that `pcalg` does a subsetting of a matrix as `mat[idx, idx]` where `idx` is an index vector coming from a connected components calculation. In some cases there are single-vertex components, so `idx` is a length-1 vector, i.e. a scalar. Then `mat[idx, idx]` returns a scalar as well ... one of those cases when R tries to be convenient for interactive work, but ends up being a pain for writing reliable, robust libraries.

We need to make a decision here about whether we support scalars as trivial adjacency matrices, and be explicit about it, i.e. have a test for this case. I would cautiously say yes, it makes sense to support these, precisely because of R's matrix subsetting behaviour I mentioned above. But I'm not familiar enough with R to make a good decision here. What do you think?

szhorvat:
Other than the two comments above, this PR looks fine.

krlmlr:
Thanks @szhorvat. The downstream authors perhaps need `drop = TRUE` :

``` r
m <- diag(3)
m
#>      [,1] [,2] [,3]
#> [1,]    1    0    0
#> [2,]    0    1    0
#> [3,]    0    0    1
m[1:2, 1:2]
#>      [,1] [,2]
#> [1,]    1    0
#> [2,]    0    1
m[1, 1]
#> [1] 1
m[1, 1, drop = FALSE]
#>      [,1]
#> [1,]    1
```

<sup>Created on 2023-10-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

We *could* special-case for a scalar here, but I wouldn't. Can you please share a minimal-ish reprex that fails with this PR, preferably using the reprex package? We can then alert the maintainer. I couldn't reproduce it on my system.

szhorvat:
> We _could_ special-case for a scalar here, but I wouldn't.

Then that's decided. In fact I there's a chance that even though `graph_from_adjacency_matrix()` appeared to work on scalars, it might have overrun a buffer (since `dim(1)` is `NULL`).



> Can you please share a minimal-ish reprex that fails with this PR, preferably using the reprex package? We can then alert the maintainer. I couldn't reproduce it on my system.

I'll try.

szhorvat:
> Can you please share a minimal-ish reprex that fails with this PR

It no longer fails after your recent changes. It now works with scalars:

```
> igraph_version()
[1] "1.5.1.9011"
> graph_from_adjacency_matrix(3)
IGRAPH a052529 D--- 1 3 -- 
+ edges from a052529:
[1] 1->1 1->1 1->1
```

So do we keep supporting this or not?

PR #905, opened 2023-10-20
Bonpow weights

### LINKED ISSUE #903: power_centrality() silently ignores edge weights
**Describe the bug**

`power_centrality()`  cannot handle weighted graphs and so returns different values for a weighted graph and for an equivalent graph with parallel edges.

`sna::bonpow()`, on which `igraph::power_centrality()` is based, does not have this problem since it accepts a weighted adjacency matrix as an input, but it's not possible to do this with the `igraph` function since it uses `ensure_igraph()` to only allow an `igraph` object as input.  As a result, the only workaround to work with a weighted `igraph` is to convert the graph to something like an adjacency matrix (with weights in its cells) and then back into a graph with parallel edges instead of weights, which can then be passed to `power_centrality()` as in the reprex below.

The reason for this behavior is that [this line](https://github.com/igraph/rigraph/blob/b558296256bae51fa5be01ab8ce99ec65bfb1a73/R/centrality.R#L1080) in the function for dense matrices and [this one](https://github.com/igraph/rigraph/blob/b558296256bae51fa5be01ab8ce99ec65bfb1a73/R/centrality.R#L1108) in the function for sparse matrices use `as_adj()` to convert the igraph to an adjacency matrix, but they do not pass any arguments to `as_adj()`.  The bug could be easily fixed by allowing the name of the weight attribute to be passed to `as_adj()`'s `attr` argument, but I'm hesitant to submit a PR since it's not clear to me how the fix would/should affect downstream compatibility.  At a minimum, this unexpected behavior should be better documented.

**To reproduce**
```
library(igraph)
#> 
#> 
#>     decompose, spectrum
#> 
#>     union

incidence <- matrix(c(1, 0, 1, 0, 1, 1, 1, 1, 1), nrow = 3, ncol = 3)
bipartite <- graph_from_incidence_matrix(incidence)
gph_weighted <- bipartite_projection(bipartite, which = TRUE)
adj_weighted <- as_adj(gph_weighted, attr = "weight")
gph_unweighted <- graph_from_adjacency_matrix(adj_weighted)
adj_unweighted <- as_adj(gph_unweighted)

identical(adj_weighted, adj_unweighted)
#> [1] TRUE

identical(power_centrality(gph_weighted), power_centrality(gph_unweighted))
#> [1] FALSE
```

**Version information**

 - R/igraph version: 1.5.1
 - R version: R version 4.1.2 (2021-11-01)
 - Operating system: Ubuntu 20.04.6 LTS

```
> sessionInfo()
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.6 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:

other attached packages:
[1] igraph_1.5.1

 [1] compiler_4.1.2    Matrix_1.6-1.1    magrittr_2.0.3    cli_3.6.1        
 [5] tools_4.1.2       rstudioapi_0.15.0 unix_1.5.5        grid_4.1.2       
 [9] rlang_1.1.1       pkgconfig_2.0.3   lattice_0.20-45 
```

davidskalinder:
This should partially fix #903 (i.e., it should allow the R code to handle weights, but not provide a full C implementation).  I tried to keep all changes as minimal as possible and reuse logic from existing functions, especially for the behavior of the `weights` arg.

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -1304,12 +1304,12 @@ harmonic_centrality <- harmonic_centrality_cutoff_impl
 
 
 
-bonpow.dense <- function(graph, nodes = V(graph),
-                         loops = FALSE, exponent = 1,
+bonpow.dense <- function(graph, nodes = V(graph), loops = FALSE,
+                         exponent = 1, weights = NULL,
                          rescale = FALSE, tol = 1e-7) {
   ensure_igraph(graph)
 
-  d <- as_adj(graph)
+  d <- as_adj(graph, attr = weights)
   if (!loops) {
     diag(d) <- 0
   }
@@ -1328,7 +1328,8 @@ bonpow.dense <- function(graph, nodes = V(graph),
 }
 
 bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
-                          exponent = 1, rescale = FALSE, tol = 1e-07) {
+                          exponent = 1, weights = NULL,
+                          rescale = FALSE, tol = 1e-07) {
   ## remove loops if requested
   if (!loops) {
     graph <- simplify(graph, remove.multiple = FALSE, remove.loops = TRUE)
@@ -1337,13 +1338,13 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
   vg <- vcount(graph)
 
   ## sparse adjacency matrix
-  d <- as_adj(graph, sparse = TRUE)
+  d <- as_adj(graph, attr = weights, sparse = TRUE)
 
   ## sparse identity matrix
   id <- as(Matrix::Matrix(diag(vg), doDiag = FALSE), "generalMatrix")
 
   ## solve it
-  ev <- Matrix::solve(id - exponent * d, degree(graph, mode = "out"), tol = tol)
+  ev <- Matrix::solve(id - exponent * d, strength(graph, mode = "out"), tol = tol)
 
   if (rescale) {
     ev <- ev / sum(ev)
@@ -1408,6 +1409,13 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
 #'   loops.  `loops` is `FALSE` by default.
 #' @param exponent exponent (decay rate) for the Bonacich power centrality
 #'   score; can be negative
+#' @param weights A numerical vector or `NULL`. This argument can be used
+#'   to give edge weights for calculating the weighted eigenvector centrality of
+#'   vertices. If this is `NULL` and the graph has a `weight` edge
+#'   attribute then that is used. If `weights` is a numerical vector then it is
+#'   used, even if the graph has a `weight` edge attribute. If this is
+#'   `NA`, then no edge weights are used (even if the graph has a
+#'   `weight` edge attribute).
 #' @param rescale if true, centrality scores are rescaled such that they sum to
 #'   1.
 #' @param tol tolerance for near-singularities during matrix inversion (see
@@ -1463,13 +1471,21 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
 #' }
 #'
 power_centrality <- function(graph, nodes = V(graph),
-                             loops = FALSE, exponent = 1,
+                             loops = FALSE, exponent = 1, weights = NULL,
                              rescale = FALSE, tol = 1e-7, sparse = TRUE) {
   nodes <- as_igraph_vs(graph, nodes)
+
+  if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
+    weights <- "weight"
+  }
+  if (any(is.na(weights))) {
+    weights <- NULL
+  }
+
   if (sparse) {
-    res <- bonpow.sparse(graph, nodes, loops, exponent, rescale, tol)
+    res <- bonpow.sparse(graph, nodes, loops, exponent, weights, rescale, tol)
   } else {
-    res <- bonpow.dense(graph, nodes, loops, exponent, rescale, tol)
+    res <- bonpow.dense(graph, nodes, loops, exponent, weights, rescale, tol)
   }
 
   if (igraph_opt("add.vertex.names") && is_named(graph)) {
--- a/man/power_centrality.Rd
+++ b/man/power_centrality.Rd
@@ -9,6 +9,7 @@ power_centrality(
   nodes = V(graph),
   loops = FALSE,
   exponent = 1,
+  weights = NULL,
   rescale = FALSE,
   tol = 1e-07,
   sparse = TRUE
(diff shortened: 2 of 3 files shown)


R/centrality.R
```diff
@@ -1177,6 +1178,13 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
 #'   loops.  `loops` is `FALSE` by default.
 #' @param exponent exponent (decay rate) for the Bonacich power centrality
 #'   score; can be negative
+#' @param weights A numerical vector or `NULL`. This argument can be used
+#'   to give edge weights for calculating the weighted eigenvector centrality of
+#'   vertices. If this is `NULL` and the graph has a `weight` edge
+#'   attribute then that is used. If `weights` is a numerical vector then it is
+#'   used, even if the graph has a `weight` edge attribute. If this is
+#'   `NA`, then no edge weights are used (even if the graph has a
+#'   `weight` edge attribute).
```
szhorvat, line 1187: This describes the behaviour we want, but the function does not currently follow it.

R/centrality.R
```diff
@@ -1072,12 +1072,12 @@ harmonic_centrality <- harmonic_centrality_cutoff_impl
...
 
-bonpow.dense <- function(graph, nodes = V(graph),
-                         loops = FALSE, exponent = 1,
+bonpow.dense <- function(graph, nodes = V(graph), loops = FALSE,
+                         exponent = 1, weights = NULL,
                          rescale = FALSE, tol = 1e-7) {
   ensure_igraph(graph)
 
-  d <- as_adj(graph)
+  d <- as_adj(graph, attr = weights)
```
szhorvat, line 1080: The `attr` parameter of `as_adj` behaves differently from typical `weights` parameter of other functions, so this is not correct. See #906. I'd like to have this fixed during the ongoing major upgrade to C/igraph 0.10.

I'm thinking that this PR could be much simpler if we postponed it until after the 0.10 upgrade is complete, but that might take quite a while.

If we won't want to do this, the solution would be a workaround like that used in `alpha_centrality` (i.e. duplicate the graph object and add the weights as an attribute—yes, this is an ugly hack).

R/centrality.R
```diff
@@ -1105,13 +1106,13 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
   vg <- vcount(graph)
 
   ## sparse adjacency matrix
-  d <- as_adj(graph, sparse = TRUE)
+  d <- as_adj(graph, attr = weights, sparse = TRUE)
```
szhorvat, line 1109: Same issue as noted for the dense version.

R/centrality.R
```diff
@@ -1105,13 +1106,13 @@ bonpow.sparse <- function(graph, nodes = V(graph), loops = FALSE,
...
 
   ## sparse identity matrix
   id <- as(Matrix::Matrix(diag(vg), doDiag = FALSE), "generalMatrix")
 
   ## solve it
-  ev <- Matrix::solve(id - exponent * d, degree(graph, mode = "out"), tol = tol)
+  ev <- Matrix::solve(id - exponent * d, strength(graph, mode = "out"), tol = tol)
```
szhorvat, line 1115: We need to have consistency between the weights used during the adj. mat. conversion and during strength computation. Thus the weights must be passed explicitly here. We need to pay attention and make sure that we handle `NULL`, `NA` or a numeric vector exactly as described in the docs. `strength` might not handle its own `weights` parameter the same way, so we need to check what it does and adapt.

szhorvat (commented):
I don't program much in R, so the main review will be done by someone else. I'm thinking it's probably better to postpone this until after the 0.10 upgrade, as the code could be simpler.

R/centrality.R
szhorvat: This text still refers to the eigenvector centrality, that should be fixed.

davidskalinder:
Yes, I think you're quite right on all counts -- I stupidly didn't test all the weights conditions carefully enough and so missed the underlying cause of all of this, which is that `as_adj()` wants the weight attribute as a string whereas both `strength()` and most of the other centrality functions (correctly) want it as a vector.  My apologies for taking up your time with a lousy PR.

I could implement the workaround you suggest of using the weights logic (and documentation) from `alpha_centrality()`, kludgey though it may be.  (Indeed I think I might have found a bug there, too -- if I can confirm that then I'll report it separately.)

However it sounds like that workaround does create the problem that once `as_adj()`'s behavior changes, `power_centrality()`'s should too (i.e., they will both switch from accepting strings as weights to accepting vectors as weights).  If you guys would prefer to avoid creating that behavior that will soon change, then maybe it's best to fall back on the strategy of simply declaring in the documentation that `power_centrality()` doesn't support weights until it can do it with a vector argument?

So I think the options are for me to:

1. Apply the weights logic from `alpha_centrality()` to `power_centrality()`.
2. Leave `power_centrality()` without the ability to handle weights, and mention this in the docs.

Let me know which one you prefer and I can resubmit.

szhorvat:
> My apologies for taking up your time with a lousy PR.

Thanks for submitting the PR! It would indeed be nice to fix this eventually, and it's already drawing attention to issues with the adj. mat. conversion functions, which is a good thing.

> (Indeed I think I might have found a bug there, too -- if I can confirm that then I'll report it separately.)

Please do! 

> Let me know which one you prefer and I can resubmit.

We would need to wait for input from @krlmlr for this.

Personally I would suggest keeping this PR open for the moment, and finalizing it once the 0.10 update branch is in a good enough shape. That's going to be the most robust solution that takes the least amount of work. But I'm not sure when the 0.10 work will be in a good enough shape that it's ready for feature contributions like this one.

krlmlr:
We're on igraph/C 0.10 now, would you like to take another stab?

davidskalinder:
> We're on igraph/C 0.10 now, would you like to take another stab?

Ah, cool, thanks!  I'll have a look and see if that's something I can help with.

davidskalinder:
Hmm, looking at this again, I think the real issue is that the direction of this PR depends on #906, which I don't think has been resolved yet.  As far as I can tell, `as_adjacency_matrix()` still can't accept a `weights` vector (though it looks like the underlying function `get.adjacency.dense()` now can).  So as it stands, `power_centrality()` would still have to pass weights to `as_adjacency_matrix()` as an edge attribute name and to `strength()` as a vector.

So I think the options are still much as before (expanded here for clarity):

1. Wait until #906 has been resolved and then have `power_centrality()` accept a weights vector only.
2. If we want to improve things before that, we could do one of the following:
    1. Apply the kludgey weights logic from `alpha_centrality()` (with changes to handle #915) to `power_centrality()` now.
    2. Leave `power_centrality()` without the ability to handle weights for now, and mention this in the docs.

I might be overlooking something however; @szhorvat might have a better idea how these components should fit together?

krlmlr:
Thanks. I don't mind waiting, unless a kludge would solve a use case now. We might need a few more months to actually resolve the adjacency matrix issue linked above.

krlmlr:
Thank you for your contribution. Closing for now, will wait until after https://github.com/igraph/rigraph/issues/906. Adding a note to the motivating issue.

PR #911, opened 2023-10-22
feat: Improve symmetry check for adjacency matrices on input

krlmlr:
Matrices must be truly symmetrical, scalars are accepted.


szhorvat:
Be sure to bring it over to 0.10 where scalar support was broken again.

PR #927, opened 2023-10-29
test: More robust `sample_pa()` test

krlmlr:
Is this a reasonable assumption? Should we run an experiment? (I remember implementing this in my undergrad 🙃 )

### DIFF
--- a/tests/testthat/test-ba.game.R
+++ b/tests/testthat/test-ba.game.R
@@ -16,12 +16,14 @@ test_that("sample_pa works", {
 })
 
 test_that("sample_pa can start from a graph", {
-  set.seed(1234)
+  set.seed(20231029)
 
   g4 <- sample_pa(10, m = 1, algorithm = "bag", start.graph = make_empty_graph(5))
   expect_that(ecount(g4), equals(5))
   expect_that(vcount(g4), equals(10))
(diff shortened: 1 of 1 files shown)


szhorvat:
I don't get this: `expect_true(sum(degree(g4) == 0) %in% 1:4)`. Why can't the number of isolated nodes be zero? Why can't we get this graph?

image

I see through a numerical experiment that we do not get it, but I am confused.

@ntamas ?

krlmlr:
Thanks. A zero would be a surprise, we can still choose a different random seed if it occurs.

The table in the comments gives the absolute frequencies in an experiment I ran.

No strong feelings though.

szhorvat:
Similarly, this is not quite correct: `expect_true(sum(degree(g4) == 1) %in% 3:8)`. We can get between 2 and 10 vertices with degree 1:

image


image

I won't try to figure this one out: `expect_true(sum(degree(g4) %in% 2:3) %in% 0:3)`

I would suggest using a very simple test, such as checking that the number of edges is indeed 5.

krlmlr:
Agree that we want low-probability outcomes too. It weakens the test, so be it.

We're testing edge count. I extracted variables to add clarity to the code.

PR #930, opened 2023-10-29
feat: `all_shortest_paths()` returns the vertex list in both `res` and `vpaths` components

krlmlr:
Are we doing this only because in 0.10 there will be an `edges` component?

### DIFF
--- a/R/structural.properties.R
+++ b/R/structural.properties.R
@@ -638,6 +638,9 @@ all_shortest_paths <- function(graph, from,
     res$res <- lapply(res$res, unsafe_create_vs, graph = graph, verts = V(graph))
   }
 
+  # Transitional, eventually, remove $res
+  res$vpaths <- res$res
+
   res
 }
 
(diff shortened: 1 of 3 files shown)


szhorvat:
Yes.

But I am not sure that `vertices` ans `edges` are the best names. We want consistency between all shortest path functions. Others have `vpath` and `epath` instead. `functions.yaml` contains `vertices` and `edges`. Whatever choice is made, it should be consistent between the different path functions.

krlmlr:
Do we aim for consistency across all target languages?

szhorvat:
> Do we aim for consistency across all target languages?

No, personally I am against that.

I was referring to consistency between different functions in R.

krlmlr:
So, the plan would be to change all instances of `$vertices` to `$vpath`, and `$edges` to `$epath`, where appropriate?

krlmlr:
`all_shortest_paths()` is the only relevant instance. Changed to `$vpaths` because it's a list of vpath-s.

szhorvat:
> Conversely, we want an epaths component too.

This we cannot have in 1.6.0, as 0.9 does not provide the edge path, and the edge path cannot be reconstructed from the vertex path in case of multigraphs. It seems to me that we cannot do more in this PR.

krlmlr:
Yeah, let's do `$epaths` for 0.10.

PR #932, opened 2023-10-29
test: Stabilize `cluster_label_prop()` test

krlmlr:
Confirmed with an experiment.


szhorvat:
I think some of the fixes in the 0.10 series may not have been backported to 0.9. I don't remember exactly.

krlmlr:
What did we fix here in 0.10?

szhorvat:
Maybe nothing. It seems #1966 was backported, and the things that are mentioned in the changelog were regression fixes.

PR #934, opened 2023-10-29
feat!: `random_walk()` adds one more step, the resulting list of vertices is of length `steps + 1`

krlmlr:
.


szhorvat:
This is again changing in 0.10 in that there is a single function that returns both the edge and vertex paths. The edge path vector is one element shorter now. The original behaviour is still available, but deprecated.

ntamas (approved):

PR #935, opened 2023-10-29
feat: `is_tree(details = TRUE)` designates the first vertex as root for non-trees

krlmlr:
.


szhorvat:
Just keep in mind that the C function does not guarantee any meaningful value in `root` when the graph is not a tree, not even a value that would be a valid vertex ID.

PR #944, opened 2023-11-09
feat: Update C core

krlmlr:


krlmlr:
Was this the intended change from the C core update?

PR #952, opened 2023-11-10
feat: Rename arguments to `assortativity(values, values.in)` from `types1` and `types2`

krlmlr:


krlmlr:
This is the last interface incompatibility introduced in the igraph-0.10 branch, as far as I can tell. Merging it here to decouple interface changes from library updates.

@szhorvat: Please confirm that this is intended.

szhorvat:
Yes, intended.

szhorvat (approved):

krlmlr:
`assortativity_nominal()` still has `types`, should we update that too?

szhorvat:
> `assortativity_nominal()` still has `types`, should we update that too?

No, because those are actually types, i.e. it's a categorical variable.

PR #953, opened 2023-11-10
fix: `random_walk(weights = )` not available yet

krlmlr:


krlmlr:
Weights exist for the edge version.

PR #957, opened 2023-11-12
fix: Reduce frequency of interruption checks

szhorvat:
@krlmlr In the interest of making progress and getting 1.6.0 out soon, I propose that for now we just limit interruption checks to every `n`th time. Shall we merge this PR now, and look for better solutions later?

Refs #940, #942


krlmlr:
Thanks!

PR #969, opened 2023-11-18
chore: Ask users to use the reprex package

krlmlr:
As discussed. The reprex package can include the platform and session details if necessary, more often than not it's not needed and we can ask separately.


krlmlr:
Is the link helpful? Should we point to the "Usage" section indeed?

szhorvat:
It's fine ... what I expect though is that people will leave that text intact and will not use reprex.

Yes, the link is helpful. But we are asking people to read documentation and figure out steps. That's likely too much. I expect that the majority of people will _at most_ follow a clear step-by-step guide, which is not what the link points to.

But give it a try and see.

krlmlr:
Thanks, fair enough. We can borrow from here, it's really pretty: https://github.com/duckdb/duckdb/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml . I'll put something together.

vtraag:
I also think that it would be useful to give a very simple outline example of steps that you expect people to take. I'm just looking now at the link myself, and probably have to spend at least more than a minute to figure out exactly what you expect in the issue. (I've never used the package myself, and I rarely use R.)

szhorvat:
The following is just opinion, I'll leave it to you:

As a user of GitHub, I am not a fan of these extra-structured issue templates, as I find that there's always a situation that doesn't fit well. That's why I didn't push for enabling them for C/igraph. If you give people the chance to open a blank issue, they'll do that even more. If you don't, they can't easily open certain kinds of issue.

As for using `reprex`, I would suggest to write up steps to use it in a wiki page, and link to that page. What's important is that people should immediately see that using this package is: (1) easy (2) there are step-by-step instructions so no need to RTFM. If people feel that they need to read documentation to use it, they won't even look into how to do it.

krlmlr:
Let's see how the current implementation works out.

@szhorvat: happy to take your feedback. Is the form clear enough?

PR #970, opened 2023-11-20
docs: use {lifecycle} instead of zzz-deprecate.R

maelle:
Replacement for #716


maelle:
still a bit of work needed

maelle:
some bug in how I find the function signature e.g. 

> autocurve.edges: possible error in curve_multiple(graph = graph, start
  = start, x = x): unused argument (x = x)

maelle:
and many test failures

maelle:
usage is parsed well but not args :thinking:

maelle:
I'll come back to this later this week, removing error one by one until the script is perfect. :innocent:

maelle:
there are aliases on two lines

maelle:
with #971 + this now I only get test failures, let me see where they come from.

maelle:
next steps:

- rm test warnings
- methods see https://github.com/igraph/rigraph/pull/716#issuecomment-1827250315

maelle:
@krlmlr my commits should be informative for a change. :sweat_smile:

maelle:
@krlmlr reg bibliometrix, it's due to https://github.com/igraph/rigraph/commit/f95ecf39a99d576e23967e58b4eda62d770a15ae and was fixed in https://github.com/cran/bibliometrix/commit/cce90b1d2261bc991f00a40f79231859d46ea3dc

krlmlr:
This won't make it into 1.6.0, unfortunately. A few revdepcheck failures look like they could be fixed on our end, but it's also worth notifying downstream at some point.

I wonder if we could split the work in this PR. Can the script create one commit/branch per function that is deprecated? Ideally, the commits would be independent of one another.

maelle:
I'd need to think about how to make it work but most probably it'd have to be one commit per script not per function :thinking:

maelle:
the update of tests could happen even before all this work

maelle:
and one can't just freely move around the lines :thinking:

PR #993, opened 2023-12-11
refactor!: introduce lifecycle instead of current deprecations

maelle:
Good news: if 1.7.0 isn't the right version, I can more easily regenerate the PR now.

This PR consists of these commits
- two commits editing zzz-deprecate.R manually to directly remove two methods and to point two functions to better equivalents.
- a commit splitting zzz-deprecate.R into a script that defines `deprecated()` and into lines, for each deprecated calls, in the script defining the new function an old functions is deprecated to.
- many commits (independent from each other) editing a script for adding proper lifecycle deprecation, removing useless alias tags (since the old functions now have their proper manual page with the lifecycle badge on it), running `document()` so including changes to NAMESPACE and Rd files.
- one last commit removing the `deprecated()` function.

So one can only cherry-pick after having the three first commits.


maelle:
ah no rebase isn't going to work

maelle:
in truth https://github.com/igraph/rigraph/pull/993/commits/7c842c426980e4b0753b9a418c2b990c47d99c25 might need `document()` (after i rebased) but nothing too bad so I'll leave this as is.

krlmlr:
Thanks! Opened a PR for the first three commits. The next version will be 2.0.0.

R/structure.info.R
```diff
@@ -55,4 +69,3 @@ are_adjacent <- function(graph, v1, v2) {
   )
 }
 #' @export are.connected
```
krlmlr, line 71: Can we also remove this `#' @export` line?

R/operators.R
```diff
@@ -1195,14 +1284,3 @@ reverse_edges <- reverse_edges_impl
 #' @export
 t.igraph <- function(x) reverse_edges(x)
 #' @export graph.complementer
```
krlmlr, line 1286: Same here -- the `#' @export` lines below are gone, this should too?

krlmlr:
Baby steps! I hope we can start merging individual commits here soon.

Thanks!

R/structure.info.R
maelle: sort of embarrassing that this was an "off by one" error :zany_face: thanks for catching it @krlmlr

krlmlr: One of the two hard things in programming 🤣

PR #999, opened 2023-12-13
chore: Avoid deprecated `R_igraph_automorphisms()`

krlmlr:


szhorvat:
Does R have a bigint type, perhaps provided through a package? This function returns a potentially very large number, which is at the moment represented as a string. This can be inconvenient. I was just trying to use this function in a StackOverflow answer, to divide some counts by the size of the automorphism group. At the moment that's quite inconvenient due to the string representation. Is there a convenient way to improve on this, at least in cases when the count fits in a standard integer? Maybe it's good to have a second convenience function that just returns a numeric, even if not all digits can be represented accurately?

krlmlr:
Yes, e.g., https://davidchall.github.io/bignum/reference/biginteger.html . Stored as a string, IIUC.

Do we need the luxury of automatically returning as this data type, or is it sufficient to point the users to this data type?

PR #1005, opened 2023-12-18
docs: add missing param docs

maelle:


krlmlr:
Handing over to @szhorvat: does the documentation for the arguments make sense this way?

szhorvat:
No, the doc text does not make sense for `cluster_label_prop`. It does make sense for `which_mutual`, but even that is only a lucky accident and is likely to break in the future.

I am not very familiar with how R docs work, so please correct me. I looked up `@inheritParams` here: https://rdrr.io/github/hadley/roxygen3/man/InheritParamsTag-class.html I don't really like that it is completely unclear what parameters it is going to bring in. It is not obvious from looking at the doc sources of `cluster_label_prop` that it will bring in `mode`. That is a major risk.

`cluster_label_prop` and `sample_bipartite` are completely unrelated functions that do very different things. The `mode` parameter just happens to have the same name, but it has a very different meaning: how to interpret edges vs what edges to generate.

Similarly, `which_mutual` and `strength` are completely unrelated. In the C library, now there are two types of `loops` parameters: two-valued logical and three-values once/twice/none. While _at the moment_ both of these are two-values, I've been pushing for a while to change`strength` to three-valued.

It seems to me that if `@inheritParams` is used at all, it should only be used between very closely related functions such as `strength` and `degree`. Even then, using it means that we have to pay very careful attention to update the `loops` parameter of _both_ `strength` and `degree` at the very same time (which I think _should_ be done anyway, so that's perhaps not a bad thing).

szhorvat:
tl;dr 

 - Only inherit between closely related functions. 
 - If this opaque inheritance mechanism is used (i.e. it's not possible to see at a glance what is being inherited) then we must be very careful to change closely related functions in tandem and make sure that their features always match.

maelle:
are there functions we can inherit these param docs from?
if not, could you please suggest some text?
Thanks!

szhorvat:
Just got back from China, so I'm super overwhelmed at the moment, and will be a bit slow to respond.

> if not, could you please suggest some text?

Generally, the C docs are a good source for copying. The below are from there:

`which_mutual`'s `loops`: Do not inherit. Text: _"Logical, whether to consider directed self-loops to be mutual."_ 

`cluster_label_prop`'s `mode`:  _Whether to consider edge directions for the label propagation, and if so, which direction the labels should propagate. Ignored for undirected graphs. "all" means to ignore edge directions (even in directed graphs). "out" means to propagate labels along the natural direction of the edges. "in" means to propagate labels backwards (i.e. from head to tail)._

Note that the C docs also say, "It is advised to set this to "all" unless you are specifically interested in the effect of edge directions." but in R the current default is "out".

maelle:
> Just got back from China, so I'm super overwhelmed at the moment, and will be a bit slow to respond.

No worries!

Thank you, I opened an issue about the default value, and changed the docs in this PR. :pray:

szhorvat (approved):

PR #1007, opened 2023-12-18
refactor: ignore upstream CITATION file to avoid R CMD check NOTE

maelle:


krlmlr:
Thanks. Is there a way to turn off the noise in the `.cff` file, to just include major-minor-patch in that file? How to make sure it's updated when we release?

Issue #1010, opened 2023-12-18
Sub-WARNING


maelle:
```r
 graph.isomorphic.34: no visible global function definition for
     ‘isomorphic_34_impl’
   has_attribute_table_impl: missing arguments not allowed in calls to
     ‘.Call’
version_impl: missing arguments not allowed in calls to ‘.Call’
```

spin-off of #1008

@krlmlr @Antonov548

PR #1011, opened 2023-12-18
fix: argument quoting

### LINKED ISSUE #1008: unquoted param values
From R CMD check

```
N  checking R code for possible problems (45.8s)
   deterministic_optimal_imitation_impl: no visible binding for global
     variable ‘MAXIMUM’
   get_adjacency_sparse_impl: no visible binding for global variable
     ‘BOTH’
   get_adjacency_sparse_impl: no visible binding for global variable
     ‘ONCE’
   get_laplacian_impl: no visible binding for global variable
     ‘UNNORMALIZED’
   get_laplacian_sparse_impl: no visible binding for global variable
     ‘UNNORMALIZED’
   graph.isomorphic.34: no visible global function definition for
     ‘isomorphic_34_impl’
   has_attribute_table_impl: missing arguments not allowed in calls to
     ‘.Call’
   regular_tree_impl: no visible binding for global variable ‘UNDIRECTED’
   sparse_adjacency_impl: no visible binding for global variable
     ‘DIRECTED’
   sparse_adjacency_impl: no visible binding for global variable ‘ONCE’
   sparse_weighted_adjacency_impl: no visible binding for global variable
     ‘DIRECTED’
   sparse_weighted_adjacency_impl: no visible binding for global variable
     ‘ONCE’
   symmetric_tree_impl: no visible binding for global variable ‘OUT’
   tree_from_parent_vector_impl: no visible binding for global variable
     ‘OUT’
   version_impl: missing arguments not allowed in calls to ‘.Call’
   voronoi_impl: no visible binding for global variable ‘RANDOM’
   wheel_impl: no visible binding for global variable ‘OUT’
   Undefined global functions or variables:
     BOTH DIRECTED isomorphic_34_impl MAXIMUM ONCE OUT RANDOM UNDIRECTED
     UNNORMALIZED
```

I'll try to make a PR!

maelle:
Fix #1008
See also, unsolved here, #1010 

I could try generating aaa-auto.R from these changes but I'd first like to know whether this should be fixed more upstream, and whether there's something to tweak to make the parameter values lower-case in the R code.

@krlmlr @Antonov548

### DIFF
--- a/src/vendor/cigraph/interfaces/functions.yaml
+++ b/src/vendor/cigraph/interfaces/functions.yaml
@@ -100,13 +100,13 @@ igraph_create:
 
 igraph_adjacency:
     PARAMS: |-
-        OUT GRAPH graph, MATRIX adjmatrix, ADJACENCY_MODE mode=DIRECTED, LOOPS loops=ONCE
+        OUT GRAPH graph, MATRIX adjmatrix, ADJACENCY_MODE mode="DIRECTED", LOOPS loops="ONCE"
 
 igraph_sparse_adjacency:
     # adjmatrix is declared as INOUT because it might be modified during the
     # construction to eliminate duplicate elements from the representation
     PARAMS: |-
-        OUT GRAPH graph, INOUT SPARSEMAT adjmatrix, ADJACENCY_MODE mode=DIRECTED, LOOPS loops=ONCE
+        OUT GRAPH graph, INOUT SPARSEMAT adjmatrix, ADJACENCY_MODE mode="DIRECTED", LOOPS loops="ONCE"
 
 igraph_sparse_weighted_adjacency:
     # adjmatrix is declared as INOUT because it might be modified during the
@@ -121,10 +121,10 @@ igraph_weighted_adjacency:
         OUT EDGEWEIGHTS weights, LOOPS loops=ONCE
 
 igraph_star:
-    PARAMS: OUT GRAPH graph, INTEGER n, STAR_MODE mode=OUT, INTEGER center=0
+    PARAMS: OUT GRAPH graph, INTEGER n, STAR_MODE mode="OUT", INTEGER center=0
 
 igraph_wheel:
-    PARAMS: OUT GRAPH graph, INTEGER n, WHEEL_MODE mode=OUT, INTEGER center=0
+    PARAMS: OUT GRAPH graph, INTEGER n, WHEEL_MODE mode="OUT", INTEGER center=0
 
 igraph_square_lattice:
     PARAMS: |-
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks. The quoted values should be lowercase, I think.

@ntamas: what part of our infrastructure is supposed to turn these into R strings, or perhaps also enumerations of possible values as R vectors?

ntamas:
The default values in `functions.yaml` are abstract, but you may list "translations" between an abstract default value and a language-specific concrete default value in a language-specific `types.yaml` file like this (this is from `types-RR.yaml`):

```yaml
BOOLEAN:
    DEFAULT:
        'False': 'FALSE'
        'True': 'TRUE'
```

Normal YAML syntax rules apply so if you actually want quotes around a value, you need to encapsulate the quotes in another set of quotes.

If this is something that you want to do unconditionally for all default values in R, then the best place would be the RR code generator in Stimulus itself.

maelle:
> If this is something that you want to do unconditionally for all default values in R, then the best place would be the RR code generator in Stimulus itself.

Could one recognize the values that are supposed to be strings, to quote them and make them lowercase? Where exactly is the code generator?

@krlmlr do you think it's better to generalize or use the YAML?

ntamas:
> Could one recognize the values that are supposed to be strings, to quote them and make them lowercase?

I don't think so, and even if we come up with a ruleset we cannot guarantee that there will be no false positive matches in the future if we extend `functions.yaml` in the C core. In my opinion it's better to be explicit - you need to do this once per abstract type anyway and not once per function so it probably wouldn't be too much work. Look at `types-RR.yaml` in the existing source tree to see how it's done for other types.

maelle:
I am coming back to this a bit later. I do not understand where I would write

```yml
    DEFAULT:
        'ONCE': '"once"'
```

in `type-RR.yaml`. How do I know what type it is?

Or is it better to just edit `functions.YAML`? If I make ONCE lowercase there, how do I ensure it does not break anything?

ntamas:
`ONCE` corresponds to the `LOOPS` type; you can know it from the fact that it's specified like this in `functions.yaml` for many functions:

`...LOOPS loops=ONCE`

Here, `LOOPS` is the type, `loops` is the name of the argument and `ONCE` is the abstract default value. You should not need to touch or quote `ONCE`; instead of that, you should open `tools/stimulus/types-RR.yaml` and add something like this:

```yaml
LOOPS:
    DEFAULT:
        ONCE: '"once"'
```

ntamas:
Note that for many of the enums in `types.yaml`, we currently have something like this:

```yaml
EDGE_TYPE_SW:
     DEFAULT:
         SIMPLE: c("simple", "loops", "multi", "all")
         LOOPS: c("loops", "simple", "multi", "all")
         MULTI: c("multi", "simple", "loops", "all")
         ALL: c("all", "simple", "loops", "multi")
     INCONV: |-
         %I% <- switch(igraph.match.arg(%I%),
           "simple"=0L, "loop"=1L, "loops"=1L, "multi"=6L, "multiple"=6L, "all"=7L)
```

It seems like for enums we used to list all possible values in an R list, with the _real_ default being the first one, and then we used `igraph.match.arg(...)` to pick the numeric value. I don't know whether you want to keep on doing this or you have completely different plans in mind, just wanted to let you know.

PR #1014, opened 2023-12-18
refactor: use {lifecycle} for deprecation

maelle:
Fix #697
Fix #977

PR #1015, opened 2023-12-19
Rename resolution_parameter to resolution

### LINKED ISSUE #883: `cluster_leiden()` `resolution_parameter` should be renamed to `resolution`
The `resolution_parameter` parameter of `cluster_leiden()` should be renamed to `resolution`. Reasoning:

 - Consistency with other functions such as `modularity` and `cluster_louvain`
 - Brevity

CC @vtraag

vtraag:
Fixes #883. I've only updated the R file and the test, all the rest should follow automatically, if my understanding is correct? If more is needed, let me know.

### DIFF
--- a/R/community.R
+++ b/R/community.R
@@ -1290,6 +1290,7 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
 #' @param resolution The resolution parameter to use. Higher
 #'   resolutions lead to more smaller communities, while lower resolutions lead
 #'   to fewer larger communities.
+#' @param resolution_parameter  `r lifecycle::badge("superseded")` Use `resolution` instead.
 #' @param beta Parameter affecting the randomness in the Leiden algorithm.
 #'   This affects only the refinement step of the algorithm.
 #' @param initial_membership If provided, the Leiden algorithm
--- a/R/community.R
+++ b/R/community.R
@@ -1302,6 +1302,7 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
 #'   If this is not provided, it will be automatically determined on the basis
 #'   of the `objective_function`. Please see the details of this function
 #'   how to interpret the vertex weights.
+#' @inheritParams rlang::args_dots_empty
 #' @return `cluster_leiden()` returns a [communities()]
 #'   object, please see the [communities()] manual page for details.
 #' @author Vincent Traag
@@ -1346,7 +1347,7 @@ cluster_leiden <- function(graph, objective_function = c("CPM", "modularity"),
   check_dots_empty()
 
   if (lifecycle::is_present(resolution_parameter)) {
-    lifecycle::deprecate_soft("1.5.0.9006",
+    lifecycle::deprecate_soft("2.0.3",
                               "cluster_leiden(resolution_parameter)",
                               "cluster_leiden(resolution)")
     resolution <- resolution_parameter
--- a/R/community.R
+++ b/R/community.R
@@ -1287,7 +1287,7 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
 #'   weights. Set this to `NA` if the graph was a \sQuote{weight} edge
 #'   attribute, but you don't want to use it for community detection. A larger
 #'   edge weight means a stronger connection for this function.
-#' @param resolution_parameter The resolution parameter to use. Higher
+#' @param resolution The resolution parameter to use. Higher
 #'   resolutions lead to more smaller communities, while lower resolutions lead
 #'   to fewer larger communities.
 #' @param beta Parameter affecting the randomness in the Leiden algorithm.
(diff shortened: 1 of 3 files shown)


maelle:
I think we need to temporarily allow the two argument names? https://lifecycle.r-lib.org/articles/communicate.html#renaming-an-argument

vtraag:
Sure, sounds like a good idea!

I'm not entirely sure why the checks fail. Are the examples from the documentation not updated before being run or something? Should I run something in order to trigger an update of the docs?

maelle:
@vtraag ah yes you need to run `devtools::document()` to update the Rd files (the manual pages).

R/community.R
```diff
@@ -1088,8 +1088,18 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
...
 cluster_leiden <- function(graph, objective_function = c("CPM", "modularity"),
-                           weights = NULL, resolution = 1, beta = 0.01,
-                           initial_membership = NULL, n_iterations = 2, vertex_weights = NULL) {
+                           weights = NULL, resolution = 1,
+                           resolution_parameter = deprecated(), beta = 0.01,
+                           initial_membership = NULL,
+                           n_iterations = 2, vertex_weights = NULL) {
+
+  if (lifecycle::is_present(resolution_parameter)) {
+    lifecycle::deprecate_warn("1.5.0.9006",
```
maelle, line 1097: the version number here should be `2.0.0` (I think?) to go with the next release. it's only confusing for those who'll install the dev version.

maelle: should it be `lifecycle::deprecate_soft()` for this version?

vtraag:
> @vtraag ah yes you need to run `devtools::document()` to update the Rd files (the manual pages).

Unfortunately can't do that now, apparently have to install something still, but [R mirrors](https://cran.r-project.org/mirrors.html) seem to be down:

image

R/community.R
vtraag: > the version number here should be `2.0.0` (I think?) to go with the next release. it's only confusing for those who'll install the dev version.

Why would this be confusing? Happy to say 2.0.0 here, but when we release any pre-2.0 release with this in it, it would be incorrect, and might be more confusing. (I now realise that I might have targeted the dev branch due to breaking changes, but they're now no longer breaking). Or am I misunderstanding something here?

vtraag:
I've now updated the docs.

By the way, I'm not sure how the deprecation should be dealt with in the docs? Should the old argument still be there, saying it's deprecated, and point to the new argument? Are there any defaults for this in R (sorry, I almost never work in R).

R/community.R
```diff
@@ -1041,7 +1041,7 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
 #'   weights. Set this to `NA` if the graph was a \sQuote{weight} edge
 #'   attribute, but you don't want to use it for community detection. A larger
 #'   edge weight means a stronger connection for this function.
-#' @param resolution_parameter The resolution parameter to use. Higher
+#' @param resolution The resolution parameter to use. Higher
```
maelle, line 1044: you'd need to keep the two docs indeed but for the old one you can add a badge before the meaning https://lifecycle.r-lib.org/reference/badge.html

Example https://github.com/tidyverse/ggplot2/blob/a4be39dee07010c67c31ace059b1f0a24129fc65/R/plot.R#L44 (but here you probably want to use "superseded" not "deprecated" :thinking: )

vtraag: I believe this should now be OK?

krlmlr:
I've seen this before, igraph also implements a `deprecated()` function. Can we rename (in a separate PR), to avoid this kind of confusion?

vtraag:
Can we for now use `lifecycle::deprecated` ? What is the `igraph::deprecated` function used for, and do we need to retain it at all? (Separate PR and/or issue might be good indeed).

maelle:
@vtraag the igraph `deprecated()` function will soon be removed, by #1014 and #1104 which aren't ready yet (and thanks to other commits).

krlmlr:
@maelle: I forgot why this is assigned to me. We probably want to add an ellipsis after the second argument and run revdepchecks. Can you please help with the ellipsis?

maelle:
Sure, I'll do that, but what's our rationale for choosing to put `check_dots_empty()` in auto-generated vs wrapper code?

maelle:
the existence of igraph's own deprecated() remains problematic

R/community.R
```diff
@@ -1330,12 +1331,27 @@ cluster_spinglass <- function(graph, weights = NULL, vertex = NULL, spins = 25,
...
+    # FIXME: change to deprecated() once we have @importFrom lifecycle deprecated,
+    # after igraph:::deprecated() is removed
+                           resolution_parameter, beta = 0.01,
+                           initial_membership = NULL,
+                           n_iterations = 2, vertex_weights = NULL) {
+
+  check_dots_empty()
+
+  if (lifecycle::is_present(resolution_parameter)) {
+    lifecycle::deprecate_soft("1.5.0.9006",
```
krlmlr, line 1349: This now needs to be 2.0.x

krlmlr (approved):
Nice! Good to go from my end.

PR #1021, opened 2023-12-23
fix: Intersections between edges and rectangle vertices are computed correctly

### LINKED ISSUE #501: Visualization: Edges may end or start at the wrong position with rectangular vertices
**Describe the bug**

With rectangular vertices, edges may end of start at the "opposite" size of the vertex

In this example, 1 -> 2 starts at the wrong place and 1 -> 4 ends at the wrong place. To see this, it is important to set the vertex colour to semi-transparent, so the edges are visible behind the vertices.

image

The bug was noted here:

https://stackoverflow.com/q/70506672/695132

**To reproduce**

Try different random seeds. It is easy to reproduce the issue, but be sure to set vertex colours to transparent.

```
set.seed(1234)
g <- erdos.renyi.game(5, 6, 'gnm', directed=T)
plot(g, 
     vertex.size=30, 
     vertex.color=rgb(0.1,0.7,0.8,0.1), 
     vertex.shape="rectangle")
```

**Version information**

1.2.10

dmurdoch:
Fixes #501.

The `l` matrix here should contain the intersections of edges with the 4 sides of the rectangular node. but due to a typo, one of the intersections was computed incorrectly.

### DIFF
--- a/tests/testthat/test-bug-501-rectangles.R
+++ b/tests/testthat/test-bug-501-rectangles.R
@@ -1,0 +1,13 @@
+
+test_that("Edges stop at outside of rectangle node", {
+  rectangle_edges <- function() {
+    set.seed(1234)
+    g <- erdos.renyi.game(5, 6, 'gnm', directed = TRUE)
+    plot(g,
+       vertex.size = 30,
+       vertex.color = rgb(0.1, 0.7, 0.8, 0.1),
+       vertex.shape = "rectangle"
+       )
+  }
+  vdiffr::expect_doppelganger("rectangle-edges", rectangle_edges)
+})
--- a/tests/testthat/test-bug-501-rectangles.R
+++ b/tests/testthat/test-bug-501-rectangles.R
@@ -1,7 +1,7 @@
 
 test_that("Edges stop at outside of rectangle node", {
   rectangle_edges <- function() {
-    g <- graph(c(1,2, 1,4, 2,1, 2,5, 2,3, 4,1, 5,2, 3,2))
+    g <- make_graph(c(1,2, 1,4, 2,1, 2,5, 2,3, 4,1, 5,2, 3,2))
     layout <- cbind(c(-2.01, -1.16, -1.24, -2.74, -0.13),
                     c(1.27, 2.1, 3.14, 0.56, 2.01))
     plot(g,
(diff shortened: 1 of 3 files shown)


szhorvat:
Thanks so much @dmurdoch !

@krlmlr, in case it helps, I tested the fix on several example. I think it is ready to merge.

dmurdoch:
Would you like me to add a test? I have one ready that should fix the codecov failure.

szhorvat:
If you have it ready anyway, it'd be nice to add it.

It may take a little while to merge this PR due to the holiday season, but it will certainly be merged :-) Thanks again!

tests/testthat/test-bug-501-rectangles.R
```diff
@@ -0,0 +1,13 @@
+
+test_that("Edges stop at outside of rectangle node", {
+  rectangle_edges <- function() {
+    set.seed(1234)
+    g <- erdos.renyi.game(5, 6, 'gnm', directed = TRUE)
```
krlmlr, line 5: The tests will be stabler in the long term if we added an explicit construction of `g`, and also for the layout, here and for the other plotting tests. This feels out of scope for this buglet because we seem to lack the necessary infrastructure to do this efficiently.

krlmlr (approved):
Thanks, good catch!

tests/testthat/test-bug-501-rectangles.R
dmurdoch: Sorry, it's not clear to me if you're asking me to change the test or to drop it.  If you'd like everything explicit I can do that.

szhorvat: If it's kept this way, `erdos.renyi.game` should be replaced with `sample_gnm` (as the former is now deprecated).

tests/testthat/test-bug-501-rectangles.R
```diff
@@ -0,0 +1,15 @@
+
+test_that("Edges stop at outside of rectangle node", {
+  rectangle_edges <- function() {
+    g <- graph(c(1,2, 1,4, 2,1, 2,5, 2,3, 4,1, 5,2, 3,2))
```
szhorvat, line 4:
```suggestion
    g <- make_graph(c(1,2, 1,4, 2,1, 2,5, 2,3, 4,1, 5,2, 3,2))
```

`graph` happens to be another deprecated/undocumented construct :-)

krlmlr:
Thanks! I wonder why the code coverage doesn't change.

dmurdoch:
Thanks for merging it!  I wondered about the coverage too -- it definitely uses that function.  Maybe because the change was in just one line, and the expression it was in took up several lines?

PR #1027, opened 2023-12-28
Update parsers and remove STR

szhorvat:
This PR updates generated parsers. For some reason there's a line number change, but this is not functional. There is an important change in the DL parser.

Additionally, it removes uses of the now deprecated STR() macro.


krlmlr:
Now in #1034.

krlmlr:
Thanks!

PR #1028, opened 2023-12-28
feat: New `k_shortest_paths()` to compute the first k shortest paths between two vertices

### LINKED ISSUE #972: Add k_shortest_paths
The igraph core contains a function to compute k shortest paths with Yen's algorithm since March 2022. I noticed it has also been made available in the Python package of igraph recently. Is there any plan to make this available also in the R package?

I could try to start a PR based on the logic of the regular shortest_paths function, but have almost no experience in wrapping C code. Therefore the question if this is a planned addition already.

**Use cases for the feature**
Finding more "good" paths then only the shortest path, but without the need to compute all simple paths and subset them (which is very time consuming).

**References**
[Issue on the igraph core repo about this function](https://github.com/igraph/igraph/issues/1397) 
[Python docs for this function](https://python.igraph.org/en/stable/api/igraph.GraphBase.html#get_k_shortest_paths)

szhorvat:
Fixes #972 by adding `k_shortest_paths()`.

This is a draft, but it should be reviewed now. Keep in mind that I don't actually know much R! In fact I wouldn't mind if someone took this over or helped out.

The reason why I did not add this to the doc page of `shortest_paths()` is that the `to` parameter is different: it must be a single vertex.

@maelle:

 - I tried to use `@inheritParams` here. 
 - I'm not sure if it's normal that so many other functions' doc pages got references to this one?

@krlmlr:

I did not attempt to add an `output` parameter like `shortest_paths` has. Both edge and vertex paths are returned unconditionally.

I hope to have this much-requested feature in 2.0.0.

CC @luukvdmeer


szhorvat:
@krlmlr This does not currently work and I need some advice on how to fix it.

```r
> k_shortest_paths(make_ring(5), 1, 2, k=3)
  'arg' must be of length 1
> traceback()
5: stop("'arg' must be of length 1")
4: match.arg(arg = arg, choices = choices, several.ok = several.ok)
3: igraph.match.arg(mode)
2: get_k_shortest_paths_impl(graph = graph, weights = weights, k = k, 
       from = from, to = to, mode = mode)
1: k_shortest_paths(make_ring(5), 1, 2, k = 3)
```

szhorvat:
@ntamas OK, I found `PARAM_ORDER`, but that's still not sufficient to avoid having a wrapper here. I need to rename the `vertex.paths` and `edge.paths` components of the result to simply `vpaths` and `epaths`. Do we have a stimulus feature for this?

szhorvat:
Never mind, I just realized that `PARAM_NAMES` works since these still _parameters_, even though output parameters.

szhorvat:
@ntamas Is there a way to set default parameter values just in R when they are not set in the C core? What if I'd like to set a default of `k=1` in this function?

ntamas:
> Is there a way to set default parameter values just in R when they are not set in the C core?

No, not at the moment, apart from copying the entire parameter specification of the function from the C core and adding the default value there. We can implement something like this in Stimulus, although the ideal solution would be to define the default value in the C core.

R/aaa-auto.R
```diff
@@ -819,14 +819,14 @@ get_k_shortest_paths_impl <- function(graph, weights=NULL, k, from, to, mode=c("
   res <- .Call(R_igraph_get_k_shortest_paths, graph, weights, k, from-1, to-1, mode)
```
krlmlr, line 819: I wonder why we have `-1` here but no inverse correction for the output. Is this expected?

R/aaa-auto.R
```diff
@@ -792,7 +792,7 @@ voronoi_impl <- function(graph, generators, weights=NULL, mode=c("out", "in", "a
   res
 }
 
-get_k_shortest_paths_impl <- function(graph, weights=NULL, k, from, to, mode=c("out", "in", "all", "total")) {
+get_k_shortest_paths_impl <- function(graph, from, to, k, weights=NULL, mode=c("out", "in", "all", "total")) {
```
krlmlr, line 795: The following would be safer:

```suggestion
get_k_shortest_paths_impl <- function(graph, from, to, ..., k, weights=NULL, mode=c("out", "in", "all", "total")) {
  check_dots_empty()
```

This would require users to supply names for `k` and the other optional arguments. Positional matching makes it harder:

- for users to skim the code (what does that fifth argument mean again?),
- for us to add new arguments (all new ones must go at the end to maintain full compatibility, which is not always desired).

I wonder if this could be part of Stimulus, or if we create a wrapper here (and in many other places).

krlmlr (commented):
Thanks. What would a test look like here?

R/aaa-auto.R
ntamas: I don't have the code in front of me right now, but I remember that there aren't any clear guidelines on where to do the +1/-1 adjustments for vertex/edge indices and sometimes it's done in the output conversion of the result. For instance, there is `R_igraph_vector_int_to_SEXP` to convert an `igraph_vector_int_t` back to R without adjustment and `R_igraph_vector_int_to_SEXPp1` to convert it back _with_ adjustments. (The `p1` suffix stands for "plus one").

Ideally, there should be guidelines on where the adjustment is done. If it is easier to do in R and there are no performance implications (like an unnecessary copying of the vector), it should probably be done in R and we could get rid of the `p1` functions in `rinterface_extra.c`.

R/aaa-auto.R
ntamas: Stimulus does not have a way to indicate a preference for keyword arguments vs positional arguments, but I guess it could be added. Python does it by adding a lone `*` in the argument list (e.g., `def compare(a, b, *, key=None):` makes `key` a keyword argument only while `a` and `b` could be defined either as positional or as keyword arguments). We could adapt something similar in Stimulus if this is a concept that seems to apply to multiple languages. Languages without kwarg support could ignore the asterisk in the argument list.

Another alternative would be to add a Stimulus keyword that specifies the name of the first argument that should be treated as keyword-only (it would be `k` here), but this could cause problems if a higher level interface decides to re-order arguments.

R/aaa-auto.R
krlmlr: There are performance implications, ideally we'd convert everything in C.

R/aaa-auto.R
krlmlr: I like the star notation. Can you please propose something, @ntamas?

ntamas: I'll come up with a patch in Stimulus and we'll see how the idea works out.

ntamas: See the Stimulus revision tagged with 0.21.0.

There are now three alternative ways to indicate keyword-only arguments:

* You can add `KW` in front of the type of the argument in the `PARAMS` key to mark an argument as keyword-only.
* You can add `*` in the middle of the argument list at the point where the keyword-only arguments start.
* You can add a key named `FIRST_KW_PARAM` next to `PARAMS` and it must contain the name of the first keyword-only argument. This is primarily meant for backward compatibility with Stimulus 0.20 and earlier as the other two options would fail on earlier versions of Stimulus.

I tried to update the RR code generator to respect keyword-only arguments, but I haven't tested it. If it does not work, look in `RRCodeGenerator`, this is where the magic happens.

Next up would be to update `functions.yaml` in the C core to indicate which arguments should be keyword-only in generated interfaces.

R/aaa-auto.R
krlmlr: Dreaming big: If we could get views that cast to int and perform the "minus one" on the fly, in the C core...

krlmlr:
I'm massaging the `.yaml` in other PRs, there will be more conflicts.

krlmlr:
After #1043, there should be no more conflicts for now.

szhorvat:
> What would a test look like here?

Yes, I will add tests.

szhorvat:
A good test would be:

```
g <- make_ring(5)
res <- k_shortest_paths(g, 1, 2, k = 3)
```

@krlmlr How do I check that `res` has the expected value (since it's just just a number, but a more complicated expression)?

szhorvat:
Conflicts resolved.

R/aaa-auto.R
szhorvat: If you want to make `k` a mandatory keyword argument, I would rather specify this as a custom parameter list for the R interface, and also give a default of `k=1`. _In this specific case_, I wouldn't mark it as mandatory keyword in the C core.  Shall I go ahead?

While this feels natural in R, it wouldn't be so in Mathematica. Not sure about Python. Note that when using this function, people would always want to specify `k`.

szhorvat: @ntamas How does the `*` work with parameter reordering? Is it allowed to be used in `PARAM_ORDER`?

ntamas: > Is it allowed to be used in PARAM_ORDER?

No, only in `PARAMS`. It's a neat idea, though. I'll see what can be done.

> How does the `*` work with parameter reordering?

In the parsed parameter specification, each parameter has a property named `is_keyword_only` that specifies whether the parameter is to be used as a keyword argument only if the language supports keyword arguments. The `*` is essentially a shortcut for specifying this for an argument and all subsequent arguments automatically. This comes before the parameter reordering.

Note that in theory, one could end up in a situation where keyword-only and non-keyword-only arguments are intermixed in the parameter ordering. It is up to the language-specific code generator to deal with such a situation. The RR code generator collects non-keyword-only args first and keyword-only args second (but respects `PARAM_ORDER` within each sublist). The Python code generator throws an error because it was easier to do it that way.

ntamas: `PARAM_ORDER` now supports `*` from Stimulus 0.21.2. FWIW it also supports `...`, which means "all the parameters that have not been listed explicitly so far, in their original order".

krlmlr:
I would do something like:

```r
expect_length(res$vpaths, 2)
expect_equal(as.numeric(res$vpaths[[1]]), c(1, 2))
expect_equal(as.numeric(res$vpaths[[2]]), c(1, 5, 4, 3, 2))
...
```

tools/stimulus/functions-R.yaml
```diff
@@ -374,6 +374,10 @@ igraph_get_all_simple_paths:
       PP: get.all.simple.paths.pp
 
 igraph_get_k_shortest_paths:
+    PARAM_ORDER: graph, from, to, *, k, ...
```
krlmlr, line 377: I wonder why we need to override this here. No strong opinion.

krlmlr (approved):
Very nice, thanks!

tools/stimulus/functions-R.yaml
szhorvat: I did this to be able to make `weights` optional. This function will often be used on unweighted graphs. Is there a better way to do this?

krlmlr: It's fine, but is this order something to be considered upstream too? To me, this order seems reasonable for other interfaces too.

szhorvat: Well, yes, but probably too late for that.  If we ever reorder arguments in all functions in the C core, this would made sense, but it's unlikely that'll happen. When we settled on the argument ordering, we forgot about the consequences for auto-generation. CC @ntamas

https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering

krlmlr: Interesting.

In R land, we have the attribute. R objects are immutable, it's easy to remove weights, or to set new weights, without changing the graph object. I wonder if we even need the `weights` argument.

PR #1026, opened 2023-12-28
refactor: clean up some auto-generation code and remove unused parts

szhorvat:
There was a lot of unused code that this PR removes. Tests seem to pass by I am not 100% confident, so have a careful look.

There are some minor cleanups as well, such as using the correct types (integer/real).

PR #1031, opened 2023-12-29
refactor: clean up some auto-generation code

krlmlr:
From #1026.

### DIFF
--- a/tools/stimulus/types-RR.yaml
+++ b/tools/stimulus/types-RR.yaml
@@ -25,7 +25,7 @@ BOOLEAN:
 
 INDEX_VECTOR:
     # Indices in R are 1-based; the conversion takes place here
-    INCONV: '%I% <- as.numeric(%I%)-1L'
+    INCONV: '%I% <- as.numeric(%I%)-1'
 
 VECTOR:
     DEFAULT:
@@ -206,11 +206,11 @@ VERTEX_COLOR:
           }
         }
         if (!is.null(%I%)) {
-          %I% <- as.numeric(%I%)-1L
+          %I% <- as.numeric(%I%)-1
         }
     OUTCONV:
         OUT: |-
-            %I% <- %I%+1L
+            %I% <- %I%+1
             if (igraph_opt("add.vertex.names") && is_named(%I1%)) {
               names(%I%) <- vertex_attr(%I1%, "name")
             }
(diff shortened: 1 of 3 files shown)


tools/stimulus/types-RR.yaml
```diff
@@ -297,7 +297,7 @@ VERTEXSET_LIST:
 
 VERTEXSETLIST_INT:
     INCONV:
-        IN: '%I% <- lapply(%I%, function(x) as.integer(x)-1L)'
+        IN: '%I% <- lapply(%I%, function(x) as.numeric(x)-1)'
```
krlmlr, line 300: This looks like a bug fix?

tools/stimulus/types-RR.yaml
```diff
@@ -258,11 +258,11 @@ EDGE_COLOR:
           }
         }
         if (!is.null(%I%)) {
-          %I% <- as.integer(%I%)-1L
+          %I% <- as.numeric(%I%)-1
```
krlmlr, line 261: This looks like a bug fix?

krlmlr (commented):
@szhorvat: The test behavior hasn't changed in the presence of bug fixes (?). Can you please point me to the relevant test cases in the C core?

tools/stimulus/types-RR.yaml
szhorvat: This is a readability fix. Now we use doubles for indices, so we should be consistent, and avoid integers. Whenever I see `1L` I'm thinking about whether there's a corresponding `int` on the C side that can overflow or otherwise prevent large indices.

szhorvat: This was a mistake on my part. `VERTEXSETLIST_INT` is unused and should be completely removed.

szhorvat: See #1049

tools/stimulus/types-RR.yaml
szhorvat: Yes it is. `R_SEXP_to_vector_int_copy()` should expect a real, not an integer.

szhorvat: Here's a test for this:

```
count_isomorphisms(make_ring(3), make_ring(3), edge.color1=c(2,2,2), edge.color2=c(2,2,2), vertex.color1=c(3,3,3), vertex.color2=c(3,3,3), method='vf2')
```

This crashed before this fix. Could you please help me by adding this to the test suite?

szhorvat: Never mind, it was easy enough. See #1050

PR #1032, opened 2023-12-29
refactor: remove unused parts from code generation

krlmlr:
From #1026.


szhorvat:
This was just a first step, focusing on types. There's a lot more that could be removed later. I did this to get some clarity, so we can continue the refactoring more effectively.

PR #1033, opened 2023-12-29
refactor: Replace loop by `lapply()` when returning vertex and edge sets 

krlmlr:
Prompted by #1028.


krlmlr:
Do the names of the return values in the affected functions make sense to you?

PR #1034, opened 2023-12-29
fix: remove use of deprecated `STR()`

krlmlr:


krlmlr:
@szhorvat: Do we have no tests for string attributes? Should we add a test that reads a DIMACS file?

szhorvat:
> Do we have no tests for string attributes?

I don't know if it's individually tested, but several tests do use string attributes. All tests that add names to vertices do.

Testing string attributes is unrelated to this PR. This PR does not need tests.

DIMACS files are not a good vehicle for testing string attributes. First, they don't really support string attributes. Second, the DIMACS reader in igraph is very rudimentary and supports only a very specific format variant. I wanted to have better DIMACS support for a long time, but it's a lot of work (it's an ad-hoc format with countless variants) and no time/resources to do it ... https://github.com/igraph/igraph/issues/1924  I wouldn't bother with DIMACS tests in R unless you have a specific reason.

krlmlr:
Thanks. In an ideal world, every PR would come with tests that increase coverage. We're not quite there yet.

PR #1035, opened 2023-12-29
feat: Use `[]` assignment for converting matrices

krlmlr:


krlmlr:
@szhorvat: How would I go about constructing relevant tests here?

szhorvat:
What exactly are you looking to test? All of the affected functions?

szhorvat:
The C code expects a matrix. We don't have robust checks in place for converting from R to C yet (one of my big concerns about going ahead with 2.0 too soon). This change causes non-matrix input to be sent to C when a matrix is expected, and since the conversion code is not prepared for that, weird things will happen. 

Before this PR:

```
> sample_sbm(10, c(1,2,3,4), block.sizes=c(7,3))
Error in sample_sbm(10, c(1, 2, 3, 4), block.sizes = c(7, 3)) : 
  At vendor/cigraph/src/games/sbm.c:96 : Preference matrix is not square. Non-square matrix
```

After this PR:

```
> sample_sbm(10, c(1,2,3,4), block.sizes=c(7,3))
Error in sample_sbm(10, c(1, 2, 3, 4), block.sizes = c(7, 3)) : 
  INTEGER() can only be applied to a 'integer', not a 'NULL'
```

krlmlr:
I'm working on a PR for checking matrix inputs.

PR #1037, opened 2023-12-29
fix: Stimulus argument dependencies

krlmlr:
The implementations are currently unused, otherwise tests would be failing.

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -950,7 +950,7 @@ spanner_impl <- function(graph, stretch, weights) {
   # Function call
   res <- .Call(R_igraph_spanner, graph, stretch, weights)
   if (igraph_opt("return.vs.es")) {
-    res <- create_es(, res)
+    res <- create_es(graph, res)
   }
   res
 }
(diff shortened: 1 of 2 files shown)


szhorvat:
I don't understand what adding `V(graph)` does here, can you explain? I'm also getting to grips with stimulus, and don't have much prior experience with it.

CC @ntamas

ntamas:
I don't understand what `V(graph)` does, but the `DEPS` clause works like this. Sometimes, the input or output conversion code of an argument needs access to other arguments. For instance, if you have an argument named `vertex` that may contain a vertex name or a vertex index, you may need access to the graph to resolve the name to the vertex index (which igraph ultimately needs). In that case, you may specify `DEPS: vertex ON graph`, and this will make the graph object available in the input or output conversion as `%I1%` -- and, if you specify more dependencies, those are made available as `%I2%`, `%I3%` and so on. I think that Kirill's conversion code needs access to `V(graph)` for some reason so he specifies `DEPS: something ON graph V(graph)`, but I don't understand why it could not be written as `DEPS: something ON graph` followed by `V(%I1%)` in the conversion template.

krlmlr:
`V(graph)` is used, e.g., here: https://github.com/igraph/rigraph/pull/1037/files#diff-5a3f6a646811c63281fbabc9a1e68a5140a8fde989ab321c535379f9694d6a94R1003

Maybe it's wrong. We'll need tests, but I gave up on the coverage checks for now.

ntamas:
I just wonder whether it would be better to rewrite the `OUTCONV` clause of the `VERTEXINDEX` and `BIPARTITE_TYPES` types in `types-RR.yaml` like this:

`vertex_attr(%I1%, "name", V(%I1%))`

instead of what we have now:

`vertex_attr(%I1%, "name", %I2%)`

If the third argument always needs to be the full vertex sequence of the first argument, then it's easier because you only need to define `graph` in the `DEPS` clause instead of `graph V(graph)`.

PR #1038, opened 2023-12-29
chore: Use new function in favor of deprecated one

krlmlr:


szhorvat:
This is not a one-to-one replacement. The new function behaves differently. The old attributes should be added in R code now.

PR #1048, opened 2023-12-30
chore: Don't pass attributes to C functions

krlmlr:


szhorvat:
Do not generate an interface for `empty_attrs`. Use only `empty`.

PR #1052, opened 2023-12-30
refactor/type-refactorings

szhorvat:
These are multi-semi related commits. Please take a look.

Some of the "no visible binding for global variable" warnings are eliminated.

 @krlmlr: We need to implement conversion for the `LOOPS` type as well, but this is a bit special and I need help. Can you assist please? The mappings should be as follows:

 - `"ignore"`: `0`
 - `"twice"`: `1` (yes, that's right, "twice" maps to one)
 - `"once"`: `2`
 - `TRUE`: `1` (just like `"twice"`)
 - `FALSE`: `0` (just like `"ignore"`)

The logical values should be supported for backwards compatibility. I don't think these can be handled with `igraph.march.arg`.

### DIFF
--- a/tools/stimulus/types-RR.yaml
+++ b/tools/stimulus/types-RR.yaml
@@ -477,3 +477,8 @@ VORONOI_TIEBREAKER:
     DEFAULT:
         RANDOM: c("random", "first", "last")
     INCONV: '%I% <- switch(igraph.match.arg(%I%), "first"=0L, "last"=1L, "random"=2L)'
+
+GETADJACENCY:
+    DEFAULT:
+        BOTH: c("both", "upper", "lower")
+    INCONV: '%I% <- switch(igraph.match.arg(%I%), "upper"=0L, "lower"=1L, "both"=2L)'
--- a/tools/stimulus/types-RR.yaml
+++ b/tools/stimulus/types-RR.yaml
@@ -337,28 +337,6 @@ RECIP:
         Default: c('default', 'ratio')
     INCONV: '%I% <- switch(igraph.match.arg(%I%), ''default''=0, ''ratio''=1)'
 
-SCGMAT:
-    DEFAULT:
-        Default: c("symmetric", "laplacian", "stochastic")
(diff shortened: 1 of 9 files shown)


krlmlr:
```r
if (is.logical(loops)) {
  if (loops) {
    loops_out <- 1L
  } else {
    loops_out <- 0L
  }
} else {
  igraph.match.arg(...)
}
```

?

krlmlr:
If you need a closed-form expression, you can do:

```r
if (is.logical(loops)) {
  if (loops) {
    1L
  } else {
    0L
  }
} else {
  switch(igraph.match.arg(loops), "ignore" = 0L, ...)
}

File: tools/stimulus/types-RR.yaml [hunk 1]
```diff
@@ -338,30 +338,8 @@ HRG:
...
-SCGNORM:
-    DEFAULT:
-        Default: c("row", "col")
-    INCONV: '%I% <- switch(igraph.match.arg(%I%), "row"=1, "col"=2)'
-
-SCGDIR:
-    DEFAULT:
-        Default: c("default", "left", "right")
-    INCONV: '%I% <- switch(igraph.match.arg(%I%), "default"=1, "left"=2, "right"=3)'
+        DEFAULT: c('default', 'ratio')
```
krlmlr, line 341: Use double quotes consistently?

```suggestion
        DEFAULT: 'c("default", "ratio")'
```

krlmlr (approved):
Thanks for looking into it!

tools/stimulus/types-RR.yaml
szhorvat: Fixed.

szhorvat:
I'd like to have this merged as-is and continue with `loops` in a different PR

szhorvat:
I just saw you approved it, so I added the `mergequeue` label. Let me know if you don't want me to do this.

krlmlr:
Fine with me, thanks! I like small PRs better to.

Issue #1053, opened 2023-12-30
Do not duplicate prototypes between rinterface.c and rinterface_extra.c


szhorvat:
rinterface.c and rinterface_extra.c have a lot of function prototypes literally inserted.

TODO:

 - [ ] Move these to a common header, and include that header in both files
 - [ ] Remove unused functions
 - [ ] Remove unused prototypes
 - [ ] Change `int` return types to `igraph_error_t` where necessary. See https://github.com/igraph/rigraph/pull/1051#issuecomment-1872573532

The current setup is a crash risk and increases maintenance costs.

@krlmlr @Antonov548 can either of you do this please?

krlmlr:
Done.

PR #1055, opened 2023-12-30
refactor: do not duplicate prototypes, use a common header for rinterface.c / rinterface_extra.c

szhorvat:
Ref #1053


szhorvat:
@krlmlr Have a look before I continue.

krlmlr (approved):
Looks good if it passes checks. Thanks for taking this on.

szhorvat:
@krlmlr I added a second commit which:

 - fixes error return types
 - solidified error handling by adding more IGRAPH_CHECK

Good to merge?

There may still be issues with `PROTECT`, which I don't understand, but I don't think this PR introduced _new_ issues.

krlmlr (approved):

PR #1057, opened 2023-12-30
fix: local (weighted) transitivity now produces a named vector

### LINKED ISSUE #943: Get the name of the vertex in the igraph::transitivity(graph, type = "local") function in R
When I use the degree(graph) function from the igraph package in R, the output comes up with the name of the vertex and its respective value, however, when I run the transitivity(graph, type = "local") function from the same package, the output is a list without names, I can't tell which node each value corresponds to.

I suggest that the output of the transitivity(graph, type = "local") function to be accompanied by the name of the vertices.

szhorvat:
Fixes #943. This was also a learning exercise for me after your VERTEX_QTY question, @krlmlr.

### DIFF
--- a/tests/testthat/test-transitivity.R
+++ b/tests/testthat/test-transitivity.R
@@ -27,3 +27,21 @@ test_that("no integer overflow", {
   mtr <- min(transitivity(g, type = "local"), na.rm = TRUE)
   expect_true(mtr > 0)
 })
+
+# Check that transitivity() produces named vectors, see #943
+# The four tests below check four existing code paths
+test_that("local transitivity produces named vectors", {
+  g <- make_graph(~ a-b-c-a-d)
+  E(g)$weight <- 1:4
+  t1 <- transitivity(g, type = "local")
+  t2 <- transitivity(g, type = "barrat")
+
+  vs <- c("a", "c")
+  t3 <- transitivity(g, type = "local", vids = vs)
+  t4 <- transitivity(g, type = "barrat", vids = vs)
+
+  expect_equal(names(t1), V(g)$name)
+  expect_equal(names(t2), V(g)$name)
+  expect_equal(names(t3), vs)
+  expect_equal(names(t4), vs)
+})
--- a/R/structural.properties.R
+++ b/R/structural.properties.R
@@ -1351,13 +1351,21 @@ transitivity <- function(graph, type = c(
     .Call(R_igraph_transitivity_undirected, graph, isolates)
   } else if (type == 1) {
     if (is.null(vids)) {
-      .Call(R_igraph_transitivity_local_undirected_all, graph, isolates)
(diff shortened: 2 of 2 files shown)


krlmlr:
Thanks. What would it take to use autogenerated R code for `transitivity()`? Would that be more sustainable even in the short term?

krlmlr:
How do we know that this does what it's supposed to, without a test?

szhorvat:
> What would it take to use autogenerated R code for `transitivity()`?

This is a messy function that calls to one of several C functions. I'm not up for figuring out how to autogenerate this one right now.

But my personal preference would be to split this into three new functions, one for global clustering, one for local (weighted or unweighted), and one for average local (weighted or unweighted).

I'll add tests ...

krlmlr:
We don't need to autogenerate all of it, just replace the `.Call()` with R-level calls to our own `_impl()` functions (after we generate wrappers for them).

szhorvat:
Alright, let's try that.

szhorvat:
To be fair, I don't know if I'd have time to do this properly. It also depends on https://github.com/igraph/igraph/pull/2458 and there are issues to deal with like https://github.com/igraph/rigraph/issues/1058#issuecomment-1872614043

I would recommend merging this as-is since it does fix a complaint. How do I know it works? I tested all four affected code paths. I can add a quick test if necessary.

krlmlr:
A test would be great, thanks. We can merge as is, but should strive to automate as much as possible moving forward.

szhorvat:
Well of course, in the longer term this, as most other functions, _should_ be auto-generated.

tests/testthat/test-transitivity.R
```diff
@@ -27,3 +27,21 @@ test_that("no integer overflow", {
...
+  t2 <- transitivity(g, type = "barrat")
+
+  vs <- c("a", "c")
+  t3 <- transitivity(g, type = "local", vids = vs)
+  t4 <- transitivity(g, type = "barrat", vids = vs)
+
+  expect_equal(names(t1), V(g)$name)
+  expect_equal(names(t2), V(g)$name)
+  expect_equal(names(t3), vs)
+  expect_equal(names(t4), vs)
```
krlmlr, line 46: I have a slight preference for

```suggestion
  t1 <- transitivity(g, type = "local")
  expect_equal(names(t1), V(g)$name)
  
  t2 <- transitivity(g, type = "barrat")
  expect_equal(names(t2), V(g)$name)

  vs <- c("a", "c")
  t3 <- transitivity(g, type = "local", vids = vs)
  expect_equal(names(t3), vs)
  
  t4 <- transitivity(g, type = "barrat", vids = vs)
  expect_equal(names(t4), vs)
```

because this keeps input and output closer together. No need to act here.

krlmlr (approved):
Thanks, looks good!

PR #1061, opened 2023-12-31
fix: Restore `NA` padding for `dfs(unreachable = FALSE)$order` and `$order.out` .

krlmlr:
For https://github.com/natverse/nat/issues/517.


krlmlr:
This is actually moot because revdepchecks don't change. Changing `bfs()` behavior instead.

PR #1062, opened 2023-12-31
feat!: Remove `NA` padding for `dfs(unreachable = FALSE)$order` and `$order.out` and for `bfs(unreachable = FALSE)$order` .

krlmlr:


krlmlr:
Running checks for this PR combined with #1051.

PR #1064, opened 2024-01-01
chore: Standardize the handling of some stimulus enum types

szhorvat:
Makes sure that all enum types are handled as integers  on the R side. This paves the way for eventually auto-generating everything.


krlmlr:
Thanks!

PR #1066, opened 2024-01-02
Check that matrix sizes are in the supported range before passing them to R

### LINKED ISSUE #889: Need to guard against too-large matrices
R does not seem to support matrix dimensions that do not fit into a 32-bit integer, while C/igraph does.

```
> m<-matrix(ncol=.Machine$integer.max + 1,nrow=1)
Error in matrix(ncol = .Machine$integer.max + 1, nrow = 1) : 
  invalid 'ncol' value (too large or NA)
In addition: Warning message:
In matrix(ncol = .Machine$integer.max + 1, nrow = 1) :
  NAs introduced by coercion to integer range
```

Also, [here](https://stat.ethz.ch/R-manual/R-devel/library/base/html/LongVectors.html):

> Arrays (including matrices) can be based on long vectors provided each of their dimensions is at most $2^{31}-1$ : thus there are no 1-dimensional long arrays. 

When transferring matrices from igraph to R, it is important to check that the dimensions are not too large.

This affects functions like `R_igraph_matrix_to_SEXP()`.

szhorvat:
Fixes #889

This fix is theoretical at the moment. I can't seem to test it on macOS. The test would use a very large amount of memory.

```r
> g<-make_empty_graph(.Machine$integer.max + 2)
> distances(g, v=1)
Error: vector memory exhausted (limit reached?)
```

@krlmlr @Antonov548 Can either of you try to test this on a large-memory Linux machine? I don't have access to one at this moment.  Or otherwise do we merge as-is without a through test?

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -2591,6 +2591,13 @@ SEXP R_igraph_matrix_to_SEXP(const igraph_matrix_t *m) {
   return result;
 }
 
+SEXP R_igraph_0ormatrix_to_SEXP(const igraph_matrix_t *m) {
+  if (!m) {
+    return R_NilValue;
+  }
+  return R_igraph_matrix_to_SEXP(m);
+}
+
 SEXP R_igraph_matrix_int_to_SEXP(const igraph_matrix_int_t *m) {
   SEXP result, dim;
   igraph_integer_t n = igraph_matrix_int_size(m);
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -2576,15 +2576,24 @@ SEXP R_igraph_0orvector_complex_to_SEXP(const igraph_vector_complex_t *v) {
 }
 
 SEXP R_igraph_matrix_to_SEXP(const igraph_matrix_t *m) {
-
   SEXP result, dim;
+  const igraph_integer_t nrow = igraph_matrix_nrow(m);
+  const igraph_integer_t ncol = igraph_matrix_ncol(m);
+
+  /* Assuming that this function is called in a context where
+   * igraph_error() does not return. */
+  if (nrow > INT_MAX || ncol > INT_MAX) {
+    igraph_errorf("igraph returned a matrix of size %" IGRAPH_PRId " by %" IGRAPH_PRId ". "
+                  "R does not support matrices with more than %d rows or columns.",
+                  __FILE__, __LINE__, IGRAPH_FAILURE,
+                  nrow, ncol, INT_MAX);
+  }
 
   PROTECT(result=NEW_NUMERIC(igraph_matrix_size(m)));
   igraph_matrix_copy_to(m, REAL(result));
   PROTECT(dim=NEW_INTEGER(2));
-  /* TODO check that row and column counts fit in an int */
-  INTEGER(dim)[0]=(int) igraph_matrix_nrow(m);
-  INTEGER(dim)[1]=(int) igraph_matrix_ncol(m);
+  INTEGER(dim)[0] = (int) nrow;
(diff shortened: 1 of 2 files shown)


krlmlr:
Instead of a one-off test, can we instead implement a function that checks matrix dimensions, test the behavior of that function (in an automated test), and use it where needed?

If we don't want to expose this function to R, we can use the catch testing framework in testthat: https://testthat.r-lib.org/reference/use_catch.html .

szhorvat:
Each matrix type needs a different implementation, so this can't be reduced to calling a single function three times

krlmlr:
Just the part where `ncol` and `nrow` is checked, perhaps?

szhorvat:
> test the behavior of that function (in an automated test),

How do you intend to test a C function from R? R doesn't even support the integer type that the function would be operating with. What would be the point of testing a function that just compares two integers and reports an error? I am not sure how to proceed here.

I managed to test manually that the functionality works after setting `R_MAX_VSIZE=100Gb`, using the command line instead of RStudio, and using a barebones function like `layout_in_circle()`. The R process used over 120 GB of memory.


```
> g<-make_empty_graph(n=.Machine$integer.max+1)
> l<-layout_in_circle(g)
Error in layout_in_circle(g) : 
  At rinterface_extra.c:2588 : igraph returned a matrix of size 2147483648 by 2. R does not support matrices with more than 2147483647 rows or columns. Failed
```

src/rinterface_extra.c
```diff
@@ -2576,33 +2576,59 @@ SEXP R_igraph_0orvector_complex_to_SEXP(const igraph_vector_complex_t *v) {
...
+  const igraph_integer_t ncol = igraph_matrix_ncol(m);
+
+  /* Assuming that this function is called in a context where
+   * igraph_error() does not return. */
+  if (nrow > INT_MAX || ncol > INT_MAX) {
+    igraph_errorf("igraph returned a matrix of size %" IGRAPH_PRId " by %" IGRAPH_PRId ". "
+                  "R does not support matrices with more than %d rows or columns.",
+                  __FILE__, __LINE__, IGRAPH_FAILURE,
+                  nrow, ncol, INT_MAX);
+  }
```
krlmlr, line 2590: This could be extracted into a function, say, `R_igraph_check_matrix_dims(igraph_integer_t nrow, igraph_integer_t ncol)` . This function becomes testable, e.g., through an R wrapper that we export just for testing. (Overkill, perhaps.)

We could also go one step further and make this function sensitive to a global switch (one that's perhaps only active in debug mode). The global switch would change the definition of `INT_MAX` for this function so that we can even test that (some) functions that return a matrix actually go through this code path. (Overkill, perhaps.)

But these two steps combined allow creating a version of igraph that doesn't require allocating hundreds of gigabytes to check the functionality.

No action needed here, but for the future, should we strive to take testability into account when implementing?

krlmlr:
Thanks!

src/rinterface_extra.c
szhorvat: > should we strive to take testability into account when implementing

This PR fixes a bug that manifests itself in difficult to test scenarios. Nevertheless it is a bug that should be fixed. I did verify the fix manually.

> This function becomes testable, e.g., through an R wrapper that we export just for testing.

The function would take arguments of type `igraph_integer_t`. This type is not supported by R. Using `double` makes things complicated and increases the chance of mistakes (not ever `igraph_integer_t` is representable as a `double`).

So I'm not sure how to do this in a productive way.

krlmlr: It's in the merge queue, thanks for working on this.

Happy to elaborate if this situation occurs on an issue from https://github.com/igraph/rigraph/milestone/14.

PR #1067, opened 2024-01-02
refactor!: remove graph.isomorphic.34(), fixes #1065

### LINKED ISSUE #1065: graph.isomorphic.34() is non-functional
`graph.isomorphic.34()` is currently non-functional, as it refers to `isomorphic_34_impl` which does not exist.

`graph.isomorphic.34()` should probably just be removed as it is a special high-performance path for small graphs. It will be used automatically by `isomorphic` when appropriate.

If we don't want to remove it, then the replacement of `igraph_isomorphic_34()`, called `igraph_isomorphic_small()`, should be exposed. I do not recommend this as this function has certain expectations about its input (such as graphs being simple) which the R interface does not currently guarantee.

szhorvat:
Fixes #1065.

A GitHub search does not reveal any uses of `graph.isomorphic.34()` in other packages. This is breaking, so reverse dependency checks _should_ be run, but I would be surprised if any issue came up.

My recommendation is to merge this if revdepchecks pass (i.e. remove this currently broken function) and re-add it later if anyone complains. What this function provides is better performance for small simple graphs (but without checking that the graphs are simple). It's unclear if this performance gain is significant in R, given the overhead of calling a C function.


krlmlr (approved):
Thanks. Do we also want to remove `isomorphic_34_impl` ?

szhorvat:
It didn't exist in the first place. That's why the function was broken

PR #1044, opened 2023-12-29
Script to sync C and R yaml

krlmlr:

PR #1068, opened 2024-01-02
refactor: make_empty_graph() is now fully auto-generated

szhorvat:
Related to #1140

`make_empty_graph()` is now fully auto-generated. The custom implementation is no longer needed after #1051.

This PR conflicts with the first commit from the draft PR #1044.


krlmlr:
Why is this urgent?

szhorvat:
If you don't want to merge it now your can put it on hold. This was part of verifying that scalar conversion checks are working as intended and there are no crashes

szhorvat:
I should mention that this also catches errors that the previous version did not (passing a longer vector where a scalar is expected). I don't see any disadvantage to merging this, quite the contrary.

szhorvat:
Is there any reason to delay merging this? Instead of saying it conflicts with the first commit in #1044, I should have said that it _replaces_ that commit. This is not a breaking change, just simplifies codes and uses standard error reporting instead of a custom version.

krlmlr:
We can merge right after the 2.0.0 release, hopefully next week, unless the revdepchecks show anything major.

krlmlr:
Do we never intend to expose or even use `make_empty_attrs()` ? Should we add a comment to explain the reasoning, like you did for other low-level functions earlier?

I still need to run revdepchecks on this.

szhorvat:
> Do we never intend to expose or even use `make_empty_attrs()` ?

No



> Should we add a comment to explain the reasoning, like you did for other low-level functions earlier?

Done



> I still need to run revdepchecks on this.

Risk of breakage is very low, but always a good practice to run revdepchecks

tools/stimulus/functions-R.yaml
```diff
@@ -11,7 +11,10 @@
 #######################################
 
 igraph_empty:
-    IGNORE: RR
+
+# Not needed in R, we use igraph_emtpy() instead
+igraph_empty_attrs:
+    IGNORE: RR, RC
```
krlmlr, line 17: Thanks. Why has the order changed now? I remember aligning this file with `functions.yaml` from the C core, I'd like both to be in the same order. No need to be religious about it, but let's keep the drift here to a minimum moving forward.

krlmlr:
Thanks!

tools/stimulus/functions-R.yaml
szhorvat: I changed the order to be able to see both related functions next to it. If you're bothered by it, I'll change it in the C core too.

krlmlr: Thanks. Changing in the C core would be great, if it's idiomatic there.

szhorvat: Done.

PR #1069, opened 2024-01-02
fix: add scalar conversion checks in a few critical places

szhorvat:
This PR adds checks for converting scalars (especially scalar integers) to C in a few heavily used and critical places. This is similar to #1051 but for manually written code. I will not add the checks for all manually written code, only for the most important parts. The rest should be handled by moving functions to auto-generation.

I needed to change a test which verified that `make_graph(NULL, n = NULL)` worked. IMO using `n=NULL` should not be allowed an in fact this probably triggered an invalid memory access, as the code would index into a zero-length vector (which is somehow not caught by the sanitizer). I am not sure why this test was present.

Because of the presence of this unusual test, running revdepchecks would be a good idea.

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -3865,7 +3865,8 @@ SEXP R_igraph_add_vertices(SEXP graph, SEXP pnv) {
   igraph_t g;
   SEXP result;
 
-  nv=(igraph_integer_t) REAL(pnv)[0];
+  R_check_int_scalar(pnv);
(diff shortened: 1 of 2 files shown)


szhorvat:
The test was added in https://github.com/igraph/rigraph/commit/6453014fa874cd5203f25b6be24aa18e1a23e6b3

#1068 effectively reverts the rest of that commit.

Antonov548 (approved):

krlmlr:
Thanks!

Issue #907, opened 2023-10-21
Non-numerical adjacency matrices


szhorvat:
At the moment, `as_adjacency_matrix` is _mostly_ implemented in R. In 0.10, we have _almost_ everything in place to rely on this C core for this. This is important because there are subtleties with counting self-loops in undirected graphs, which should work the same way across all interfaces of igraph, and more importantly: across all implementations of the adjacency matrix conversion functions. The code paths that use the C core should produce the same kinds of results as the code paths that use pure-R implementations.

There is one currently piece of functionality that is not implemented in the C core: creating adjacency matrices that are non-numerical.

Here's a small demo of this functionality with the current `main` branch:

```
> library(igraph)
> g<-make_graph(c(1,2, 2,3), directed = F)
> E(g)$foo <- c('bar', 'baz')
> as_adjacency_matrix(g, sparse=F)
     [,1] [,2] [,3]
[1,]    0    1    0
[2,]    1    0    1
[3,]    0    1    0
> as_adjacency_matrix(g, sparse=F, attr='foo')
     [,1]  [,2]  [,3] 
[1,] ""    "bar" ""   
[2,] "bar" ""    "baz"
[3,] ""    "baz" ""   
> as_adjacency_matrix(g, sparse=T, attr='foo')
Error in get.adjacency.sparse(graph, type = type, attr = attr, edges = edges,  : 
  Sparse matrices must be either numeric or logical,and the edge attribute is not
> E(g)$foo <- c(F, T)
> as_adjacency_matrix(g, sparse=T, attr='foo')
3 x 3 sparse Matrix of class "lgCMatrix"
          
[1,] . : .
[2,] : . |
[3,] . | .
> as_adjacency_matrix(g, sparse=F, attr='foo')
      [,1]  [,2]  [,3]
[1,] FALSE FALSE FALSE
[2,] FALSE FALSE  TRUE
[3,] FALSE  TRUE FALSE
```

As you can see, the dense adj mat function supports multiple different non-numeric types, while the sparse version supports numeric and logical.

----

@krlmlr We need to make a decision about how to handle this during the 0.10 upgrade. Please comment here so we have the decision in writing.

My suggestion:

 - Drop this functionality for now and support only numeric types, nothing else. This way the adj mat conversion implementation can fully rely on the C core, ensuring that it behaves consistently.
[...]

PR #1072, opened 2024-01-02
feat!: `as_adjacency_matrix()` no longer supports attributes of type `character`

krlmlr:
For #907.

I see how it is useful to advance on that front. Running revdepchecks, if they succeed, I'll merge. Anything beyond that will have to wait until after 2.0.0.


szhorvat (approved):

krlmlr:
Two packages have new breakages, I believe SCORPIUS is a false positive, but checking again.

There are two paths from here:

- We stop support for this functionality, and ask downstream to change their implementation
- We add a test for this functionality and keep supporting it

No urgency from my end.

szhorvat:
It makes sense to stop support, as this can't be implemented in the C core, and its usefulness is dubious. Eventually all adjacency matrix stuff should rely on the C implementation.

That said, it would be fair to contact these packages and see _why_ they are using this functionality, and whether it is easy for them to stop relying on it. I don't see any good reason for working with string-based adjacency matrices, but maybe I am missing something.

krlmlr:
@sarahleavitt @schochastics @robertjankowski: The nbTransmission and signnet packages are using `as_adjacency_matrix()` with an edge attribute of type character. Is this on purpose, or perhaps a mistake? Can you work around so that your packages continue working with this PR?

Install with `pak::pak("igraph/rigraph@f-907-adj-chr")` .

schochastics:
Just checked my usecase and I can work around this without issues. I will fix it in the next few days

schochastics:
@krlmlr already fixed it. I will submit signnet 1.0.4 when CRAN submissions are online again

szhorvat:
@schochastics I am interested in what your use case was an how you worked around it. Can you share it? Are you working with complex-values adjacency matrices, or complex-values attributes? Is this something that makes sense to support in C/igraph in the future? I'm primarily interested in the theory behind this, not the implementation.

schochastics:
@szhorvat Sure! The technical details are in [this paper](https://doi.org/10.1080/0022250X.2019.1711376). 
When you project signed two mode networks, a third type of tie can occur (called the "ambivalent tie"). To support it in signnet, I use an edge attribute `type` with values "N", "P", and "A" and those get transformed to a complex adjacency matrix with P=1+0i, N=0+1i and A=0.5+0.5i. It was convenient to build it with `as_adj(g, attr="type")` and then convert "P" "N" and "A" entries to complex values, but it was trivial to work around this.

I would like to keep the "N", "P", and "A" nomenclature, because it is easier to understand for users and it is enough for most of the functionality of signnet. The complex matrix part is still very experimental but I think there are some other papers out there that work with complex adjacency matrices (in different context), so maybe it is something to support in future versions.

sarahleavitt:
I just checked on this for my package and it is not an issue for me. I simply carried over a large dataframe as the dataframe for creating the edge list. I have removed the irrelevant columns and resubmitted my package with the changes. So, my use-case does not require a character edge attribute.

Thank you,
Sarah
________________________________

PR #1079, opened 2024-01-08
feat: Use 30 bits of R's RNG

### LINKED ISSUE #782: Finalize update of RNG to 0.10
Note that R's RNG interface produces floating point values, not integers, which hides the actual "amount of randomness" it is capable of creating.

 - [ ] Check how many bits of random values R is actually capable of creating, and set up the `get` RNG method accordingly. If it's possible to use 32 bits instead of 31, that should improve the performance of integer generation. Note that R has multiple RNG back-ends and they may not be all of the same quality.
 - [ ] Add R's own gamma generator
 - [ ] Add R's own Poisson generator

Ref: https://github.com/wch/r-source/blob/trunk/src/main/RNG.c

Note that in R's own code, they only seem to trust `unif_rand()` up to 16 bits. See the `rbits()` internal function in `RNG.c`.

krlmlr:
Closes #782.

### DIFF
--- a/src/rrandom.c
+++ b/src/rrandom.c
@@ -58,7 +58,8 @@ static igraph_error_t igraph_rng_R_seed(void *state, igraph_uint_t seed) {
 }
 
 static igraph_uint_t igraph_rng_R_get(void *state) {
-    return (unif_rand() * 0x7FFFFFFFUL);
+    // unif_rand() returns a double in [0, 1)
+    return (unif_rand() * 0x40000000UL);
 }
 
 static igraph_real_t igraph_rng_R_get_real(void *state) {
@@ -99,7 +100,7 @@ static igraph_real_t igraph_rng_R_get_pois(void *state, igraph_real_t rate) {
 
 static igraph_rng_type_t igraph_rng_R_type = {
     /* name= */      "GNU R",
-    /* bits = */     31,
+    /* bits = */     30, // tested by test-rng.R, #782
     /* init= */      igraph_rng_R_init,
     /* destroy= */   igraph_rng_R_destroy,
     /* seed= */      igraph_rng_R_seed,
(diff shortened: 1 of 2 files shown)


szhorvat:
I am confused about why this is not causing mass test failures. This changes the behaviour of the RNG. For the same seed, it should return a different sequence of integers.

I just checked in `rrandom.c` that the only sampler that we don't rely directly on R for is `get_int`. 

https://github.com/igraph/rigraph/blob/b806ec209943846380015cb5ba3a7e605799d93c/src/rrandom.c#L101-L116

Nevertheless, `get_int` is used by many functions, e.g. `barabasi_game()`. I just checked that `sample_pa(10)` continues to return the same result before and after this PR.

Is something wrong? Do you understand why nothing changed, @ntamas ?

All this said, I strongly recommend running revdepchecks again after this PR, @krlmlr The output of some randomized functions _should_ change (and that's fine, but may break some other packages' tests)

szhorvat:
`sample_pa()` was the wrong function to test as it only uses `RNG_INTEGER()` in specific situations. But `sample_tree()` always uses it, yet it still continues to return the same result. That does not look right.

krlmlr:
We did not change `igraph_rng_R_get_real()` .

krlmlr:
Confirming the same results for yesterday's version and the current main branch.

``` r
set.seed(20240108)
igraph::sample_tree(20)
#> IGRAPH 596b969 U--- 20 19 -- 
#> + edges from 596b969:
#>  [1] 14--16 13--16  5--13  8--13  8--10 10--11  5-- 6  6--15  9--15  9--12
#> [11] 12--18  7--16 12--17  4--17  2--11  2--20  3--18  1--16 15--19
```

<sup>Created on 2024-01-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

It looks legit, though. In `sample_tree()` we draw random numbers for very small integer ranges. These are derived from the most significant bits in the number returned by the integer RNG. The MSBs stay the same (and are very very unlikely to change) for the same initial seed, though, because we get them by multiplying a real in [0; 1) with 2^31 - 1 (old, slightly wrong) or 2^30 (new).

krlmlr:
Anyway, running revdepchecks is a good idea here.

szhorvat:
Yes, I also looked in more detail and I think I see why this is happening.

Nevertheless, there should be a note in the changelog that randomized functions may not return the same result for the same seed.

Issue #1082, opened 2024-01-08
Improve top-level code in tests


maelle:
For instance at the top of test-indexing there are two helper functions with names so short I find them unclear. 

For top level code
- rename helper functions, save them in tests/testthat/helper- files
- move it to inside the tests.

maelle:
@schochastics 

- `test-operators.R` is a false positive of my script.
- `tests/testthat/test-embedding.R` defines its own functions, seems ok?
- same for `tests/testthat/test-glet.R`
- `tests/testthat/test-other.R` has top-level code towards the end, but the variables are used in two tests only. Is that fine so?

schochastics:
To me this looks all fine. What about `test-scan.R`? Do you think this should be improved? If so I could do it in https://github.com/igraph/rigraph/pull/1982

maelle:
Oh right I forgot that one. If you can improve it, that would be great and we'd have one issue fewer. Thank you!!

PR #1080, opened 2024-01-08
chore!: Hard-deprecate `as_adjacency_matrix(edges = )`

krlmlr:
For #815, more work needed. Running revdepchecks now.

PR #1085, opened 2024-01-08
test: remove top-level code in test-indexing.R

maelle:
@szhorvat if `am()` and `mm()` had longer names, what would those be? I'd like to replace them with longer, more informative names.

Removing top-level code in test files help make each test more self-contained, therefore easier to re-run interactively when it fails (as happens in #1080)

Part of #1082

### DIFF
--- a/tests/testthat/helper-indexing.R
+++ b/tests/testthat/helper-indexing.R
@@ -1,9 +1,9 @@
-mm <- function(...) {
+vector_to_square_matrix <- function(...) {
   v <- as.numeric(as.vector(list(...)))
   matrix(v, nrow = sqrt(length(v)))
 }
 
-am <- function(x) {
+canonicalize_matrix <- function(x) {
   x <- as.matrix(x)
   dimnames(x) <- NULL
   x
--- a/tests/testthat/helper-indexing.R
+++ b/tests/testthat/helper-indexing.R
@@ -1,0 +1,24 @@
+mm <- function(...) {
+  v <- as.numeric(as.vector(list(...)))
+  matrix(v, nrow = sqrt(length(v)))
+}
+
+am <- function(x) {
+  x <- as.matrix(x)
+  dimnames(x) <- NULL
+  x
+}
+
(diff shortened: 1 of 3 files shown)


maelle:
@szhorvat based on your input I'll rename the two functions (if you agree!) then once it's merged I can rebase #1080 and remove the test that's failing there (it's failing because it's using edges, which we now expect to not work).

szhorvat:
I'm not sure what these names are referring to.

tests/testthat/helper-indexing.R
```diff
@@ -0,0 +1,24 @@
+mm <- function(...) {
+  v <- as.numeric(as.vector(list(...)))
+  matrix(v, nrow = sqrt(length(v)))
+}
+
+am <- function(x) {
```
maelle, line 6: am() is a helper function in tests. maybe it could be called as_matrix? @szhorvat

tests/testthat/helper-indexing.R
```diff
@@ -0,0 +1,24 @@
+mm <- function(...) {
```
maelle, line 1: this is mm(). It takes a vector and makes into a matrix. vector_to_matrix?

tests/testthat/helper-indexing.R
szhorvat: How about `canonicalize_matrix`? (Since you asked. But use anything you like.)

tests/testthat/helper-indexing.R
szhorvat: How about `vector_to_square_matrix`?

tests/testthat/helper-indexing.R
maelle: Thank you!

tests/testthat/helper-indexing.R
maelle: Thanks!

szhorvat (approved):
Just to dismiss the yellow "review requested" bar.

Issue #1006, opened 2023-12-18
Warning from R CMD check


maelle:
```r
W  checking whether package ‘igraph’ can be installed (2m 7.3s)
   Found the following significant warnings:
     rinterface_extra.c:3149:3: warning: ‘igraph_incidence’ is deprecated [-Wdeprecated-declarations]
     rinterface_extra.c:3261:34: warning: passing argument 5 of ‘igraph_sparsemat_view’ from incompatible pointer type [-Wincompatible-pointer-types]
     rinterface_extra.c:3261:53: warning: passing argument 6 of ‘igraph_sparsemat_view’ from incompatible pointer type [-Wincompatible-pointer-types]
     rinterface_extra.c:4416:3: warning: ‘igraph_lattice’ is deprecated [-Wdeprecated-declarations]
     rinterface_extra.c:5004:3: warning: ‘igraph_erdos_renyi_game’ is deprecated [-Wdeprecated-declarations]
     rinterface_extra.c:6479:3: warning: ‘igraph_read_graph_dimacs’ is deprecated [-Wdeprecated-declarations]
     rinterface_extra.c:6537:3: warning: ‘igraph_write_graph_dimacs’ is deprecated [-Wdeprecated-declarations]
     rinterface_extra.c:8327:3: warning: ‘igraph_laplacian’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:3853:3: warning: ‘igraph_hub_score’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:3902:3: warning: ‘igraph_authority_score’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:5300:3: warning: ‘igraph_random_edge_walk’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:5927:3: warning: ‘igraph_bipartite_game’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:9970:3: warning: ‘igraph_subisomorphic_function_vf2’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:11829:3: warning: ‘igraph_get_stochastic_sparsemat’ is deprecated [-Wdeprecated-declarations]
     rinterface.c:11853:3: warning: ‘igraph_hrg_dendrogram’ is deprecated [-Wdeprecated-declarations]
     vendor/cigraph/src/core/matrix.c:234:13: warning: ‘igraph_vector_e_tol’ is deprecated [-Wdeprecated-declarations]
[...]

PR #1087, opened 2024-01-08
chore: Reduce warnings from `R CMD check`

krlmlr:
For #1006.


szhorvat:
It seems that it's including `src/vendor/igraph_export.h`. That's the file you need to modify.

I'm not sure why `igraph_export.h` is in `tools/stimulus`, but it doesn't seem to be used anymore. @Antonov548, do you know why that file is there?

Antonov548:
> It seems that it's including `src/vendor/igraph_export.h`. That's the file you need to modify.
> 
> I'm not sure why `igraph_export.h` is in `tools/stimulus`, but it doesn't seem to be used anymore. @Antonov548, do you know why that file is there?

I think it's there because file from `tools/stimulus` is replacing the file in `src/vendor` after call `Makefile-cigraph` which basically makefile for vendoring igraph

szhorvat:
> file from `tools/stimulus` is replacing the file in `src/vendor` after call `Makefile-cigraph`

That's not actually happening. After running `make -f Makefile-cigraph` I still see `#define IGRAPH_DEPRECATED __attribute__ ((__deprecated__))` in `src/vendor/igraph_export.h`.

Antonov548:
> > file from `tools/stimulus` is replacing the file in `src/vendor` after call `Makefile-cigraph`
> 
> That's not actually happening. After running `make -f Makefile-cigraph` I still see `#define IGRAPH_DEPRECATED __attribute__ ((__deprecated__))` in `src/vendor/igraph_export.h`.

Yes, you are right. I have created pull reuqest: https://github.com/igraph/rigraph/pull/1089

krlmlr:
Thanks, good catch!

Issue #1040, opened 2023-12-29
Fix conversion of sparse matrices


krlmlr:
We need to copy instead of creating a view.

https://github.com/igraph/rigraph/actions/runs/7359662568/job/20034815374?pr=1012#step:9:83

szhorvat:
This is a prerequisite: #1063

szhorvat:
We don't have a copying version of `igraph_sparsemat_view()` in the C core. I will add a private one so we can make progress on this. It can be made public in version 1.0 of the C core.

krlmlr:
Deferring, moving to another milestone.

krlmlr:
#1090 has code and conflicts.

schochastics:
What exactly is the issue here?

krlmlr:
Let's review this when we have bumped to igraph/C 1.0.1 .

PR #1090, opened 2024-01-09
fix: incompatible pointers due conversion in `R_SEXP_to_sparsemat`

Antonov548:
https://github.com/igraph/rigraph/issues/1006


src/rinterface_extra.c
```diff
@@ -3172,9 +3172,13 @@ igraph_error_t R_SEXP_to_sparsemat(SEXP pakl, igraph_sparsemat_t *akl) {
   SEXP p=GET_SLOT(pakl, Rf_install("p"));
   SEXP x=GET_SLOT(pakl, Rf_install("x"));
 
+  igraph_vector_int_t ci, cp;
+  IGRAPH_CHECK(R_SEXP_to_vector_int_copy(i, &ci));
+  IGRAPH_CHECK(R_SEXP_to_vector_int_copy(p, &cp));
```
szhorvat, line 3177: This memory will need to be released.

Antonov548: Yes, it should. I have created draft to discuss it. 
Is it sounds okay to make a copy for all vectors in `cs_igraph_sparse` and use `igraph_sparsemat_destroy` after conversion?

Antonov548: Anyway we need to copy 2 of vectors `i` and `p` due to the `number()` to `igraph_integer_t` conversion. Maybe it will just good enough to make `R_SEXP_to_sparsemat_copy`?

krlmlr:
Thanks. Can we shelve this for now and just remove the offending functions and all functions that call it? Would this break anything?

krlmlr:
Don't need to change Stimulus definitions, just remove code, even if it's generated.

szhorvat:
> Can we shelve this for now and just remove the offending functions and all functions that call it? Would this break anything?

That would be the best, since doing this right requires some attention. It's better not to rush it.

tools/stimulus/functions-R.yaml
```diff
@@ -83,8 +83,10 @@ igraph_adjacency:
     IGNORE: RR
 
 igraph_sparse_adjacency:
+    IGNORE: RC, RR
```
krlmlr, line 86: Let's keep Stimulus unchanged, we want the autogenerated code to come back after the CRAN release.

R/aaa-auto.R
```diff
@@ -3989,6 +3938,18 @@ stochastic_imitation_impl <- function(graph, vid, algo, quantities, strategies,
   res
 }
 
+has_attribute_table_impl <- function() {
```
krlmlr, line 3941: Similarly, this and the function below previously were removed on purpose.

krlmlr (commented):
Thanks, great!

krlmlr:
We can also open a new PR with a squashed commit to keep the changes here for the next iteration.

tools/stimulus/functions-R.yaml
szhorvat: OK, I only saw this comment after I suggested the opposite in the other thread.

krlmlr:
Too many conflicts, let's revisit when handling sparse matrices.

krlmlr:
Closing, #1040 has a reference here. Conflicts, need to redo/resolve when ready.

PR #1092, opened 2024-01-10
Fixes for Makevars.in for unix (revert accidental file updates) 

jeroen:
I think you may have accidentally copied a wrong version of [src/Makevars.in](https://github.com/igraph/rigraph/blob/main/src/Makevars.in) in #840. What you have now is a broken version of a Windows Makevars, not the unix one.

This PR restores it to the correct version from before #840 plus a small improvement to use the new `sources.mk`.

### DIFF
--- a/src/Makevars.in
+++ b/src/Makevars.in
@@ -1,15 +1,15 @@
 include sources.mk
 
-LIB_XML ?= $(MINGW_PREFIX)
-GLPK_HOME ?= $(MINGW_PREFIX)
-LIB_GMP ?= $(MINGW_PREFIX)
+PKG_CFLAGS=$(C_VISIBILITY)
+PKG_CXXFLAGS=$(CXX_VISIBILITY)
+PKG_FFLAGS=$(F_VISIBILITY)
 
 PKG_CPPFLAGS=@cflags@ -DUSING_R -I. -Ivendor -Ivendor/cigraph/src -Ivendor/cigraph/include -Ivendor/cigraph/vendor \
 	-DNDEBUG -DNTIMER -DNPRINT -DINTERNAL_ARPACK -DIGRAPH_THREAD_LOCAL= \
 	-DPRPACK_IGRAPH_SUPPORT \
 	-D_GNU_SOURCE=1
 
-PKG_LIBS = -L"${LIB_XML}/lib" -lxml2 -lz -lstdc++ -L"${GLPK_HOME}/lib" \
-  -lglpk -lgmp -L"${LIB_GMP}/lib" "${BLAS_LIBS}" "${LAPACK_LIBS}"
+PKG_LIBS = -L"${LIB_XML}/lib" -lxml2 -lz -L"${GLPK_HOME}/lib" \
+  -lglpk -lgmp -L"${LIB_GMP}/lib" $(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS)
 
 OBJECTS=${SOURCES}


krlmlr:
We're moving away from autoconf.

What exactly would you like to see changed? The mainline works well enough for us on GHA.

jeroen:
Your unix Makevars.in uses hardcoded variable from an old version of winbuilder, this unlikely to work on most platforms.

None of those variables actually exist. Also you are hardcoding `-lstdc++` which breaks linux clang builds. Also CMD check warns you that the way you link to lapack is wrong:

```
Warning:   apparently using $(LAPACK_LIBS) without following $(BLAS_LIBS) in ‘src/Makevars’
  apparently using $(BLAS_LIBS) without following $(FLIBS) in ‘src/Makevars’
  apparently using $(LAPACK_LIBS) without following $(BLAS_LIBS) in ‘src/Makevars.in’
  apparently using $(BLAS_LIBS) without following $(FLIBS) in ‘src/Makevars.in’
```

You should be using `$(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS)` as you did before.

Either way if this is intentional I'll leave it up to you to fix it, thought maybe it was a mistake.

szhorvat:
It's hard to make do with plain makefiles without using all the auto-detection magic of either autoconf or CMake. Kirill, I know why you prefer simplicity (we all do when feasible), but the same level of portability that igraph's C core has simply isn't achievable without an advanced build system generator such as CMake or autoconf. If we don't do that in the long term, we will only be able to support simple and predictable environment, e.g. target CRAN only.

As for `libstdc++`, the only reliable way is to not link it explicitly, but find a way to use the C++ compiler to perform the linking (e.g. link with `g++` instead of `gcc`). This is what CMake does as well.

jeroen:
Instead of autoconf you can manually write a simple configure script that calls`pkg-config` to populate `Makevars.in` like many other packages do, something like this: https://github.com/r-lib/ragg/blob/main/configure

> As for libstdc++, the only reliable way is to not link it explicitly, but find a way to use the C++ compiler to perform 

See [section 5.6.1](https://cran.r-project.org/doc/manuals/R-exts.html#External-C_002b_002b-code) from WRE: 

> Even fewer external libraries use C++ internally but present a C interface, such as [rgeos](https://cran.r-project.org/package=rgeos). These require the C++ runtime library to be linked into the package’s shared object/DLL, and this is best done by including a dummy C++ file in the package sources.

So just create some file `src/dummy.cpp` and then R will automatically use the C++ linker.

krlmlr:
Thank you for your input. I'll leave this open for tracking.

jeroen:
OK I have updated this PR to fix at least the urgent issues.

krlmlr:
Thanks, `R CMD check` looks much better now.

jeroen:
With this merged, at least the cross binaries are working again. So you can install macos-arm64 binaries from https://igraph.r-universe.dev/igraph

szhorvat:
> Instead of autoconf you can manually write a simple configure script that calls`pkg-config` to populate `Makevars.in` like many other packages do

Many packages don't provide pkg-config files. GLPK does not and GMP only does with certain versions on certain systems. If GMP is actually MPIR (a compatible fork) then that's yet another story.

How would you detect these then in a way that is as robust as autoconf?

jeroen:
> How would you detect these then in a way that is as robust as autoconf?

autoconf is not magic, it just tries `pkg-config` and otherwise falls back to some default. You can accomplish the same without the bloat and complexity of autotools.

```sh
if pkg-config --exists gmp; then
GMP_LIBS=`pkg-config --libs gmp`
else 
GMP_LIBS="-lgmp"
fi
```

szhorvat:
This does not detect if GMP is available, it simply assumes it. We need a way to actually detect it, and fall back to a vendored version otherwise. Note that the vendored version is not CRAN-friendly.

Many of our users are non-technical people. They are students, researchers, academics, in fields like sociology or biology. It is important that they be able to install igraph on Linux with as little effort as possible. Furthermore, many of them are stuck with using HPC clusters that don't have the prerequisite dependencies and less-than-helpful sysadmins ...

PR #1093, opened 2024-01-10
fix: temporary remove functions with sparse matrices

Antonov548:
https://github.com/igraph/rigraph/issues/1006


R/aaa-auto.R
```diff
@@ -3989,6 +3938,18 @@ stochastic_imitation_impl <- function(graph, vid, algo, quantities, strategies,
   res
 }
 
+has_attribute_table_impl <- function() {
```
krlmlr, line 3941: Can you please remove the two R functions too?

krlmlr (approved):
Thanks!

szhorvat:
This could have been done by adding empty entries to `functions-R.yaml`. That way we don't create conflicts with auto-generation. For example:

```
igraph_sparse_weighted_adjacency:

```

krlmlr:
Moving forward, we could add comments if we add intentional drift so that the next person that updates the generated code sees where the drift comes from.

PR #1096, opened 2024-01-11
Update vendored igraph/C to igraph/igraph@e1203464653cbcb5e87a16881ec…

Antonov548:
https://github.com/igraph/igraph/issues/2459

### DIFF
--- a/tools/update-cigraph.sh
+++ b/tools/update-cigraph.sh
@@ -18,4 +18,4 @@ rm -rf src/vendor/cigraph/.git
 
 rm -rf src/vendor/cigraph/.github src/vendor/cigraph/doc src/vendor/cigraph/examples src/vendor/cigraph/fuzzing src/vendor/cigraph/tests src/vendor/cigraph/tools
 git add src/vendor/cigraph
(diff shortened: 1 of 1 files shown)


szhorvat:
Thanks! Can you please test https://github.com/igraph/igraph/issues/2459 after merging this, _just in case_?

Antonov548:
> Thanks! Can you please test [igraph/igraph#2459](https://github.com/igraph/igraph/issues/2459) after merging this, _just in case_?

I have tried this example locally. I don't see segfault compared to main branch.

krlmlr:
I didn't see this, pushed to main.

krlmlr:
Ideally, if we already have R code that tests this, why only test locally and not add a test that will make sure that the bug doesn't return?

szhorvat:
> Ideally, if we already have R code that tests this, why only test locally and not add a test that will make sure that the bug doesn't return?

Of course. I didn't push it because I did add a test in C already. If you add a test in R, ideally it would use a different disconnected graph than what was in the original test case, to distinguish it from the C test case.

An invalid memory access should be triggered when requesting _only_ modularity and the graph is not connected.

PR #1101, opened 2024-01-11
feat!: `get.edge.ids(multi = TRUE)` is no longer supported

### LINKED ISSUE #1097: get.edge.ids() broken in 2.0
### What happens?

`get.edge.ids()` does not return the IDs of parallel edges in the same way as 1.6 did.

### To Reproduce

1.6.0:

```
> g<-make_graph(c(1,2,2,3,2,1), directed = F)
> get.edge.ids(g, c(1,2, 1,2), multi=T)
[1] 3 1
> get.edge.ids(g, c(1,2, 1,2), multi=F)
[1] 3 3
```

Current `main`:

```
> g<-make_graph(c(1,2,2,3,2,1), directed = F)
> get.edge.ids(g, c(1,2, 1,2), multi=T)
[1] 3 3
> get.edge.ids(g, c(1,2, 1,2), multi=F)
[1] 3 3
```

krlmlr:
Closes #1097.

### DIFF
--- a/R/interface.R
+++ b/R/interface.R
@@ -490,10 +490,8 @@ get.edges <- function(graph, es) {
 #' @param error Logical scalar, whether to report an error if an edge is not
 #'   found in the graph. If `FALSE`, then no error is reported, and zero is
 #'   returned for the non-existant edge(s).
-#' @param multi Logical scalar, whether to handle multiple edges properly. If
-#'   `FALSE`, and a pair of vertices are given twice (or more), then always
-#'   the same edge id is reported back for them. If `TRUE`, then the edge
-#'   ids of multiple edges are correctly reported.
+#' @param multi
+#'   `r lifecycle::badge("deprecated")`
 #' @return A numeric vector of edge ids, one for each pair of input vertices.
 #'   If there is no edge in the input graph for a given pair of vertices, then
 #'   zero is reported. (If the `error` argument is `FALSE`.)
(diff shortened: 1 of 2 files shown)


krlmlr:
The SAN failures are a consequence of a mismatch between Docker image and version of the Matrix package. We might see failures in the mainline too. I tried to fix the root cause, let's see if the fix was successful.

R/interface.R
```diff
@@ -510,23 +508,29 @@ get.edges <- function(graph, es) {
...
+get.edge.ids <- function(
+    graph,
+    vp,
+    directed = TRUE,
+    error = FALSE,
+    multi = FALSE) {
   ensure_igraph(graph)
 
+  if (lifecycle::is_present(multi)) {
+    if (isTRUE(multi)) {
```
maelle, line 527: why the difference @krlmlr?

R/interface.R
```diff
@@ -510,23 +508,29 @@ get.edges <- function(graph, es) {
...
-#' eim
-#' E(g)[eim]
-#'
-get.edge.ids <- function(graph, vp, directed = TRUE, error = FALSE, multi = FALSE) {
+get.edge.ids <- function(
+    graph,
+    vp,
+    directed = TRUE,
+    error = FALSE,
+    multi = FALSE) {
```
maelle, line 523: I'd remove the default value from the definition?

PR #1113, opened 2024-01-15
refactor: document() with latest roxygen2

maelle:
Many changes due to @krlmlr's https://github.com/r-lib/roxygen2/pull/1563 
Merging this will allow for a cleaner diff next time we have to use `document()` :grin:


maelle:
nothing changed in the end!

PR #1115, opened 2024-01-15
docs: remove scg related docs

### LINKED ISSUE #1109: There is still documentation on SCG: remove this; do not associate `stochastic_matrix()` with SCG
SCG functions were still removed, but we are still shipping active documentation on SCG. This should be removed, ideally before 2.0 ... 

There is still `R/scg.R`, and an `@scg` family. This file contains the `stochastic_matrix()` function, which is NOT related to SCG, and shouldn't be included in this file or in this family.

CC @maelle

maelle:
Fix #1109

One of the commits is from #1113


szhorvat (approved):

maelle:
@szhorvat where in the reference would `stochastic_matrix()` fit? A new subsection?

szhorvat:
The stochastic matrix is related to other graph matrices such as the adjacency matrix and the Laplacian matrix. It makes sense to include these three together in the same section.

maelle:
"Laplacian matrix" is under "Structural properties" whereas `as_adjacency_matrix()` is under "Conversion". Should I put `stochastic_matrix()` under "Structural properties"? Thank you!

szhorvat:
Let's put a "Matrices" subcategory under Structural Properties and put both the Laplacian and stochastic matrices there. If it is possible to include functions in more than one category, then let's also include `as_adjacency_matrix()` there (in addition to Conversion).

maelle:
I'll rebase after #1113

PR #1120, opened 2024-01-16
feat!: advance lifecycle deprecation from soft to warn

maelle:


krlmlr:
Happy to run revdepchecks once the build is green.

maelle:
@krlmlr the build is green now :innocent: could you please run revdeps checks? :pray: thank you!

krlmlr:
Nice! I'd like to tackle this after the 2.0.0 release. The goal of that release is to use the new C core.

maelle:
What's the next version number? Could you please run revdeps checks for this branch?

krlmlr:
For the change from soft to warn, do we really advance the version number?

I feel it's a little too early for that. Should we first move everything to {lifecycle} and give downstream some time to adapt? I suspect most users aren't even aware of the API change.

maelle:
> For the change from soft to warn, do we really advance the version number?

Oh right, no.

> I feel it's a little too early for that. Should we first move everything to {lifecycle} and give downstream some time to adapt? I suspect most users aren't even aware of the API change.

Ok.

PR #1075, opened 2024-01-04
fix: remove `R_SEXP_to_hrg` use `R_SEXP_to_hrg_copy` to avoid memory leak

Antonov548:
Closes https://github.com/igraph/rigraph/issues/1073

PR #1123, opened 2024-01-20
fix: add `DOUBLE` type

krlmlr:
Used to be part of #1075.


krlmlr:
Will redo if needed.

szhorvat:
IMO not needed. The functions that use it don't need to be exposed in R.

PR #1124, opened 2024-01-20
fix: `bfs()` no longer pads `order` with zeros with `igraph_options(return.vs.es = FALSE)`

krlmlr:
@jefferis: For https://github.com/natverse/nat/issues/517#issuecomment-1902167741.

### DIFF
--- a/tests/testthat/test-graph.dfs.R
+++ b/tests/testthat/test-graph.dfs.R
@@ -6,4 +6,7 @@ test_that("DFS uses 1-based root vertex index", {
 test_that("DFS does not pad order", {
   g <- make_star(3)
   expect_equal(as.numeric(dfs(g, root = 2, unreachable = FALSE)$order), c(2, 1))
+
+  local_igraph_options(return.vs.es = FALSE)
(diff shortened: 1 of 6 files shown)


jefferis:
@krlmlr this was great, but unfortunately we need the same change for `dfs()` as well.

tests/testthat/test-graph.dfs.R
```diff
@@ -6,4 +6,7 @@ test_that("DFS uses 1-based root vertex index", {
 test_that("DFS does not pad order", {
   g <- make_star(3)
   expect_equal(as.numeric(dfs(g, root = 2, unreachable = FALSE)$order), c(2, 1))
+
+  local_igraph_options(return.vs.es = FALSE)
+  expect_equal(as.numeric(bfs(g, root = 2, unreachable = FALSE)$order), c(2, 1))
```
jefferis, line 11: I guess this should have been 

```
expect_equal(as.numeric(**dfs**(g, root = 2, unreachable = FALSE)$order), c(2, 1))
```

krlmlr: Thanks, investigating.

PR #1128, opened 2024-01-21
test: prevent `decompose()` test from being reported as skipped

szhorvat:
I guess this test didn't need any explicit check, but without one it will be reported as skipped.


krlmlr (approved):
Thanks. It's better to have an expectation in each test, to avoid the noise.

Issue #1145, opened 2024-01-22
Could the docs be more automatically generated


maelle:
It seems we often manually port C docs to R. Would there be a way to automate the generation more? Would that necessitate a change on the C side as well (for instance storing parameter meaning in YAML/CSV etc).

@krlmlr @szhorvat

PR #1151, opened 2024-01-22
docs: improve docs of assortativity

### LINKED ISSUE #1132: Replace  @param normalized TBD
in `assortativity()` docs, probably check the C docs

maelle:
Fix #1132 

Based on https://igraph.org/c/doc/igraph-Structural.html#igraph_assortativity

@szhorvat is this correct?

### DIFF
--- a/R/assortativity.R
+++ b/R/assortativity.R
@@ -65,7 +65,7 @@ assortativity.degree <- function(graph, directed = TRUE) { # nocov start
 #'
 #' M.E.J. Newman defined two kinds of assortativity coefficients, the first one
 #' is for categorical labels of vertices. `assortativity_nominal()`
-#' calculates this measure. It is defines as
+#' calculates this measure. It is defined as
 #'
 #' \deqn{r=\frac{\sum_i e_{ii}-\sum_i a_i b_i}{1-\sum_i a_i b_i}}{
 #' r=(sum(e(i,i), i) - sum(a(i)b(i), i)) / (1 - sum(a(i)b(i), i))}
--- a/R/assortativity.R
+++ b/R/assortativity.R
@@ -95,7 +95,7 @@ assortativity.degree <- function(graph, directed = TRUE) { # nocov start
 #' is not symmetric, so it is possible to assign different values to the
 #' outgoing and the incoming end of the edges.
 #'
-#' `assortativity_degree()` uses vertex degree (minus one) as vertex values
+#' `assortativity_degree()` uses vertex degree as vertex values
 #' and calls `assortativity()`.
 #'
 #' @aliases assortativity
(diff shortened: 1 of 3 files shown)


szhorvat:
I'll need to have a more thorough read later, but the documentation should be read carefully and updated as a whole. We need to make sure that the description / formulas for assortativity match the updated included in the C core docs.

What happened here (in the C library) was that a more general version of assortativity for implemented, and the docs were significantly expanded.

maelle:
I really hope we can find a solution for #1145 :smile_cat:

szhorvat:
IMO that is not realistic, so let's not count on this happening automatically in the future. Even if it does happen, it will take many many months.

szhorvat:
Can you fix the top paragraph?

> The assortativity coefficient is positive **if** similar

Note the word I bolded, `s/is/if`

szhorvat:
Please add this somewhere at the bottom:

> Undirected graphs are effectively treated as directed ones with all-reciprocal edges. Thus, self-loops are taken into account twice in undirected graphs.

szhorvat:
There are some `sigma` which should be `\sigma` in LaTeX notation. 

image

szhorvat:
Please remove `(minus one)` from:

> assortativity_degree() uses vertex degree (minus one) as vertex values and calls assortativity().

In the end it makes no difference if one is subtracted or not.

R/assortativity.R
```diff
@@ -116,7 +116,8 @@ assortativity.degree <- function(graph, directed = TRUE) { # nocov start
 #'   `TRUE` here to do the natural thing, i.e. use directed version of the
 #'   measure for directed graphs and the undirected version for undirected
 #'   graphs.
-#' @param normalized TBD
+#' @param normalized Boolean, whether to compute the normalized covariance,
+#'  i.e. Pearson correlation. Supply true here to compute the standard assortativity.
```
szhorvat, line 120:
```suggestion
#' @param normalized Boolean, whether to compute the normalized assortativity.
#' The non-normalized nominal assortativity is identical to modularity.
#' The non-normalized value-based assortativity is simply the covariance of the
#' values at the two ends of edges.
```

We should later elaborate on what "normalized" means separately for `assortativity()` and `assortativity_nominal()`. The latter is not a Pearson correlation.

szhorvat:
I'm sorry, this was a very sloppy review by me, but with this the docs should at least be accurate and complete. We can fix them up when we have the time, after 2.0 ...

maelle:
It didn't seem sloppy! I've made changes.

maelle:
@szhorvat could you please re-review this? :pray:

szhorvat (approved):
I think this should be merged now, as it is all correct. I'll try to make improvements later. Thank you Maëlle!

PR #1165, opened 2024-01-29
refactor!: use `rlang::arg_match()` in `igraph.match.arg()`

### LINKED ISSUE #1059: Use `rlang::arg_match()` in `igraph.match.arg()`
For better error messages.

https://github.com/igraph/rigraph/issues/1058#issuecomment-1872614043

maelle:
Fix #1059

## Breaking change

`rlang::arg_match()` does not allow partial matching.

## Argument names

The `several.ok` and `choices` arguments are not used hence my respectively dropping/renaming them.

## Wondering about case sensitivity

I see that tests fail if I try making this case sensitive, because of lines such as

https://github.com/igraph/rigraph/blob/3626862d8be6f66b0cc45b3ff515a6cffc0a3511/R/community.R#L1336-L1344

The usage of `tolower()` also means the error message is not aligned with the documentation. Should I work on aligning the usage of values in code with the usage of values in documentation? For instance using "CPM" not "cpm" in the `switch()`call. However maybe there are people who use "cpm" in their function calls, which would currently work, so that might mean a breaking change.


maelle:
Should we run revdeps with this?

krlmlr:
Starting checks now.

maelle:
The revdeps checks looked good but I had forgotten to rebase this branch, I am sorry. :disappointed:

PR #1167, opened 2024-01-29
docs: remove scg related docs

### LINKED ISSUE #1109: There is still documentation on SCG: remove this; do not associate `stochastic_matrix()` with SCG

maelle:
Fix #1115
Fix #1109


szhorvat (approved):

szhorvat:
Is anything missing from this PR still @maelle ? Can we have it in Tuesday's release @krlmlr ? We should not be shipping documentation for non-existent functionality and this PR fixes that.

krlmlr:
Thanks!

Issue #915, opened 2023-10-23
alpha_centrality() crashes when `weights` argument is a string other than "weight"


davidskalinder:
**Describe the bug**
The docs for `alpha_centrality()` say that the `weights` argument is

> A character scalar that gives the name of the edge attribute to use in the adjacency matrix. If it is NULL, then the ‘weight’ edge attribute of the graph is used, if there is one. Otherwise, or if it is NA, then the calculation uses the standard adjacency matrix.

The function seems to work properly when the `weights` argument is "weight", `NULL`, or `NA`, but crashes when it's anything else.

The problem seems to be with one or more conditions in the block of `if`s in both the dense and sparse functions:

https://github.com/igraph/rigraph/blob/2810da82278d0284f22911fb596700200fcd9425/R/centrality.R#L1260

https://github.com/igraph/rigraph/blob/2810da82278d0284f22911fb596700200fcd9425/R/centrality.R#L1301

It looks like the lines that go wrong were introduced in 81e82d8f0 and/or 4dda4fd12.

#910 might obviate the need for this in the long term, so any fix might be temporary (but this logic shouldn't be reused without edits in the meantime, e.g., in #905).

**To reproduce**
``` r
library(igraph)
#> 
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#> 
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#> 
#>     union

star <- make_star(10)
E(star)$myweight <- sample(ecount(star))
alpha_centrality(star, weights = "myweight", sparse = TRUE)
#> Error in h(simpleError(msg, call)): error in evaluating the argument 'x' in selecting a method for function 't': no such edge attribute
```

<sup>Created on 2023-10-21 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

**Version information**

 - R/igraph version: 1.5.1
 - R version: 4.1.2 (2021-11-01)
 - Operating system: Ubuntu 20.04.6 LTS

```
> sessionInfo()
R version 4.1.2 (2021-11-01)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.6 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3

locale:
[...]

Issue #910, opened 2023-10-21
`alpha_centrality()` should be updated for new `as_adj()` weight handling


szhorvat:
After #906 is fixed, `alpha_centrality()` should be updated to use the new `weights` parameter of `as_adj()`.

In short, `alpha_centrality()` needs to be able to build an adjacency matrix based on an arbitrary weight vector. As of igraph 1.5, `as_adjacency_matrix()` cannot do this directly. Therefore `alpha_centrality()` copies the graph and adds a special edge attribute just to be able to transfer the weights.

maelle:
I do not understand how this case can come up:

https://github.com/igraph/rigraph/blob/238ec98707f86d81e287f841379012ab72400101/R/centrality.R#L1537-L1538

maelle:
see what I made of `alpha.centrality.dense` for now in https://github.com/igraph/rigraph/pull/1170

PR #1170, opened 2024-01-29
feat!: as_adjacency_matrix() should use weights instead of attr parameter

### LINKED ISSUE #906: `as_adjacency_matrix()` should use `weights` instead of `attr` parameter
In `as_adjacency_matrix()`, the `attr` parameter should be replaced by `weights`, to bring this in function in line with others that deal with weighted graphs.

 - `attr` takes an edge attribute name uses the elements from it as weights.
 - `weights` takes a weight vector.

This change was already made in the C core for 0.10, so the R interface should mirror it if we want to keep things simple. Usign `weights` is also more flexible as it makes it possible for weights to come from elsewhere than an attribute, which is important for implementing certain kinds of functionality (see e.g. #905).

The change can be made gradually, with supporting both `attr` and `weights` for a while.

maelle:
Fix #906 

- How to use weights in the indexing function.
- That function has an attr argument that is passed to as_adjacency_matrix(). Should we double all calls based on whether attr is missing or not, to avoid the deprecation message.

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -1486,26 +1486,19 @@ alpha.centrality.dense <- function(graph, nodes = V(graph), alpha = 1,
 
   exo <- rep(exo, length.out = vcount(graph))
   exo <- matrix(exo, ncol = 1)
-
-  if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
-    ## weights == NULL and there is a "weight" edge attribute
-    attr <- "weight"
-  } else if (is.null(weights)) {
-    ## weights == NULL, but there is no "weight" edge attribute
-    attr <- NULL
+  if (is.null(weights) && is_weighted(graph)) {
+    weights <- get_weights(graph)
   } else if (is.character(weights) && length(weights) == 1) {
     ## name of an edge attribute, nothing to do
-    attr <- "weight"
-  } else if (any(!is.na(weights))) {
-    ## weights != NULL and weights != rep(NA, x)
-    graph <- set_edge_attr(graph, "weight", value = as.numeric(weights))
-    attr <- "weight"
-  } else {
-    ## weights != NULL, but weights == rep(NA, x)
-    attr <- NULL
+    ## ???
+    weights <- get_weights(graph)
   }
 
-  d <- t(as_adj(graph, attr = attr, sparse = FALSE))
+  if (is.na(weights)) {
+    weights <- NULL
+  }
+
+  d <- t(as_adjacency_matrix(graph, weights = weights, sparse = FALSE))
   if (!loops) {
     diag(d) <- 0
   }
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -1486,6 +1486,7 @@ alpha.centrality.dense <- function(graph, nodes = V(graph), alpha = 1,
 
   exo <- rep(exo, length.out = vcount(graph))
   exo <- matrix(exo, ncol = 1)
+
   if (is.null(weights) && is_weighted(graph)) {
     weights <- get_weights(graph)
   } else if (is.character(weights) && length(weights) == 1) {
@@ -1494,6 +1495,7 @@ alpha.centrality.dense <- function(graph, nodes = V(graph), alpha = 1,
     weights <- get_weights(graph)
   }
 
+  # seems weights = NA happens
   if (!is.null(weights) && length(weights) == 1 && is.na(weights)) {
     weights <- NULL
   }
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -1494,7 +1494,7 @@ alpha.centrality.dense <- function(graph, nodes = V(graph), alpha = 1,
     weights <- get_weights(graph)
   }
 
-  if (is.na(weights)) {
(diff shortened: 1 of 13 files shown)


R/conversion.R
```diff
@@ -176,19 +177,21 @@ get.adjacency.dense <- function(graph, type = c("both", "upper", "lower"),
...
-      res <- matrix(0, nrow = vcount(graph), ncol = vcount(graph))
-    } else {
-      stop(
-        "Matrices must be either numeric or logical, ",
-        "and the edge attribute is not"
+
+    res <- switch(
+      typeof(exattr),
+      logical = matrix(FALSE, nrow = vcount(graph), ncol = vcount(graph)),
+      double = matrix(0, nrow = vcount(graph), ncol = vcount(graph)),
```
maelle, line 187: I am doubting myself reg double vs numeric.

man/get.adjacency.Rd
```diff
@@ -22,15 +22,16 @@ right triangle of the matrix is used, \code{lower}: the lower left triangle
 of the matrix is used. \code{both}: the whole matrix is used, a symmetric
 matrix is returned.}
 
-\item{attr}{Either \code{NULL} or a character string giving an edge
+\item{attr}{\ifelse{html}{\href{https://lifecycle.r-lib.org/articles/stages.html#deprecated}{\figure{lifecycle-deprecated.svg}{options: alt='[Deprecated]'}}}{\strong{[Deprecated]}} Use \code{weights} instead.
```
maelle, line 25: I wonder how to best handle the deprecation of the argument in this... deprecated function. :zany_face:

R/conversion.R
szhorvat: Just for my R eduction, is there a difference between `double` and `numeric`?

szhorvat:
Just in case you haven't seen them, I'm linking #910 and #915 (and later #903)

maelle:
Thanks! I'll come back to this later this week!

R/conversion.R
maelle: I fixed it! numeric = double AND integer. See example below. To be honest I had to look it up yesterday. :sweat_smile: 

``` r
mode(1L)
#> [1] "numeric"
typeof(1L)
#> [1] "integer"
mode(1.0)
#> [1] "numeric"
typeof(1.0)
#> [1] "double"
```

<sup>Created on 2024-01-30 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

maelle: so in the switch I am now using the mode not the type.

R/indexing.R
```diff
@@ -190,34 +220,20 @@
       if (any(res != 0)) {
         res[res != 0] <- edge_attr(x, attr, res[res != 0])
       }
+    } else if (!is.null(weights)) {
+      # TODO how to use weights here?!
```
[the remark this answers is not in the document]
szhorvat: First, I am not familiar with this part of the code, and I haven't really used the `g[ ... ]` syntax to treat the graph as a matrix.

But it seems to me that the problem is what I pointed out when the `multi` argument of `get.edge.ids()` was deprecated because it wouldn't function anymore.

We need a function that _can_ return multiple edges between two vertices correctly, and we don't have that now. We need an interface to `igraph_get_all_eids_between()` first, before you can proceed here.

This is why I keep protesting about doing partial upgrades, removing functionality (due to changes in C) but without exposing an interface to the replacement.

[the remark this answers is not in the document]
szhorvat: It might be necessary (for reasonable performance) to implement part of this in pure C, and perhaps add it to the C core directly. It's late here now and I can't quite see how to best do this. Let me get back to you later.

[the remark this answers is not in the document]
szhorvat: https://github.com/igraph/igraph/issues/2476

R/centrality.R
```diff
@@ -1487,25 +1487,20 @@ alpha.centrality.dense <- function(graph, nodes = V(graph), alpha = 1,
...
 
-  if (is.null(weights) && "weight" %in% edge_attr_names(graph)) {
-    ## weights == NULL and there is a "weight" edge attribute
-    attr <- "weight"
-  } else if (is.null(weights)) {
-    ## weights == NULL, but there is no "weight" edge attribute
-    attr <- NULL
+  if (is.null(weights) && is_weighted(graph)) {
+    weights <- get_weights(graph)
   } else if (is.character(weights) && length(weights) == 1) {
```
maelle, line 1492: @szhorvat should the functions `alpha.centrality.dense()` and `alpha.centrality.sparse()` actually error when provided `weights` as a character?

maelle:
So, to go forward with this PR despite the blockage in indexing.R here is what I suggest I do

- I make the refactoring changes independently (for instance the change to `switch()`)
- I make the soft deprecation in centrality.R, and conversion.R. In indexing.R I keep using the attr argument with with a local option so that we don't get our own deprecation message. 
- I open an issue about the need for a new function to completely fix indexing.R

Does that sound good @szhorvat?

krlmlr:
Abandoning, will pick up when we get back to #906.

Issue #1045, opened 2023-12-29
Add ellipsis


krlmlr:
https://github.com/igraph/rigraph/pull/1028#discussion_r1438319072

krlmlr:
@ntamas: Works like a charm, idempotent, fd74ecf7ad17bd741578cb1908da6fb190989798 is an example of an added `*` . We'll wait for those to be available in the C core.

What does it take to add a `check_dots_empty()` call when a `*` is present in `PARAMS`?

ntamas:
`check_dots_empty()` added in Stimulus 0.21.1.

krlmlr:
Works as expected, in 894015601109f56718977aa85d4c02dacd4679e9.

szhorvat:
> `check_dots_empty()`

Is this function waiting to be implemented? Let me know when that's done and I'll continue with #1028 afterwards.

szhorvat:
Ah, so this is an rlang thing. Should it be called as `rlang::check_dots_empty()`? Or what is the solution to the following?

```r
> k_shortest_paths(make_ring(5),1,2)
Error in check_dots_empty() : could not find function "check_dots_empty"
```

szhorvat:
I guess I need to add `#' @importFrom rlang check_dots_empty`?

krlmlr:
Thanks, missed that. Waiting for `*` to be implemented in the core, eager to roll out here and run revdepchecks.

krlmlr:
One more thing: in my book, the `weights` argument should always come after the ellipsis (if it exists, which I assume it will for most functions). This is contrary to https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering .

https://github.com/igraph/rigraph/pull/1028#discussion_r1438727059

ntamas:
The guidelines you cited above are for the C core. High-level interfaces are free to come up with their own guidelines in order to make the generated functions fit better with the conventions of the high-level language.

krlmlr:
Could Stimulus take care of that for the autogenerated code? Do we agree it's a good idea?

ntamas:
It could, but I'm not sure whether it's in the scope of the project.

I'd say that in the long term Stimulus should probably be refactored such that the repo contains the core only and the individual language-specific generators are in the repositories of the higher-level interfaces (with some kind of a plug-in system using Python's standard entry points from `importlib`). If that happens, you could do whatever you want with the R code generator as it would be entirely within your domain.

[...]

PR #1173, opened 2024-01-29
feat: New `voronoi_cells()` to compute the Voronoi partitioning of a graph

szhorvat:
@krlmlr @maelle Please have a look before I add a test.


R/community.R
```diff
@@ -2782,3 +2782,44 @@ communities <- groups.communities
...
+#' vertices. Each vertex is assigned to the generator vertex from (or to) which
+#' it is closest.
+#'
+#' [groups()] may be used on the output of this function.
+#'
+#' @param graph The graph to partition into Voronoi cells.
+#' @param generators The generator vertices of Voronoi cells.
+#' @inheritParams distances
+#' @param mode Character string. In directed graphs, whether to compute
+#'   distances from generator vertices to other vertices (\dQuote{out}), to
```
maelle, line 2802: instead of dQuote you can use a back tick.  https://roxygen2.r-lib.org/articles/rd-formatting.html#code

```suggestion
#'   distances from generator vertices to other vertices (`out`), to
```

R/community.R
```diff
@@ -2782,3 +2782,44 @@ communities <- groups.communities
...
+#'   generatorvertices from other vertices (\dQuote{in}), or ignore edge
+#'   directions entirely (\dQuote{all}). Ignored in undirected graphs.
+#' @param tiebreaker Character string that specifies what to do when a vertex
+#'   is at the same distance from multiple generators. \dQuote{random} assigns
+#'   a minimal-distance generator randomly, \dQuote{first} takes the first one,
+#'   and \dQuote{last} takes the last one.
+#' @return A named list with two components:
+#'   \item{membership}{numeric vector giving the cluster id to which each vertex
```
maelle, line 2810: https://roxygen2.r-lib.org/articles/rd-formatting.html#lists

Use "*" instead of item

R/community.R
```diff
@@ -2782,3 +2782,44 @@ communities <- groups.communities
...
+#'   belongs.}
+#'   \item{distances}{numeric vector giving the distance of each vertex from its
+#'   generator}
+#' @author Szabolcs Horvát \email{szhorvat@gmail.com}
```
maelle, line 2814: maybe a question beyond this PR, do we want to keep registering authors of single functions?

R/aaa-auto.R
```diff
@@ -739,8 +739,9 @@ distances_floyd_warshall_impl <- function(graph, from=V(graph), to=V(graph), wei
   res
 }
 
-voronoi_impl <- function(graph, generators, weights=NULL, mode=c("out", "in", "all", "total"), tiebreaker=c("random", "first", "last")) {
+voronoi_impl <- function(graph, generators, ..., weights=NULL, mode=c("out", "in", "all", "total"), tiebreaker=c("random", "first", "last")) {
   # Argument checks
+  check_dots_empty()
```
maelle, line 744: have we documented this pattern somewhere? it looks familiar but I'm not entirely sure I understand.

maelle (commented):
Thanks! I added a few comments.

R/community.R
szhorvat: So like this?

```
 * `membership` numeric vector giving ...
```

R/community.R
szhorvat: I would cautiously vote for no, because eventually multiple people will end up contributing.

However, I think we should give more credit to external people who contribute, to encourage them. What would be a good way to do that?

 - We already have the all-contributors grid in the READM
 - Include their name in the changelog
 - Any other ideas @maelle ?
 - Maintain  contributors

R/aaa-auto.R
szhorvat: This is auto-generated, not written manually by me. The purpose is to make sure that any parameters following `...` are specified explicitly, with name. 

https://rlang.r-lib.org/reference/check_dots_empty.html

See #1045

CC @krlmlr

szhorvat:
Thanks for the review @maelle !

R/aaa-auto.R
krlmlr: Yes, the `*` in `functions-R.yaml` .

R/community.R
krlmlr: We have this pattern everywhere else in the codebase: `@return ... \item{} ...` . I checked the rendered output, it looks good. I propose to keep it as is.

tools/stimulus/functions-R.yaml
```diff
@@ -367,6 +367,10 @@ igraph_distances_johnson:
 igraph_distances_floyd_warshall:
 
 igraph_voronoi:
+    PARAMS: |-
+        GRAPH graph, OUT VECTOR_INT membership, OUT VECTOR distances,
+        VERTEX_INDICES generators, *, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT,
+        VORONOI_TIEBREAKER tiebreaker=RANDOM
```
krlmlr, line 373: Seems to me this could also be a change in the C core? How can we make it easy to perhaps implement the change in `src/vendor/igraph` and then backport it to the C core repo?

krlmlr (commented):
Looks good, thanks!

krlmlr:
Why do the checks fail though?

krlmlr:
Undocumented arguments in documentation object 'voronoi'
    ‘...’
  
See https://rlang.r-lib.org/reference/args_dots_empty.html?q=args_dots_e#null .

tools/stimulus/functions-R.yaml
szhorvat: I'm not sure this will work very well in the general case, so I'm not sure adding the `*` in the C repo is a good idea. Different interfaces might want that `*` to go in different places, and then there's the convention that in C `weights` now goes directly after `graph`, but `weights` should always be keyword-only in R, so we need to handle that somehow. Perhaps with reordering like we did with `k_shortest_paths`. 

However, doing this in the C core does not affect me or the Mathematica interface, at least not at this moment.  I propose you discuss this with @ntamas, and we go ahead with whatever the two of you decide.

Merging this PR is not urgent for me.

szhorvat:
Thanks for the comments, @krlmlr. This PR is not urgent for me. I want to take this slow and use the opportunity to learn how to best do this.

tools/stimulus/functions-R.yaml
ntamas: One option is to leave the `*` out in the C codebase and use `FIRST_KW_PARAM` in `functions-R.yaml` to tell Stimulus which parameter (by name) is the first one that is keyword-only.

If we want to do parameter reordering _and_ declaring them as keyword-only for a function, there's also an option for that: you could declare the new order in `PARAM_ORDER` _and_ insert a `*` in the parameter list on `PARAM_ORDER` at the appropriate point.

szhorvat: Used `FIRST_KW_PARAM`

szhorvat:
This PR has revealed a larger bug: in auto-generated code, vectors created with `R_SEXP_to_vector_int_copy` are not freed. Can you please have a look at this @Antonov548? Similar memory leak bugs may be present in existing functions.

szhorvat:
@Antonov548 In `typec-RC.yaml`, I can see that some of non-allocating conversions were changed to allocating ones without adding cleanup code. Please always add the cleanup code as well, or if this is not trivial, do create a reminder issue to do it.

Whenever there is allocation in the INCONV section, there needs to be deallocation in the OUTCONV section. The `IN` part of at least `VERTEX_INDICES`, `VERTEX_INDEX_PAIRS` are affected. Can you add the deallocation and check if anything else is missing it?

It seems to me that there are lots of memory leaks at the moment.

szhorvat:
@krlmlr Now that the memory leak issue is fixed, this PR is ready for merging or another review (but not urgent).

krlmlr:
@maelle: Do these new featores conflict with the lifecycle work?

@szhorvat: Should we use a more descriptive name, perhaps `voronoi_part()` or `voronoi_partitioning()` ?

szhorvat:
I think `voronoi()` is sufficiently descriptive, and consistent with how other functions are named. But it's good to be aware that there are several other R packages which have a function of the same name, which compute a geometric Voronoi tessellation: https://rdrr.io/search?q=voronoi Some of these are methods, sometimes it is even a class name. At this point it becomes a technical decision what to do, so I'll leave it to you.

If we do go with a different name, I'd like to use `voronoi_cells()`.

maelle:
> @maelle: Do these new features conflict with the lifecycle work?

@krlmlr no they do not, especially as adding an experimental function does not require registering for which version it was added.

krlmlr:
Thanks. `voronoi()` -> `voronoi_cells()` sounds good to me, we'd also need to update the PR's title. Postponing.

szhorvat:
Renamed to `voronoi_cells()`.

PR #1186, opened 2024-02-01
feat: `distances()` now supports the Floyd-Warshall algorithm

szhorvat:


krlmlr:
Thanks. Moving forward, should we add a test for every new feature? We could also label the new algorithm as "experimental".

szhorvat:
I don't think we need to mark this as experimental in R. The reason why it is in C is that the interface might change there (but not the behaviour—it will always return a distance matrix).

Yes, I'll add tests.

This is not urgent to merge.

krlmlr:
Thanks!

krlmlr (approved):

PR #1204, opened 2024-02-05
feat: `is_biconnected()` checks if a graph is biconnected

szhorvat:
Includes auto-gen update for just for this function, but all new additions after the recent C core update.

Marked as experimental out of principle, since it's experimental in C as well. But this will almost certainly not change.

Not urgent, but safe for tomorrow's release.

I know the name of the branch is incorrect, that was an accident, but it doesn't matter.


krlmlr:
Thanks!

Issue #1189, opened 2024-02-03
Detect availability of `__attribute__ ((deprecated))`


krlmlr:
in `./configure`, and use conditionally.

Per https://github.com/conda-forge/r-igraph-feedstock/pull/56.

Or simply avoid using this in CRAN releases.

szhorvat:
Marking enum values as deprecated is not extremely useful (not comparable to marking functions as deprecated). It was me who added this to the C core, to have a bit more security than just _documenting_ some values as deprecated. In hindsight, we didn't get much use out of it.

So I'd suggest just defining `IGRAPH_DEPRECATED_ENUMVAL` to empty, since in R we don't have the fancy CMake mechanism to easily check the availability of various compiler extensions.

szhorvat:
Removing from phoenix, since a quickfix was committed with #1205, which is good for the next release.

Leaving it open for now in case you want proper detection. If so, you can use [this test code](https://github.com/igraph/igraph/blob/c22d6b21b7ec02b509803b39f54b13038dbb6460/etc/cmake/attribute_support.cmake#L17).

I changed the title to be accurate: This is not about detecting if `__attribute__ ((deprecated))` is available in general—it is available as a function attribute in all GCC/Clang versions we care about. It's about whether it can be used with enum value, which is a very niche use case.

krlmlr:
I propose to try adding a `cmake` call to `./configure`, and to encode the relevant rules in our `CMakeLists.txt` . Currently:

```sh
cmake -Ssrc/vendor/cigraph -Bsrc/vendor/cigraph/build
# -- Version number from Git: v2.0.2.9007-4-g5cece5071b
# CMake Error at etc/cmake/version.cmake:66 (list):
#   list index: 1 out of range (-1, 0)
# Call Stack (most recent call first):
#   CMakeLists.txt:31 (include)
# 
# 
# CMake Error at etc/cmake/version.cmake:67 (list):
#   list index: 2 out of range (-1, 0)
# Call Stack (most recent call first):
#   CMakeLists.txt:31 (include)
# 
# 
# CMake Error at CMakeLists.txt:34 (project):
#   VERSION "v2.0.2.9007" format invalid.
# 
# 
# -- Configuring incomplete, errors occurred!
```

Is there a way to make version detection optional? Can/should we add our top-level `CMakeLists.txt`?

krlmlr:
(If `cmake` is not available, we can fall back to sensible defaults.)

PR #1205, opened 2024-02-05
fix: workaround for deprecated enum values not being supported with old GCC

szhorvat:
@krlmlr Quick workaround for #1189 in the interest of getting the quickfix release out soon.


krlmlr:
Thanks!

PR #1206, opened 2024-02-05
docs: document 'dsatur' heuristic for greedy_vertex_coloring()

szhorvat:
@krlmlr Please merge for the next release.


krlmlr:
Absolutely, please use the mergequeue label for these things. One request though: if you could use Markdown for the PR title, as I did here, it would save a bit of time. This is what is collected into the changelog which then becomes the release notes with a manual process.

Issue #1176, opened 2024-01-30
Go back to autogenerating all code


krlmlr:
I have allowed drift to speed up the 2.0.0 release. We need to recover from that to be functional again.

Each deviation from autogenerated code in `aaa-auto.R` is there for a reason.

maelle:
goal: edit the Stimulus stuff until running it creates the current R code

maelle:
```
docker run --rm -ti --platform linux/amd64 -v $(pwd):/root/workspace ghcr.io/cynkra/rig-ubuntu-igraph:main make -f Makefile-cigraph -B R/aaa-auto.R
```

maelle:
maybe add ignore as in #1044

szhorvat:
Drift is fixed right now.

maelle:
Re-opening to pick up #1044 at some point

krlmlr:
We're good here now.

PR #1207, opened 2024-02-06
refactor: eliminate manual edits from autogenerated files

szhorvat:
@krlmlr Can we please have this merged now to reduce the pain of having to untangle manual edits to autogenerated files whenever re-running stimulus?

Issues opened:

 - https://github.com/igraph/stimulus/issues/6

Explanation of changes:

 - `is_bipartite` and `realize_bipartite_degree_sequence` are new in 0.10.9, hence added due to the C core update
 - `has_attribute_table` and `igraph_version` were edited to fix a compiler warning, but we don't actually use them, so removed; issue opened in stimulus
 - `get_subisomorphisms_vf2_callback`: the generated R code for this was manually removed, and it's simpler to also remove the generated C code until we actually need it.

Ref: #1176

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -215,6 +215,21 @@ realize_degree_sequence_impl <- function(out.deg, in.deg=NULL, allowed.edge.type
   res
 }
 
+realize_bipartite_degree_sequence_impl <- function(degrees1, degrees2, allowed.edge.types=c("simple", "loops", "multi", "all"), method=c("smallest", "largest", "index")) {
+  # Argument checks
+  degrees1 <- as.numeric(degrees1)
+  degrees2 <- as.numeric(degrees2)
+  allowed.edge.types <- switch(igraph.match.arg(allowed.edge.types),
+    "simple"=0L, "loop"=1L, "loops"=1L, "multi"=6L, "multiple"=6L, "all"=7L)
+  method <- switch(igraph.match.arg(method), "smallest"=0L, "largest"=1L, "index"=2L)
+
+  on.exit( .Call(R_igraph_finalizer) )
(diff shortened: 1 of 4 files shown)


szhorvat:
Update: Second commit eliminates the last remaining manual edit. I'm not even sure if this was a manual edit, or the result of a C-side bugfix in the interface definition.

R/aaa-auto.R
```diff
@@ -970,9 +985,7 @@ edge_betweenness_subset_impl <- function(graph, eids=E(graph), directed=TRUE, so
   on.exit( .Call(R_igraph_finalizer) )
   # Function call
   res <- .Call(R_igraph_edge_betweenness_subset, graph, eids-1, directed, sources-1, targets-1, weights)
-if (igraph_opt("add.vertex.names") && is_named(graph)) {
-    names(res) <- vertex_attr(graph, "name", V(graph))
-  }
```
krlmlr, line 975: Is this intentional?

krlmlr (approved):
Thanks for fixing the drift! Please add the "mergequeue" label when good.

R/aaa-auto.R
szhorvat: This was incorrect, so I allowed it to be removed by the auto-generation. This is trying to add vertex names to a vector that contains edge-quantities, not vertex quantities. I also removed a dependence of the result on `V(g)` in `functions-R.yaml` for the same reason, although this seemed to have no ill effect.

If you find anything else that looks suspicious, let me know!

szhorvat: Also: As far as I can tell, we're not exposing this function at the moment, so this bug did not have any consequences so far.

PR #1211, opened 2024-02-06
feat: weight support for `eccentricity()` and `radius()`

### LINKED ISSUE #893: Could the functions eccentricity() and radius() be implemented for weighted graphs?
Could the functions `eccentricity()` and `radius()` be updated to support weighted graphs, similar to how `distances()` and `diameter()` do when passed the weights argument?

I know I can calculate eccentricity and radius for a weighted graph like this:
```{r}
g <- sample_pa(1000, directed = FALSE)
E(g)$weight <- runif(ecount(g))
my_eccentricity <- apply(distances(g), 1, max)
my_radius <- min(my_eccentricity)
```
However, for consistency and usability, I think it would be better if `eccentricity()` and `radius()` accepted a `weights` argument like `distances()` and `diameter()`.

szhorvat:
Fixes #893


szhorvat:
Waiting on https://github.com/igraph/stimulus/issues/7

szhorvat:
@maelle I need your help and advice here as this may require some lifecycle work.

maelle:
> some changes to arguments, but also because the weight edge attribute is now used automatically, if present.

What changes to arguments exactly?

Regarding the use of the attribute, how did user use to opt in? Could there be a way for users to still opt in?

szhorvat:
> Regarding the use of the attribute, how did user use to opt in? Could there be a way for users to still opt in?

It is a bad idea to make this opt-in as it creates inconsistencies and unpredictability in the long term. Let's not do that. If there is a `weight` attribute, and the function supports it, it should always be used automatically.

There's a way to opt out with `weights=NA`.

szhorvat:
How shall we proceed with this?

maelle:
It's in my backlog, I still intend to come back to it... later.

krlmlr:
I spoke too soon, this is actually a breaking change. Example for compatibility code:

```patch
diff --git a/R/community.R b/R/community.R
index dd1272ca36..7d9a0b83c4 100644
--- a/R/community.R
+++ b/R/community.R
@@ -677,7 +677,34 @@ modularity <- function(x, ...) {
 #' modularity(wtc)
 #' modularity(g, membership(wtc))
 #'
-modularity.igraph <- function(x, membership, weights = NULL, resolution = 1, directed = TRUE, ...) {
+modularity.igraph <- function(x, membership, ..., weights = NULL, resolution = 1, directed = TRUE) {
+
+  if (...length() > 0) {
+      "3.0.0",
+      "modularity.igraph(... =)",
+      details = "The arguments `weights`, `resolution` and `directed` must be named."
+    )
+
+    dots <- list(...)
+    stopifnot(all(names2(dots) == ""))
+
+    if (is.null(weights) && length(dots) > 0) {
+      weights <- dots[[1]]
+      dots <- dots[-1]
+    }
+
+    if (missing(resolution) && length(dots) > 0) {
+      resolution <- dots[[1]]
+      dots <- dots[-1]
+    }
+
+    if (missing(directed) && length(dots) > 0) {
+      directed <- dots[[1]]
+      dots <- dots[-1]
+    }
+  }
+
   # Argument checks
   ensure_igraph(x)
   if (is.null(membership) || (!is.numeric(membership) && !is.factor(membership))) {
```

maelle:
I'll come back to this PR next week.

R/paths.R
```diff
@@ -294,7 +308,45 @@ eccentricity <- eccentricity_impl
...
+#'
+#' The center of a graph is the set of its vertices with minimal eccentricity.
+#'
+#' @inheritParams eccentricity
+#' @inheritParams rlang::args_dots_empty
+#' @return The vertex IDs of the central vertices.
+#' @seealso [eccentricity()], [radius()]
+#' @family paths
+#' @export
+graph_center <- graph_center_dijkstra_impl
```
maelle, line 349: this is new so it does not need any lifecycle infrastructure :baby_chick:

R/paths.R
```diff
@@ -259,29 +255,47 @@ max_cardinality <- maximum_cardinality_search_impl
...
-eccentricity <- eccentricity_impl
+eccentricity <- function(graph, vids = V(graph), ..., weights = NULL, mode = c("all", "out", "in", "total")) {
+    if (...length() > 0) {
+    lifecycle::deprecate_soft(
+      "2.0.2",
+      "eccentricity(... =)",
+      details = "The arguments `weights` and `mode` must be named."
+    )
+
+    rlang::check_dots_unnamed()
```
maelle, line 266: It seemed easier to use the rlang function rather than adding something home-made.

maelle:
- I added the lifecycle infrastructure. Does this seem correct @krlmlr?
- I added test cases. Do they make any sense @szhorvat? If so it would make sense to add the same test cases as examples especially for `graph_center()` that currently has no example.

R/paths.R
```diff
@@ -259,29 +255,47 @@ max_cardinality <- maximum_cardinality_search_impl
 #' eccentricity(g)
 #' @family paths
 #' @export
-eccentricity <- eccentricity_impl
+eccentricity <- function(graph, vids = V(graph), ..., weights = NULL, mode = c("all", "out", "in", "total")) {
+    if (...length() > 0) {
+    lifecycle::deprecate_soft(
+      "2.0.2",
```
krlmlr, line 261:
```suggestion
      "2.1.0",
```

Bad Copilot suggestion?

krlmlr (approved):
Looks good, thanks!

szhorvat:
> * 
> 
> * I added test cases. Do they make any sense @szhorvat? If so it would make sense to add the same test cases as examples especially for `graph_center()` that currently has no example.

I had a very quick look, seems fine as a smoke test, and yes, the example is actually a good one for `graph_center()`.

krlmlr:
`cpp11::cpp_register()`

szhorvat:
I merged `main` into this branch and fixed the conflict by regenerating this file using `R -q -e 'cpp11::cpp_register()'`

krlmlr:
Aviator might be fine, I can't identify the reason why it sometimes fails to rebase.

maelle:
thanks both!

PR #1212, opened 2024-02-06
feat: `realize_bipartite_degseq()` creates a bipartite graph from two degree sequences

szhorvat:


szhorvat:
@krlmlr @maelle I need your help here. I'm not sure how to best deal with this:

```
> realize_bipartite_degseq(c(1,1), c(2))
  'arg' must be of length 1
```

This comes from `match.args()` happening in `realize_bipartite_degree_sequence_impl` and not in `realize_bipartite_degseq`. Can you suggest specific code to fix this?

szhorvat:
I pushed a commit with one possible fix. Let me know if this is the idiomatic way to handle this.

szhorvat:
@maelle I'm confused about why the pkgdown test fails. The documentation of this function is structured exactly the same way as that of `realize_degseq()`. What is the difference between them that causes the failure?

krlmlr:
The repeated argument matching seems like a safe way to handle this, yes.

We need to add this and every other new function to `_pkgdown.yml` . Some are added automatically via concepts or name patterns: https://github.com/igraph/rigraph/blob/4875489730791962778941a86f23f84c7816b0c0/_pkgdown.yml#L39-L42

Issue #1203, opened 2024-02-05
biconnected_components() returns invalid result fields


szhorvat:
`biconnected_components()` returns a result list with named components `tree_edges`, `component_edges`, `articulation_points`.  These list components have valid values, However, the result _also_ has components `tree.edges`, `component.edges` and `articulation.points`. These are empty.

Note that `biconnected_components()` is fully auto-generated, but somewhere during auto-generation the `_` is changed to a `.`, but only in some places, not everywhere.

```
> res<-biconnected_components(make_graph(c(1,2, 2,3, 3,1, 1,4), directed = F))
> res
$no
[1] 2

$tree_edges
$tree_edges[[1]]
[1] 2 1

$tree_edges[[2]]
[1] 4


$component_edges
$component_edges[[1]]
[1] 3 2 1

$component_edges[[2]]
[1] 4


$components
$components[[1]]
+ 3/4 vertices, from c488bbb:
[1] 3 2 1

$components[[2]]
+ 2/4 vertices, from c488bbb:
[1] 4 1


$articulation_points
[1] 1

$articulation.points
+ 0/4 vertices, from c488bbb:
```

Also notice the function definition:

```
biconnected_components_impl <- function(graph) {
  # Argument checks
  ensure_igraph(graph)

  on.exit( .Call(R_igraph_finalizer) )
  # Function call
  res <- .Call(R_igraph_biconnected_components, graph)
  if (igraph_opt("return.vs.es")) {
    res$tree.edges <- lapply(res$tree.edges, unsafe_create_es, graph = graph, es = E(graph))
  }
  if (igraph_opt("return.vs.es")) {
    res$component.edges <- lapply(res$component.edges, unsafe_create_es, graph = graph, es = E(graph))
  }
  if (igraph_opt("return.vs.es")) {
    res$components <- lapply(res$components, unsafe_create_vs, graph = graph, verts = V(graph))
  }
  if (igraph_opt("return.vs.es")) {
    res$articulation.points <- create_vs(graph, res$articulation.points)
  }
  res
}
```

ntamas:
The replacement is done in Stimulus in `src/stimulus/generators/r.py` in the function named `get_r_parameter_name` -- that part of the code is actually replacing underscores with dots. I don't have more time to debug this, but start looking there.

szhorvat:
This broke between 1.2.11 and 1.3.0. Isn't that when stimulus was separated into a library?

1.2.11 used underscore names, `res$articulation_points`

[...]

PR #1213, opened 2024-02-07
fix: `biconnected_components()` now returns edge and vertex sequences again

szhorvat:
This is a temporary workaround for #1203 until auto-generation is updated to handle this situation. It restores the 1.2.11 behaviour.

@krlmlr I recommend this for the next bugfix release, but we'll have to find a proper solution later, with auto-generation.

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -2235,6 +2235,28 @@ articulation_points_impl <- function(graph) {
   res
 }
 
+biconnected_components_impl <- function(graph) {
+  # Argument checks
+  ensure_igraph(graph)
+
+  on.exit( .Call(R_igraph_finalizer) )
+  # Function call
+  res <- .Call(R_igraph_biconnected_components, graph)
+  if (igraph_opt("return.vs.es")) {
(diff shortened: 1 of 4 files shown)


krlmlr:
Thanks. We want to move away from using `.Call()` outside `aaa-auto.R` . Can we achieve the same by merely wrapping `biconnected_components_impl()` instead of killing it?

krlmlr:
Can you please add a test? Happy to propose something.

szhorvat:
I won't have time for a test at the moment, but since this is an obvious bugfix, it would be good to get it out soon ...

Of course eventually we'll need a test. If we had one, this breakage would have been detected.

szhorvat:
@krlmlr There is already a test for this function. What we need to do is make sure that the components of the result list are named as expected, and are of the correct class (`igraph.vs` and not `vector`). How do you do this? Just check `names()` and `class()`?

krlmlr:
[5556282](https://github.com/igraph/rigraph/pull/1213/commits/5556282c22abf2b8fc0fd703cc272531bb93fbf3) is what I mean by wrapping. I believe there's always a way, our autogeneration is not that bad.

The other commits are cosmetics and add a test.

PR #1215, opened 2024-02-07
fix: Link Fortran runtime on Windows as needed by arpack.

kalibera:
Fortran libraries (Fortran runtime) is needed e.g. by arpack. This change is to fix building on LLVM/aarch64.

### DIFF
--- a/src/Makevars.ucrt
+++ b/src/Makevars.ucrt
@@ -11,6 +11,6 @@ PKG_CPPFLAGS = -I"${LIB_XML}/include/libxml2" -I"${LIB_XML}/include" -DLIBXML_ST
         -D_GNU_SOURCE=1
 
 PKG_LIBS = -L"${LIB_XML}/lib" -lxml2 -liconv -lz -lws2_32 -lstdc++ \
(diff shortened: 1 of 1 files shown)


krlmlr (approved):
Thanks. Do we need the same change in `Makevars.win` ?

kalibera:
> Thanks. Do we need the same change in `Makevars.win` ?

It would be probably cleaner to do so, because Fortran runtime is used. But you probably don't need it, certainly not for aarch64.

I am not sure if you need Makevars.win at all, maybe you could delete it and rename the current Makevars.ucrt to Makevars.win. That depends on how old versions of R you want to support (on CRAN).

Makevars.ucrt is used by R >= 4.2.0, when it exists. When it doesn't exist, R >= 4.2.0 uses Makevars.win as well as the older versions of R do (details are in Writing R Extensions).

At least with R >= 4.2.0 (and so Rtools >= 42), you don't even need to specify "-I" options for LIB_XML, GLPK_HOME, LIB_GMP, the R_TOOLS_SOFT directory is by default made available by the toolchain. So you might need a distinct Makevars.win only if you wanted to support R < 4.2 (with Rtools < 42).

PR #1219, opened 2024-02-08
fix: `isomorphisms()` and `subgraph_isomorphisims(method = "vf2")` work again, regression introduced in 2.0.0

### LINKED ISSUE #1218: Error in `igraph::subgraph_isomorphisms`:  REAL() can only be applied to a 'numeric', not a 'integer'
### What happens?

Since the major release of the igraph package, I get some weird error which I cannot trace down because it seems to be related to some internal calls from the R function to the C code. That is, getting the error message, I have no idea what I could do to get rid of this error. Either it might be a bug in igraph, or I need some help on how to fix this problem.

So, the problem occurs when using `subgraph_isomorphisms` function with method "vf2" and specifying vertex colors. Then, the following error occurs:

```R
 REAL() can only be applied to a 'numeric', not a 'integer'
```
In previous versions of igraph, this did not fail. Do you have any idea on what's going wrong here? Thanks in advance!

Below I provide a (stupid) example which already triggers the above stated error. There was no error in this examples prior to igraph 2.0.

### To Reproduce

```R
motif = igraph::make_empty_graph(directed = FALSE) + 
    igraph::vertices("D1", "D2", type = c("type1", "type1")) + 
    igraph::edges("D1", "D2", type = c("type2"))
igraph::subgraph_isomorphisms(target = motif, pattern = motif, method = "vf2", 
                              vertex.color1 = 2, vertex.color2 = 1)
```

krlmlr:
Closes #1218.

@szhorvat: With this, I'm seeing:

``` r
motif <- igraph::make_empty_graph(directed = FALSE) +
  igraph::vertices("D1", "D2", type = c("type1", "type1")) +
  igraph::edges("D1", "D2", type = c("type2"))
igraph::subgraph_isomorphisms(
  target = motif, pattern = motif, method = "vf2",
  vertex.color1 = 2, vertex.color2 = 1
)
#> Error in graph.get.subisomorphisms.vf2(target, pattern, ...): At vendor/cigraph/src/isomorphism/vf2.c:1089 : Invalid vertex color vector length, Invalid value
```

<sup>Created on 2024-02-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Do you expect this error?

### DIFF
--- a/R/topology.R
+++ b/R/topology.R
@@ -110,7 +110,7 @@ graph.get.isomorphisms.vf2 <- function(graph1, graph2, vertex.color1,
     }
   }
   if (!is.null(vertex.color1)) {
-    vertex.color1 <- as.integer(vertex.color1) - 1L
+    vertex.color1 <- as.numeric(vertex.color1) - 1
   }
   if (missing(vertex.color2)) {
     if ("color" %in% vertex_attr_names(graph2)) {
@@ -120,7 +120,7 @@ graph.get.isomorphisms.vf2 <- function(graph1, graph2, vertex.color1,
     }
   }
   if (!is.null(vertex.color2)) {
-    vertex.color2 <- as.integer(vertex.color2) - 1L
+    vertex.color2 <- as.numeric(vertex.color2) - 1
   }
   if (missing(edge.color1)) {
     if ("color" %in% edge_attr_names(graph1)) {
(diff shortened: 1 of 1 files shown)


szhorvat:
Yes, it is expected. The colour vector must contain one entry for each vertex. So if you have two vertices, you need a colour vector of length 2.

krlmlr:
According to your remark in the issue, this fix isn't optimal, we should be using `INTEGER()` instead of `REAL()` here? Would you like to propose a better fix?

szhorvat:
Can I push to this branch?

szhorvat:
I'm sorry, I misspoke. This is fine. Please merge this. I'll submit a separate improvement for the LAD function later. Don't wait for that with the new bugfix release.

PR #1222, opened 2024-02-08
fix: `shortest.paths()` no longer fails

szhorvat:
@krlmlr Should fix most revdep failures.


krlmlr:
Thanks. I think we should `igraph.match.arg(algorithm)` instead. Why do you think this will fix many failures?

szhorvat:
Because locally this fixed this error for me when using `shortest.paths()`:

```
  'arg' must be of length 1
```

This was the majority of the revdep failures.

I fixed the bad docs now, hopefully CI passes

szhorvat:
Please go ahead.

krlmlr:
Fixes 32 out of 33 failures, the 33rd still running. Good catch, I'm so relieved!

PR #1223, opened 2024-02-08
fix: Avoid duplicate objects

krlmlr:
Thanks @dmurdoch!


szhorvat:
Fixed in the C core now, you can just update the C core and regenerate if you like

PR #1225, opened 2024-02-08
test: Add tests for `isomorphisms()` and `subgraph_isomorphisms()`

krlmlr:


tests/testthat/test-topology.R
```diff
@@ -41,3 +41,37 @@ test_that("automorphism_group works with colored graphs", {
...
+
+test_that("isomorphisms works", {
+  motif <- make_empty_graph(directed = FALSE) +
+    vertices("D1", "D2", type = c("type1", "type1")) +
+    edges("D1", "D2", type = c("type2"))
+  subgraph_isomorphisms(
+    target = motif, pattern = motif, method = "vf2",
+    vertex.color1 = 2:1, vertex.color2 = 1:2
+  )
+})
```
szhorvat, line 53: The `type` attribute is unused here. It's not useless to test with `target` and `pattern` the same, but normally they should be different, and `pattern` should be smaller.

For example, let `target` be a `make_ring(10,circular=T)` and `pattern` a `make_ring(5, circular=F)`.

tests/testthat/test-topology.R
```diff
@@ -41,3 +41,37 @@ test_that("automorphism_group works with colored graphs", {
...
+  motif <- make_empty_graph(directed = FALSE) +
+    vertices("D1", "D2", type = c("type1", "type1")) +
+    edges("D1", "D2", type = c("type2"))
+  out <- subgraph_isomorphisms(
+    target = motif, pattern = motif, method = "vf2",
+    vertex.color1 = 2:1, vertex.color2 = 1:2
+  )
+  expect_length(out, 1)
+  expect_equal(as.numeric(out[[1]]), 2:1)
+})
```
szhorvat, line 77: This is near-identical to the first one, can the redundancy be reduced?

PR #1230, opened 2024-02-09
docs: split is_bipartite() manual page from other manual page

### LINKED ISSUE #1202: is_bipartite() has wrong summary in doc tooltip
The `is_bipartite()` doc tooltip in RStudio is misleading. This function does not create a graph, it just checks some attributes. 

image

Can we have this put in a separate doc page @maelle ?

(Mathematically speaking, it doesn't even check if the graph is bipartite! But that's a topic for #709)

maelle:
Fix #1202


szhorvat (approved):
Thanks Maëlle, I missed this earlier. It looks great!

PR #1233, opened 2024-02-11
refactor: always use bundled mini-gmp

szhorvat:
This is an experiment to see if using the bundled mini-gmp triggers any warnings CRAN would complain about. I noticed that we are already compiling `mini-gmp.c` even though we were not using it.

There are currently no disadvantages to using mini-gmp instead of the full-blown GMP, other than the risk of CRAN demanding source changes to fix warnings. We should not change mini-gmp ourselves it's too risky.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -71,4 +71,4 @@ Config/testthat/start-first: vs-es, scan, vs-operators, weakref,
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.3.1
-SystemRequirements: gmp, libxml2, glpk (>= 4.57)
+SystemRequirements: libxml2, glpk (>= 4.57)
--- a/src/Makevars.in
+++ b/src/Makevars.in
@@ -4,12 +4,12 @@ PKG_CFLAGS=$(C_VISIBILITY)
 PKG_CXXFLAGS=$(CXX_VISIBILITY)
 PKG_FFLAGS=$(F_VISIBILITY)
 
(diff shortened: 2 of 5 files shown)


szhorvat:
This seems to trigger no warnings. Hopefully the LTO run wouldn't either.

Since we were already compiling `mini-gmp.c`, it is likely that we were inadvertently using _it_ instead of an external GMP anyway.

Therefore I suggest merging this and using mini-gmp only until igraph gains features for which an external GMP has tangible benefits.

szhorvat:
If this is merged, that's 1 out of 3 done.

krlmlr:
Thanks. We also need to edit `DESCRIPTION` to indicate that GMP is no longer needed.

Checking LTO now.

krlmlr:
No preference, but consistency seems important. What about the other includes that also might be misplaced?

szhorvat:
Will fix those in a separate commit, within this PR. Give me 5 min.

krlmlr:
Why not do this in a completely separate PR instead?

szhorvat:
Because it affects the same file, so the next PR would be dependent on this. Do you want me to remove that commit temporarily?

krlmlr:
Let's continue here, it's fine. If the commits are clean and self-contained, that's good enough.

Antonov548 (approved):

krlmlr:
LTO seems fine her too.

krlmlr:
Would you like to finish the move to `config.h` here and add the "mergequeue" label?

szhorvat:
The move is finished in the sense that we're now consistent with the C core in what goes into `config.h`. Adding the label.

PR #1234, opened 2024-02-11
chore: remove unused patch files

szhorvat:
These were patches applied to the C core. They are no longer used. Let's clean out all the old junk.


krlmlr (commented):
Thanks!

szhorvat:
`src/config.h.in` is also unused. It was part of the old autoconf build system. If we're not going back to autoconf, that should be removed as well (though I still think that sticking with autoconf may be the easiest way forward).

PR #1229, opened 2024-02-09
refactor: access igraph R package version more easily

maelle:
Related to #1208

I don't understand why the code was so complicated, maybe I'm missing a nuance.

PR #1236, opened 2024-02-12
Update build-requirements.txt

maelle:
@Antonov548 does a tag need to be created or would this work?

Context https://github.com/igraph/rigraph/pull/1229#discussion_r1484238950

### DIFF
--- a/tools/build-requirements.txt
+++ b/tools/build-requirements.txt
@@ -3,4 +3,4 @@
 # environment to detect when to re-create the virtualenv so the hash
 # needs to change whenever a package is updated.
 
-git+https://github.com/igraph/stimulus@0.21.4#egg=stimulus
+git+https://github.com/igraph/stimulus@2a273aaa6cf31470c4af30077aece3015c3e4f7f#egg=stimulus
--- a/tools/build-requirements.txt
+++ b/tools/build-requirements.txt
@@ -3,4 +3,4 @@
 # environment to detect when to re-create the virtualenv so the hash
 # needs to change whenever a package is updated.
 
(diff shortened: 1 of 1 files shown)


Antonov548:
I think it might work. 
Previously we were try to keep some version of stimulus. Maybe we can ask @ntamas to create new release. I just don't have permissions in stimulus repository.

maelle:
@ntamas if you have time to create a new tag in the Stimulus repo, that'd be great :pray:

szhorvat:
This works. Let's use this for the moment in the interest of getting the bugfix release out ASAP, as Tamás is travelling at the moment.

maelle:
@szhorvat ok! What will this update exactly? In particular will this update the Docker stuff used in https://github.com/igraph/rigraph/blob/0dabc9ee8d77bb169288a7944f6079e6628d7147/src/README.md?plain=1#L23

szhorvat:
This will use the new version of stimulus when you re-generate interfaces using

```
make -f Makefile-cigraph src/rinterface.c R/aaa-auto.R
```

maelle:
Thanks! I wonder about the Docker stuff but I'll need to explore a bit more (I use Docker rather than make directly because it has worked better for me)

ntamas:
Stimulus 0.21.5 is now tagged so you can use that. Note that you don't need a tag for quick fixes like this as you can point `build-requirements.txt` to a [specific Git commit hash](https://stackoverflow.com/a/35998253/156771) instead.

maelle:
thank you!!

PR #1237, opened 2024-02-12
chore: update C core and regenerate interfaces

szhorvat:
@krlmlr This includes a cache-related bugfix which should be in the next release.


krlmlr (approved):
Thanks!

Issue #742, opened 2023-03-30
add some explanation of aaa-auto.R generation in CONTRIBUTING.md


maelle:
Obviously asking for a friend :sweat_smile:

krlmlr:
Maybe a reference to https://github.com/igraph/rigraph/blob/main/src/README.md, and document there?

I agree the current docs are sparse.

maelle:
some inspiration for the contributing guide https://arrow.apache.org/docs/r/articles/index.html#developer-guides

szhorvat:
That's a great contributor's guide that would of course be nice to have. The bottleneck is the time required to write it.  Another problem is that the infrastructure is in flux at the moment.

I'd suggest drafting a guide in the GitHub wiki. If you figure out something that wasn't properly documented, feel free to write it down there. Once we have something solid, it can be migrated to a public location like `CONTRIBUTORS.md`

maelle:
current status of my using the generation #850

krlmlr:
I use:

```sh
docker run --rm -ti --platform linux/amd64 -v $(pwd):/root/workspace ghcr.io/cynkra/rig-ubuntu-igraph:main make -f Makefile-cigraph
```

for full reproducibility. The only noisy thing is lex and yacc, we could add version checks here too.

maelle:
what should pwd be?

maelle:
I am not sure I understand how one would actually run the command.

szhorvat:
> The only noisy things are lex and yacc

Let's call these `flex` and `bison`. We do use extensions provided by GNU Flex and GNU Bison.

Not sure what version you're using now, but I strongly recommend the latest Bison, namely 3.8.2. Bison 3.8 generates better end-user error messages (i.e. messages that igraph users will see) than earlier versions.

As for Flex, it hasn't had a release in ages. Any platform we might ever use should have Flex 2.6.4.

krlmlr:
Run

```sh
docker run --rm -ti --platform linux/amd64 -v $(pwd):/root/workspace ghcr.io/cynkra/rig-ubuntu-igraph:main make -f Makefile-cigraph
```

from a local clone of this repository. With this, `$(pwd)` expands to the right location.

Just checked, we have the right versions in our Docker image:

```
docker run --rm -ti --platform linux/amd64 -v $(pwd):/root/workspace ghcr.io/cynkra/rig-ubuntu-igraph:main bison --version
# bison (GNU Bison) 3.8.2
[...]

PR #1238, opened 2024-02-12
refactor: create helper function for code regeneration

maelle:
Related to #742

Questions

- what other files can be regenerated
- should there be a special value of file for regeneration all of them (NULL)

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -49,3 +49,4 @@
 ^\.ccache$
 ^\.aviator/config\.yml$
 ^src/vendor/cigraph/CITATION\.cff$
+^R/stimulus\.R$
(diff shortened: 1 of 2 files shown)


krlmlr:
This is showing weird

```
fatal: detected dubious ownership in repository at '/root/workspace'
```

errors when run.

Why do we need an R wrapper for a shell call?

We can work together on cleaning up the process and the documentation.

maelle:
> Why do we need an R wrapper for a shell call?

I thought it'd be smoother (since I am mostly working in the R console) but if it's not possible, I'll keep using the shell call.

krlmlr:
Need to look at the error messages together.

krlmlr:
We no longer run `make` in Docker, instead the `Makefile` contains `docker run` instructions.

PR #1252, opened 2024-02-15
docs: Fix error messages mentioning to mention `upgrade_graph()` instead of the nonexisting `upgrade_version()`

louisaslett:
I tried to work with some old saved graphs and encountered the error message:

```
Error in warn_version(graph) : 
    This graph was created by a now unsupported old igraph version.
    Call upgrade_version() before using igraph functions on that object.
```

However, there is no `upgrade_version()` function exported by the `igraph` package, but there is an `upgrade_graph()` function which correctly upgrades the old objects.  This pull requests just modifies the incorrect error message (and the test results which expect the error output).


krlmlr:
Thanks!

krlmlr:
Good catch!

PR #1254, opened 2024-02-17
refactor: auto-generate `are_adjacent()`, avoid deprecated `igraph_are_connected()` C function

### LINKED ISSUE #1253: `igraph_are_adjacent()`
https://github.com/igraph/igraph/pull/2486

szhorvat:
Fixes #1253

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -533,6 +533,25 @@ sample_dirichlet_impl <- function(n, alpha) {
   res
 }
 
+are_adjacent_impl <- function(graph, v1, v2) {
+  # Argument checks
+  ensure_igraph(graph)
+  v1 <- as_igraph_vs(graph, v1)
+  if (length(v1) == 0) {
+    stop("No vertex was specified")
+  }
+  v2 <- as_igraph_vs(graph, v2)
+  if (length(v2) == 0) {
+    stop("No vertex was specified")
+  }
+
+  on.exit( .Call(R_igraph_finalizer) )
+  # Function call
+  res <- .Call(R_igraph_are_adjacent, graph, v1-1, v2-1)
+
+  res
+}
+
 distances_impl <- function(graph, from=V(graph), to=V(graph), mode=c("out", "in", "all", "total")) {
   # Argument checks
   ensure_igraph(graph)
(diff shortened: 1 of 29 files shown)


krlmlr:
Thanks. Seeing `-Waddress` warnings in https://github.com/igraph/rigraph/actions/runs/7944281043/job/21689682420?pr=1254#step:9:84:

```
  vendor/cigraph/src/community/leading_eigenvector.c:606:28: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
  vendor/cigraph/src/graph/visitors.c:261:21: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
  vendor/cigraph/src/graph/visitors.c:551:67: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
  vendor/cigraph/src/isomorphism/vf2.c:667:71: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
  vendor/cigraph/src/isomorphism/vf2.c:1504:71: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
  vendor/cigraph/src/misc/motifs.c:396:25: warning: the address of 'ret' will always evaluate as 'true' [-Waddress]
```

Could be a change in the build envionment, or related to this PR. New run: https://github.com/igraph/rigraph/actions/runs/7946903820

szhorvat:
Forgot to re-add a manual patch after auto-generation.

szhorvat:
Fixed.

krlmlr:
Do we need two macros instead of one?

https://github.com/igraph/rigraph/compare/4b5fd0554f47303c08857734be444af498d1341c..2211c0f35c3642196c0e557892afc0d6e2f12e15

szhorvat:
CC @ntamas since he added this macro originally.

I vote for changing the macro in the C core in the same way as above, as this warning has been a constant annoyance when testing with various compilers, and so far there has been no need for performing this NULL-check. But it should be decided by Tamás.

krlmlr:
I suspect it could be replaced with a static assert?

ntamas:
That macro was the usual defensive programming habit from my side. If we can replace the check with a static assert, that's great. If we can't, let's add a runtime assert instead if possible.

PR #1256, opened 2024-02-19
docs: GMP is no longer a dependency

szhorvat:


krlmlr:
Thanks. Removing the "blocked" label after a push to main does this. Let's see what happens.

Issue #1259, opened 2024-02-20
C/igraph version is not set before compiling


szhorvat:
### What happens?

The C/igraph version is not set before compiling. See `igraph_version.h`, which contains `"@PACKAGE_VERSION@"`. This then leaks through in some contexts, for example when writing GML files.

### To Reproduce

```r
write_graph(make_graph(c(1,2)), "foo.gml", format="gml")
```

The resulting file will contain:

```
Creator "igraph version @PACKAGE_VERSION@ Tue Feb 20 11:20:57 2024"
```

maelle:
@Antonov548 could you please have a look?

Antonov548:
> @Antonov548 could you please have a look?

Yes, I will

szhorvat:
You might find that fixing this is not simple, as the version is normally extracted from git tags. The script that updates the C core could be changed to write commit hash into the version field, or similar ... but that's maybe too much trouble?

For writing GMP specifically, it is possible to pass your own Creator string instead of using the default. You could pass in the R/igraph version, or just a fixed string, whatever you like.

For getting the `igraph_version()` C function working properly: this is not really critical.

Is there anything else affected? I'm not sure. Maybe not.

Issue #1261, opened 2024-02-20
Improve `Makefile-cigraph`


krlmlr:
- [ ] Update all redundant files using just `make -f Makefile-cigraph`
    
    Some components (flex, bison) in the `Makefile` need specific versions of the tools, which is why we advise to run `make` in Docker. But we actually don't need to run the entire `Makefile`, running only the flex and bison parts in Docker is sufficient.

- [ ] More precise rules

    Running `make -f Makefile-cigraph` should "just work" in more cases, even without the `-B` switch. These days, often it seems we need `-B` because changes are not detected

- [ ] Include other tooling into the Makefile

    Example: updating the vendored sources

- [ ] Documentation

PR #1262, opened 2024-02-21
Extract version in `igraph_version.h` during vendoring

Antonov548:
Closes https://github.com/igraph/rigraph/issues/1259

A bit more compicated process of vendoring cigraph. But it extracts exact version from cigraph using cmake.
Since this only for dev team I think it should be fine.


szhorvat:
I'm really not the right person to review this. Perhaps Tamás or Kirill should take a look instead.

A small request from me if possible: Let's keep the `--depth` large enough but still finite so that the amount downloaded is kept in check. This is useful when working with bad connections (on a train, in China, etc.)

Antonov548:
> I'm really not the right person to review this. Perhaps Tamás or Kirill should take a look instead.
> 
> A small request from me if possible: Let's keep the `--depth` large enough but still finite so that the amount downloaded is kept in check. This is useful when working with bad connections (on a train, in China, etc.)

Good suggestion. I just not sure what will be enough to extract version information?

szhorvat (approved):
No objection from me, but again, I'm not the best person to review this.

Please don't let my `--depth` request become a blocker for merging this.

szhorvat:
Would it be possible to break this up into several stages: first update the files, then commit. There would be multiple make targets, with the option to pull a certain branch of the C core and NOT commit the changes.

This would make it easier to test changes to the C core before merging them. I did this yesterday, but I did it manually: I edited the script to pull a certain branch, I let it commit the changes, and finally I undid the last commit.

All of this would be the topic of a future PR, of course.

Antonov548:
> Would it be possible to break this up into several stages: first update the files, then commit. There would be multiple make targets, with the option to pull a certain branch of the C core and NOT commit the changes.
> 
> This would make it easier to test changes to the C core before merging them. I did this yesterday, but I did it manually: I edited the script to pull a certain branch, I let it commit the changes, and finally I undid the last commit.
> 
> All of this would be the topic of a future PR, of course.

I think it could be part of https://github.com/igraph/rigraph/issues/1261

krlmlr:
The vendoring code for some reason kills the `.mk` files.

I'll install a GitHub Action that periodically updates the vendored C core, similarly to the duckdb R package.

krlmlr:
Moved the important bits into the other PR.

I wonder if we also should run `cmake` in our `./configure` to generate the relevant headers, without using the full-fledged build environment generated by `cmake` . That would be very nice and indeed simplify our logic in `./configure` .

PR #1264, opened 2024-02-25
docs: Improve `min_st_separators()` documentation

szhorvat:
@maelle Is this the best way to comment on examples? Can we comment using normal text that includes math instead of writing code comments? If yes, can you push to this branch and demonstrate?

### DIFF
--- a/R/flow.R
+++ b/R/flow.R
@@ -812,12 +812,14 @@ dominator_tree <- function(graph, root, mode = c("out", "in", "all", "total")) {
 #'
 #' chvatal <- make_graph("chvatal")
 #' min_st_separators(chvatal)
-#'
-#' # Note that {1, 3} is returned despite its subset {1} being a separator
-#' # as well. This is because {1, 3} is minimal with respect to separating
-#' # vertices 2 and 4.
+#' @section Note:
+#' Note that \eqn{1, 3} is returned despite its subset \eqn{1} being a separator
+#' as well. This is because \eqn{1, 3} is minimal with respect to separating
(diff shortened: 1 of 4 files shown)


maelle:
Good question!

We can't have maths in example but a workaround would be to create an arbitrary section using R Markdown support. The code in the code chunk is evaluated every time the `document()` function is run,
- which seems fine as far as time is concerned as this is fast code apparently;
- which made me add the same code as a test case.

https://roxygen2.r-lib.org/articles/rd-formatting.html#code-chunks

Note that the section name is currently really bad. :grin:

szhorvat:
Thank you!

szhorvat:
A problem is that every time we re-generate docs, something changes:

image

Should we do something about this?

Perhaps we should just go with the original code comment version? The math is actually not that important for this specific example.

krlmlr:
In the tests, we have `local_igraph_options(print.id = FALSE)` to inhibit this, but I'm not sure what's the best way to use this with a roxygen2 example.

@maelle: Can you help please?

maelle:
Probably not the best solution as this is an unexported function :zany_face: (so especially confusing for users) I need to make it work to put the call in a separate chunk (using the `.in` argument), will try again later this week.

szhorvat:
Let's just add the comments as an actual code comment, problem solved. Agreed?

I would rather not add distracting code to an example which is making an important  point about how a function behaves.

maelle:
I'll experiment with a knitr hook to modify the output without a comment. I fear the comment is distracting too.

PR #1268, opened 2024-02-27
docs: rm duplicate seealso from sample_pa.Rd

maelle:


szhorvat:
Can you explain what exactly went wrong here, to make sure the rest of us won't cause this kind of problem in the future? What do I need to pay attention to?

maelle:
Sure! If a topic is documented on the same page as another one using `@rdname`, one should not repeat the `@family` tag from that main topic, otherwise that duplicates the Seealso sections. 

I opened an issue in roxygen2 a while back, but I'm pretty sure it's an user error (=my mistake). https://github.com/r-lib/roxygen2/issues/1530

PR #1276, opened 2024-03-02
chore: Model vendoring after duckdb

### LINKED ISSUE #1259: C/igraph version is not set before compiling

krlmlr:
~~Includes an update of the vendored sources.~~

Closes #1262 (=supersedes it).

Fixes #1259.


krlmlr:
@szhorvat @maelle @Antonov548: One step towards automating the vendoring process. Needs https://github.com/igraph/igraph/pull/2527 for full automation.

We now only need `make -f Makefile-cigraph`, the Docker call is embedded into the Makefile. Tracking the loose ends in #1261.

PR #1278, opened 2024-03-02
chore: Fix drift after code generation

krlmlr:


szhorvat:
Would it be possible to not make Docker mandatory for local development? I promise not to commit anything that didn't go through the standard process, but it'd be good to not enforce such a heavy dependency ..

krlmlr:
One option would be to change the core to include the generated files. I see that this only shifts this problem, but then it becomes Somebody Else's. Docker is my way of dealing with it.

How can we implement it in a way that the default would be to run with Docker, but that the targets that require Docker can be omitted if needed?

szhorvat:
In this setup, if the `DOCKER` variable is empty, we're back to the original setup. Is this correct?

Perhaps an environment variable could control if `DOCKER` is set, so `SKIP_DOCKER=1 make -f Makefile-cigraph` would not use Docker?

krlmlr:
I'd prefer `IGRAPH_SKIP_DOCKER`, but yes, that would do it.

Or, my preferred way, not run `flex` and `bison` here at all, to avoid dealing with this in Python and Mathematica, too. Remember, the issue is the drift that occurs from different versions of these tools, which is hard to control. If there was a code owner for the files, they would be responsible for updating the generated code in Git, and could do this differently.

PR #1279, opened 2024-03-04
docs: add link to discussion forum

maelle:
@szhorvat which version do you prefer, link above under links or in a separate section? In the screenshot below I am showing both options at once.

image


szhorvat:
Just under "Links"

PR #1280, opened 2024-03-04
docs: add logo

maelle:


szhorvat:
Thanks @maelle ! Would it be possible to prettify a bit?

Now it's like this (unbalanced, overlaps with line below):

image

This looks better to me (balanced with text size):

image

Do you agree?

maelle:
Right, it works better with hex logos like https://pkgdown.r-lib.org/ -- is there a hex version of the igraph logo?

Issue #1217, opened 2024-02-08
Improve libxml2 detection


krlmlr:
https://github.com/igraph/rigraph/pull/1197#issuecomment-1928275316

- [x] Rename `xml2_include_paths` variable to `xml2_cflags`
- [x] Also use linker flags by `xml2-config` or `pkg-config` instead of hard-coding them

szhorvat:
I checked both boxes here, as they're done. But there's one more thing: if libxml2 is not found, it'd be nice to just disable GraphML support instead of failing to build.

maelle:
> just disable GraphML support

What should then happen if the user tries using the related functionality?

I wonder whether other R packages also have this sort of "partial installation".

szhorvat:
There will be a nice and clear error message explaining that this functionality is not available. This is all already implemented, so nothing needs to be done in R code. The C sources simply need to be compiled with the appropriate macros (un)defined.

PR #1282, opened 2024-03-05
chore: make libxml optional

Antonov548:
Closes https://github.com/igraph/rigraph/issues/1217

### DIFF
--- a/configure
+++ b/configure
@@ -8,7 +8,7 @@ elif pkg-config --version >/dev/null 2>&1; then
     libxml2_cflags=$(pkg-config --cflags libxml-2.0)
     libxml2_libs=$(pkg-config --libs --static libxml-2.0)
 else
-    echo "Warning: libxml not found. Neither xml2-config nor pkg-config is available"
+    echo "Warning: libxml2 not found. Neither xml2-config nor pkg-config is available."
 fi
 
 if [ -n "$libxml2_cflags" ]; then
--- a/configure
+++ b/configure
@@ -22,7 +22,7 @@ if [ -n "$libxml2_libs" ]; then
     PKG_LIBS="$PKG_LIBS $libxml2_libs"
     echo "libxml2 library link flags: $libxml2_libs"
 else
-    echo "Warning: libxml2 library link flags not found"
+    echo "Warning: libxml2 library link flags not found."
 fi
 
 OBJECTS='${SOURCES} ${MINIGMPSOURCES}'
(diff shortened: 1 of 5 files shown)


krlmlr (approved):
Thanks! Can you please update the `SystemRequirements` section in `DESCRIPTION` too, also for GLPK?

configure
```diff
@@ -8,24 +8,21 @@ elif pkg-config --version >/dev/null 2>&1; then
     libxml2_cflags=$(pkg-config --cflags libxml-2.0)
     libxml2_libs=$(pkg-config --libs --static libxml-2.0)
 else
-    echo "Error: Neither xml2-config nor pkg-config is available"
-    exit 1
+    echo "Warning: libxml not found. Neither xml2-config nor pkg-config is available"
```
szhorvat, line 11:
```suggestion
    echo "Warning: libxml2 not found. Neither xml2-config nor pkg-config is available."
```

configure
```diff
@@ -8,24 +8,21 @@ elif pkg-config --version >/dev/null 2>&1; then
...
 fi
 
 if [ -n "$libxml2_cflags" ]; then
-    PKG_CFLAGS="$PKG_CFLAGS $libxml2_cflags"
+    PKG_CFLAGS="$PKG_CFLAGS $libxml2_cflags -DHAVE_LIBXML"
     echo "libxml2 include directories: $libxml2_cflags"
 else
-    echo "Error: libxml2 include directory not found"
-    exit 1
+    echo "Warning: libxml2 include directory not found"
```
szhorvat, line 18:
```suggestion
    echo "Warning: libxml2 include directory not found."
```

configure
```diff
@@ -8,24 +8,21 @@ elif pkg-config --version >/dev/null 2>&1; then
...
 fi
 
 if [ -n "$libxml2_libs" ]; then
     PKG_LIBS="$PKG_LIBS $libxml2_libs"
     echo "libxml2 library link flags: $libxml2_libs"
 else
-    echo "Error: libxml2 library link flags not found"
-    exit 1
+    echo "Warning: libxml2 library link flags not found"
```
szhorvat, line 25:
```suggestion
    echo "Warning: libxml2 library link flags not found."
```

szhorvat (approved):
If libxml2 detection fails at either of the three steps, could we print a message saying something like "GraphML import will not be available." ?

szhorvat:
Also, it would be good if someone could test this on a system where libxml2 is not available: make sure it compiles, links, and trying to import a GraphML file gives a nice error. I don't see any reason why it wouldn't work, but just in case ... it's hard for me to hide libxml2 since it comes as standard on macOS.

szhorvat (approved):

PR #1287, opened 2024-03-09
refactor: names in `update-cigraph.sh` script

Antonov548:


krlmlr (commented):
Thanks. If we must rename, can we please use `upstream_dir` ? I'd like this script to be meaningfully diffable with what's in the duckdb-r repo.

krlmlr:
Thanks!

Issue #1195, opened 2024-02-04
Run doc examples with sanitizer


szhorvat:
Can we run the doc examples with sanitizer checks? It has the potential to reveal invalid memory accesses and memory leaks. In fact, there are such memory leaks at the moment, see https://github.com/igraph/rigraph/pull/1173#issuecomment-1925481007

Antonov548:
@szhorvat I have reviewed types conversion. Seems you covered all wrong conversions. At least I didn't find with manual search other types that have memory leaks. I think we can try to add examples to CI. 

I just not sure what is right way to run only examples for locally installed igraph.

@maelle May you suggest some solution for running examples? 
```R
library(igraph)
functions_list <- ls("package:igraph")
for (func in functions_list) {
  example(func)
}
```

Or can we just use `R CMD check` which will run only examples?

szhorvat:
@Antonov548 Something I wasn't sure about is cases like this:

https://github.com/igraph/rigraph/blob/7c0540e46006840f73fdf4b4ff3da83e58f42ffa/tools/stimulus/types-RC.yaml#L154-L173

Does the `INOUT` part need its own `OUTCONV`?

Antonov548:
> INOUT

This is I also get confusen. But seems stimulus use `OUT` conversion, even if it's `INOUT` and `INOUT` for `OUTCONV` is not exists. For example `igraph_layout_drl` which has `INOUT MATRIX res`, and `MATRIX` hasn't `OUTCONV` for `INOUT`.

So, it should works fine.

maelle:
@Antonov548 sorry seeing this only now. `devtools::run_examples(run_donttest = TRUE, run_dontrun = TRUE, fresh = TRUE)`!

Antonov548:
> @Antonov548 sorry seeing this only now. `devtools::run_examples(run_donttest = TRUE, run_dontrun = TRUE, fresh = TRUE)`!

I see the ` *** caught segfault ***` when trying to run examples with this command. I was trying to find which exactly example produce it, but without success.

Do you also have this issue?

szhorvat:
Do you get at least an ASan stack trace?

Antonov548:
> Do you get at least an ASan stack trace?

I didn't try with ASan yet. Need to setup it locally after refresh of my pc.

krlmlr:
Need to find a good way to run all examples from the installed package for ASAN.

krlmlr:
[...]

PR #1286, opened 2024-03-09
ci: is sanitizer actually working?

Antonov548:

PR #1288, opened 2024-03-10
ci: run examples with sanitizer

Antonov548:
Closes: https://github.com/igraph/rigraph/issues/1195

Proof that it detects memory leaks in examples: https://github.com/igraph/rigraph/pull/1286

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -251,3 +251,4 @@ jobs:
             cd rigraph
             RDcsan CMD INSTALL . --no-byte-compile
             TESTTHAT_PARALLEL=false CI=true RDcsan -q -e 'testthat::test_local(reporter = c("location", "summary"), load_package = "installed")'
+            RDcsan -q -e 'source("tools/run_examples.R")'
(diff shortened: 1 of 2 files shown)


szhorvat:
So does this mean that we're good, no crashes or leaks triggered by examples?

Antonov548:
> So does this mean that we're good, no crashes or leaks triggered by examples?

Sanitizer showed some warnings about undefined behavior (or similar), but it doesn't cause the crash.

Yes, examples good.

szhorvat:
> Sanitizer showed some warnings about undefined behavior

That may still indicate a real issue, though not with certainty. Would you be able to extract the full message? I can only see a partial thing here:

https://github.com/igraph/rigraph/actions/runs/8221181042/job/22481341484#step:5:6601

We can then fine-tune the specific checks UBSan does so that it only warns about real issues.

szhorvat:
Never mind, I looked at line 80. We cast a real to integer and back to see if the real is representable as an integer. We do this on purpose, not an issue.

krlmlr (approved):
Nice, thanks!

maelle:
@Antonov548 a limitation to `R2dex()` that I just realized is that it does not run the examples that are wrapped in `\dontrun{}`. I'll make a PR.

Issue #876, opened 2023-07-20
sample_degseq() method parameter values should be renamed to match C core


szhorvat:
The allowed value for `sample_degseq`'s `method` parameter should be renamed to match the 0.10 C core. Old names can be kept for compatibility. The renaming can be done now, even before the 0.10 upgrade.

See the C docs:

https://igraph.org/c/html/latest/igraph-Generators.html#igraph_degree_sequence_game

The R docs should be updated. Feel free to copy and paste from the C docs.

Renaming should be done like so:

 - `simple` -> `configuration`
 - `simple.no.multiple` -> `fast.heur.simple`
 - `simple.no.multiple.uniform` -> `configuration.simple`
 - `vl` does not need renaming
 - when upgrading to 0.10, there will be a new method called `edge.switching.simple`.

The new names are rationalized. When they contain the word "simple", it means that they produce simple graphs (not multigraphs).

If someone will do the renaming, and deprecates the old names, I can do the doc update.

szhorvat:
@krlmlr This would be very good to fix before a 2.0.

krlmlr:
It's used by seven packages: https://github.com/search?q=org%3Acran%20sample_degseq&type=code .

More than zero, but also more like a "nice to have" in my book.

szhorvat:
These names are quite misleading and have been changed a long time ago in other interfaces. I plan to reuse names like "simple" in the C core in the near future. The sooner this is fixed the better.

krlmlr:
I don't disagree, it's just not necessary for 2.0.0. We can rename parameters without breaking other packages.

szhorvat:
I'd like to ask to give this some priority, and to expose `edge.switching.simple` soon.

PR #1297, opened 2024-03-12
docs: fix `sample_degseq()` example

### LINKED ISSUE #1289: error
Related to #1286 (@Antonov548) I extracted all examples in an Rmd file to knit it and see what fails.

I get

```r
Error in `sample_degseq()`:
! At vendor/cigraph/src/games/degree_sequence_vl/gengraph_mr-connected.cpp:150 : Cannot realize the given degree sequence as an undirected, simple graph. Invalid value
Backtrace:
 1. igraph::sample_degseq(degs, method = "vl")
```

for 

```r
### Name: sample_degseq
### Title: Generate random graphs with a given degree sequence
### Aliases: sample_degseq degseq
### Keywords: graphs

### ** Examples


## The simple generator
g <- sample_degseq(rep(2, 100))
degree(g)
is_simple(g) # sometimes TRUE, but can be FALSE
g2 <- sample_degseq(1:10, 10:1)
degree(g2, mode = "out")
degree(g2, mode = "in")

## The vl generator
g3 <- sample_degseq(rep(2, 100), method = "vl")
degree(g3)
is_simple(g3) # always TRUE

## Exponential degree distribution
## Note, that we correct the degree sequence if its sum is odd
degs <- sample(1:100, 100, replace = TRUE, prob = exp(-0.5 * (1:100)))
if (sum(degs) %% 2 != 0) {
  degs[1] <- degs[1] + 1
}
g4 <- sample_degseq(degs, method = "vl")
all(degree(g4) == degs)

## Power-law degree distribution
## Note, that we correct the degree sequence if its sum is odd
degs <- sample(1:100, 100, replace = TRUE, prob = (1:100)^-2)
if (sum(degs) %% 2 != 0) {
  degs[1] <- degs[1] + 1
}
g5 <- sample_degseq(degs, method = "vl")
all(degree(g5) == degs)



```

but running this code directly, or with reprex... does NOT lead to any error. Maybe due to some randomness?

maelle:
fix #1289 

- renamed variables, added explaining variables for clarity.
- conditionally use withr, to make sure it works.

note that `rlang::check_installed()` will prompt the user to install withr, which is handy IMHO.

### DIFF
--- a/R/games.R
+++ b/R/games.R
@@ -802,35 +802,39 @@ random.graph.game <- erdos.renyi.game
 #' @examples
 #'
 #' ## The simple generator
-#' g <- sample_degseq(rep(2, 100))
-#' degree(g)
-#' is_simple(g) # sometimes TRUE, but can be FALSE
-#' g2 <- sample_degseq(1:10, 10:1)
-#' degree(g2, mode = "out")
-#' degree(g2, mode = "in")
+#' undirected_graph <- sample_degseq(rep(2, 100))
+#' degree(undirected_graph)
+#' is_simple(undirected_graph) # sometimes TRUE, but can be FALSE
+#'
+#'
+#' directed_graph <- sample_degseq(1:10, 10:1)
+#' degree(directed_graph, mode = "out")
+#' degree(directed_graph, mode = "in")
 #'
 #' ## The vl generator
-#' g3 <- sample_degseq(rep(2, 100), method = "vl")
-#' degree(g3)
-#' is_simple(g3) # always TRUE
+#' vl_graph <- sample_degseq(rep(2, 100), method = "vl")
+#' degree(vl_graph)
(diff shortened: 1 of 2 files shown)


szhorvat:
How about just doing something like the following and avoiding `withr`?

```
while (! is_graphical(degs <- sample(1:100, 100, replace = TRUE, prob = (1:100)^-2)) ) {}
```

This keeps sampling until we hit on a graphical degree sequence (which happens with a high probability in this case).

I am not really in favour of the name `vl_graph` ... The Viger-Latapy method creates a simple and connected undirected graph, but we probably don't want to use a variable that spells all that out. I'd rather stick to `g1`, `g2`, `g3`, etc. and spell out their properties later.

No other comments except that I really hate the current method names and I am hoping that https://github.com/igraph/rigraph/issues/876 will be fixed soon ...

PR #1299, opened 2024-03-12
chore: Update vendored sources to igraph/igraph@0f383457a13864b86a29eed4faf4efaade4c1254

### LINKED ISSUE #1296: Fix builds with the develop branch of igraph/C
Switch the igraph/C repo to the `develop` branch, run `tools/update-cigraph.sh` . This leads to compilation errors that need to be fixed.

Can we make this our priority?

krlmlr:
Update to the develop branch of igraph/C. @Antonov548: Can you please pick up from here?

Closes #1296.


Antonov548:
@krlmlr I couldn't get from where is version of igraph/C which is used in this pull request and why there so much difference with the current. For me it seems it's some old version.

Antonov548:
Was this line of code added on purpose?
https://github.com/igraph/rigraph/blob/6eeb6e4a575a8c0f718f0606cfc8805932b9e604/tools/update-cigraph.sh#L58

krlmlr:
Have you seen #1296?

See https://github.com/igraph/rigraph/blame/6eeb6e4a575a8c0f718f0606cfc8805932b9e604/tools/update-cigraph.sh#L58 for where the line in question was added. True, this should have been a comment in the code.

Antonov548:
@szhorvat @ntamas 
This is what I added manually to the C/igraph. There is missmutch of definition and declaration:

https://github.com/igraph/rigraph/blob/90bb9e9492cd5d2997ce53633f16325359da02fd/src/vendor/cigraph/src/isomorphism/bliss.cc#L261-L264

ntamas:
Thanks, fixed in 44541b9ef

Antonov548:
> Thanks. So, now we can compile, but tests fail? Can we identify which, and then perhaps create a checklist?

Some tests are failing with stop of execution. Can you remind please how I can run without stop of execution? To see all failing tests.

krlmlr:
```sh
TESTTHAT_PARALLEL=false R -q -e 'testthat::test_local(reporter = "location")
```

Antonov548:
Teasts are passing 👍 
Examples are failing 👎

szhorvat:
To make this work well, I suggest basing the upgrade on the changelog, not on the test. Reading the changelog is a must to be able to adapt not just functions, but also their documentation, appropriately.

szhorvat:
In the meantime some more breaking changes were made on the C side.

src/rinterface_extra.c
```diff
@@ -2299,9 +2299,9 @@ igraph_error_t R_igraph_interrupt_handler(void *data) {
    */
   if (R_ToplevelExec(checkInterruptFn, NULL) == FALSE) {
     IGRAPH_FINALLY_FREE();
-    return IGRAPH_INTERRUPTED;
+    return 1;
```
szhorvat, line 2302: It's a good idea to use `true` / `false` instead of `1` and `0`. That way it's immediately clear that what's returned is a boolean, not an integer. The same goes for `NULL` vs `0`.

krlmlr:
@Antonov548: can you please refresh this PR, to see where we'll be at when upgrading the C core?

krlmlr:
Keeping this open as a reminder to bring in the develop branch of igraph/C.

krlmlr:
On the next branch now, link to this discussion posted to the relevant PR.

PR #1303, opened 2024-03-12
feat!: Stop accepting setting `layout` as a graph attribute if it is a matrix

### LINKED ISSUE #775: Disturbed layout in subgraph
- Consider the code below.
```
## bug / undesired / unexpected behaviour.
g        <- make_ring(10)
g$layout <- layout_in_circle(g)
sg       <- subgraph(g, c(1,2))
## delete_graph_attr(sg, "layout")
plot(sg)
```
Note that vertices 1 and 2 are plotted multiple times.
This is because the layout vector of the original graph is copied to the sub graph.

**Version information**
 - R/igraph version: [1] igraph_1.4.1

subgraph-layout-bug

krlmlr:
Closes #775.

@maelle: Can you please pick up from here?

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -337,12 +337,14 @@ set_graph_attr <- function(graph, name, value) {
   ensure_igraph(graph)
 
   # Code that accesses g$layout can stay for now, revisit in 2029.
+  # https://github.com/igraph/rigraph/issues/775
   if (name == "layout" && is.matrix(value)) {
     lifecycle::deprecate_stop(
-      "2.0.3",
+      "2.1.0",
       "set_graph_attr(layout = 'matrix(...)')",
       "set_vertex_attr(layout = )",
-      details = "Using a matrix for the `layout` attribute is deprecated. Set the vertex attribute `layout` instead."
+      details = "Using a matrix for the `layout` attribute is deprecated.
+      Set the vertex attribute `layout` instead."
     )
     return(set_vertex_attr(graph, name, value))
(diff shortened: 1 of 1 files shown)


szhorvat:
Then there will need to be a convenient way to enter the layout information as vertex attributes (`x` and `y`, or `x`, `y` and `z`).

I'm not exactly sure what features exist for this.

There's e.g. `add_layout_()` which actually adds a _graph_ attribute ...

krlmlr:
Accept a vertex attribute `layout` that is decomposed into `x`, `y` and `z`, and return a `layout` attribute when asked and if `x` and `y` exist?

szhorvat:
Something of that sort, yes.

The problem is that the function that compute layouts return matrices.

But in order to store layouts in graphs properly, we need to split these matrices into x, y and maybe z components.

For convenience, there should be some easy way to convert between the two representations. In particular, there should be a very easy way to compute a layout (produced as a matrix) and store it in the graph as x, y vertex attributes. Perhaps the `add_layout_()` function should be updated to do precisely this. This is a "mildly breaking" change.

(If you ask me, the best solution would be if coordinate tuples could be stored in a single "position" vertex attribute, but unfortunately the C core inly supports scalars at the moment, not vectors/tuples, as attribute values associated with individual vertices. I hope we get the new grant from the CZI and can work on improving this.)

krlmlr:
@szhorvat: Something like 20dc1e4?

PR #1306, opened 2024-03-13
test: Add skip condition for `rglplot` test

mfansler:
The `rgl` package is a `Suggests` dependency. As such, testing of functions depending on `rgl` should skip if it is not installed. This PR adds such a skip condition.

Test failures due to lack of skip condition were encountered during [Conda Forge builds of v2.0.3](https://github.com/conda-forge/r-igraph-feedstock/pull/59).


krlmlr:
Thanks!

PR #1309, opened 2024-03-19
ci: Update lock.yaml

maelle:
- There were two warnings on https://github.com/igraph/rigraph/actions/runs/8335586645
- The docs of the action show V5, the permissions field. https://github.com/dessant/lock-threads?tab=readme-ov-file#examples


krlmlr (approved):
Thanks!

PR #1322, opened 2024-03-26
refactor: add assert_named_list() helper to assert the value is a named list with no duplicate names

### LINKED ISSUE #1321: refactor value assertion in R/attributes.R
something like this

```r
 # if (!is.list(value) || (length(value) > 0 && is.null(names(value))) ||
 #    any(names(value) == "") || any(duplicated(names(value)))) 
   

not_list <- 1:10

unnamed_list <- as.list(1:10)

empty_name <- rlang::set_names(unnamed_list, c(as.character(1:9), ""))

dups <- rlang::set_names(unnamed_list, rep("bla", 10))

assert_named_list <- function(value) {
  if (!is.list(value) || !rlang::is_named(value) || anyDuplicated(names(value)) > 0) {
    rlang::abort("{.arg value} must be a named list with unique names")
  }
}
 
```

maelle:
Fix #1321

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -349,10 +349,7 @@ graph.attributes <- function(graph) {
 #' @export
 "graph.attributes<-" <- function(graph, value) {
   ensure_igraph(graph)
-  if (!is.list(value) || (length(value) > 0 && is.null(names(value))) ||
-    any(names(value) == "") || any(duplicated(names(value)))) {
-    stop("Value must be a named list with unique names")
-  }
+  assert_named_list(value)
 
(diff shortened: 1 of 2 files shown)


R/attributes.R
```diff
@@ -349,10 +349,7 @@ graph.attributes <- function(graph) {
 #' @export
 "graph.attributes<-" <- function(graph, value) {
   ensure_igraph(graph)
-  if (!is.list(value) || (length(value) > 0 && is.null(names(value))) ||
```
maelle, line 352: I don't see why value would have no length, but I added the case.

maelle: well it does happen according to what happened in tests before I accounted for the case

PR #1330, opened 2024-03-29
refactor: use 'better' logic for updating vertex attributes

### LINKED ISSUE #1327: refactor puzzling loop in attributes.R
https://github.com/igraph/rigraph/blob/36ad3c905d6b9dd34fe50b6b2f369ed090ba3c8b/R/attributes.R#L557-L564

maelle:
Fix #1327


R/attributes.R
```diff
@@ -539,28 +539,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
     index <- as_igraph_vs(graph, index)
 
-    if (any(duplicated(index)) || any(is.na(index))) {
+    if ((anyDuplicated(index) > 0) || (anyNA(index) > 0)) {
       stop("Invalid vertices in index")
     }
   }
 
+  index_number_not_vcount <- (length(index) != vcount(graph))
+  index_not_vertices_indices <- (!all(index %in% V(graph)))
```
maelle, line 555: previously it did not use `%in%` because that part was not executed if index didn't have `vcount(graph)` as length.

R/attributes.R
```diff
@@ -539,28 +539,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
   if (!missing(index) &&
-    (length(index) != vcount(graph) || any(index != V(graph)))) {
-    vs <- V(graph)
```
maelle, line 556: no need to create this intermediary value as one can use `vcount(graph)` directly.

R/attributes.R
```diff
@@ -539,28 +539,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
-    for (i in seq_along(value)) {
-      tmp <- value[[i]]
+    (index_number_not_vcount || index_not_vertices_indices)) {
+
+    make_value_correct_vector <- function(value_item, vcount) {
+      tmp <- value_item
       length(tmp) <- 0
```
maelle, line 561: This does something different based on whether `value_item` is a vector or a list.

``` r
thing <- list(1, 2, 4)
length(thing) <- 0
length(thing) <- 4
thing
#> [[1]]
#> NULL
#> 
#> [[2]]
#> NULL
#> 
#> [[3]]
#> NULL
#> 
#> [[4]]
#> NULL

stuff <- c(1, 2, 4)
length(stuff) <- 0
length(stuff) <- 4
stuff
#> [1] NA NA NA NA
```

<sup>Created on 2024-04-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/attributes.R
```diff
@@ -539,28 +539,34 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
-    for (i in seq_along(value)) {
-      tmp <- value[[i]]
-      length(tmp) <- 0
-      length(tmp) <- length(vs)
-      tmp[index] <- value[[i]]
-      value[[i]] <- tmp
+    (index_number_not_vcount || index_not_vertices_indices)) {
+
+    make_value_correct_vector <- function(value_item, vertices) {
+      tmp <- vctrs::vec_init_along(value_item, vertices)
```
maelle, line 560: it feels wrong to pass vertices when all I'm interested in is the number of vertices.

R/attributes.R
```diff
@@ -539,28 +539,34 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
-    vs <- V(graph)
-    for (i in seq_along(value)) {
-      tmp <- value[[i]]
-      length(tmp) <- 0
-      length(tmp) <- length(vs)
-      tmp[index] <- value[[i]]
-      value[[i]] <- tmp
+    (index_number_not_vcount || index_not_vertices_indices)) {
+
+    make_value_correct_vector <- function(value_item, vertices) {
```
maelle, line 559: to be further improved

1. name
2. body, or is it already the shortest it can be.

maelle: now it's short but is it readable

R/attributes.R
```diff
@@ -532,28 +532,34 @@ vertex.attributes <- function(graph, index = V(graph)) {
 
   assert_named_list(value)
 
-  if (any(sapply(value, length) != length(index))) {
+  if (!all(sapply(value, length) == length(index))) {
```
krlmlr, line 535:
```suggestion
  if (!all(lengths(value) == length(index))) {
```

R/attributes.R
```diff
@@ -532,28 +532,34 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
     stop("Invalid attribute value length, must match number of vertices")
   }
 
   if (!missing(index)) {
     index <- as_igraph_vs(graph, index)
 
-    if (any(duplicated(index)) || any(is.na(index))) {
+    if ((anyDuplicated(index) > 0) || (anyNA(index) > 0)) {
```
krlmlr, line 542: Good enough:

```suggestion
    if (anyDuplicated(index) || anyNA(index)) {
```

R/attributes.R
krlmlr: Natural sequence.

R/attributes.R
```diff
@@ -532,28 +532,34 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
+      value_item %>%
+        vctrs::vec_init_along(y = vertices) %>%
+        vctrs::vec_assign(i = index, value = value_item)
     }
+
+    value <- purrr::map(
+      value,
+      make_value_correct_vector,
+      vertices = V(graph)
+    )
```
krlmlr, line 562:
```suggestion
value <- map(value, ~.x[match(seq_along(vs), index)])
```

R/attributes.R
```diff
@@ -530,34 +530,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
-    (length(index) != vcount(graph) || any(index != V(graph)))) {
-    vs <- V(graph)
-    for (i in seq_along(value)) {
-      tmp <- value[[i]]
-      length(tmp) <- 0
-      length(tmp) <- length(vs)
-      tmp[index] <- value[[i]]
-      value[[i]] <- tmp
-    }
+  index_is_natural_sequence <- (length(index) == vcount(graph) && all(index == V(graph)))
```
krlmlr, line 558:
```suggestion
  index_is_natural_sequence <- (length(index) == vcount(graph) && identical(index, seq_len(vcount(graph))))
```

krlmlr (commented):
Thanks, looks good.

@clpippel: what do you think?

maelle:
@clpippel friendly reminder if you have time and want to have a look :smile_cat:

maelle:
@clpippel friendly reminder :smile_cat:

clpippel:
I am not sure this is OK. I tried:
```
library(igraph)
graph <- make_ring(1)
index <- V(graph)
(length(index) == vcount(graph) && identical(index, seq_len(vcount(graph))))
[1] FALSE
However I expected equality.

```
This was also a surprise to me:
```
index <- V(g)
identical(index, V(g))
[1] FALSE

```

clpippel:
@maelle, @krlmlr,
Note that E(g) is not identical with itself. Functtion identical(E(g), E(g)) returns "FALSE".  Object E(g) contains a "<weak reference>", which changes after every call. So it is not a good idea to use identical() together with E(g) or V(g).

It seems consistent to apply the same changes to vertex.attributes() (line 512).

R/attributes.R
```diff
@@ -530,34 +530,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
+  if (!missing(index) && !index_is_natural_sequence) {
+    value <- map(value, function(x) {
```
maelle, line 560:
```suggestion
    value <- map(value, set_value_at, idx = index, length_out = length(V(graph)))
```

R/attributes.R
```diff
@@ -530,34 +530,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
+      set_value_at(x, idx = index, length_out = length(V(graph)))
```
maelle, line 561:
```suggestion
```

R/attributes.R
```diff
@@ -530,34 +530,36 @@ vertex.attributes <- function(graph, index = V(graph)) {
...
+    })
```
maelle, line 562:
```suggestion
```

maelle:
Thank you @clpippel!!

Yes I have opened #1332 to not forget about the other "confusing" loop. Thanks for the reminder!

R/attributes.R
krlmlr: Let's do `seq(1, vcount(graph))` instead.

krlmlr (commented):
Thanks both for looking into it!

R/attributes.R
clpippel: @krlmlr, @maelle,

This will not work if `vcount(graph) == 0`.
seq(1,0) returns `[1] 1 0`.

Issue #1177, opened 2024-01-30
Avoid using deprecated functions from igraph/C


krlmlr:
Change to `#define IGRAPH_DEPRECATED __attribute__ ((__deprecated__))` and fix errors, one PR for each deprecated function.

szhorvat:
`igraph_hub_score()` and `igraph_authority_score()` are deprecated because it makes sense to always compute these two scores together (for efficiency, interpretability and to make sure we get a matching pair). The new function is called `igraph_hub_and_authority_scores()`.

I suggest merging these two in R under the name `hits_scores()`. See https://en.wikipedia.org/wiki/HITS_algorithm for where the name comes from. This name is short and there's precedence (both Mathematica and NetworkX use a variation of it).

In the longer term I also suggest deprecating the separate `hub_score()` and `authority_score()` functions in favour of the combined one (`hits_scores()`).

maelle:
@szhorvat should this comment be a separate issue?

krlmlr:
Running revdepchecks.

Issue #994, opened 2023-12-12
Update `make_lattice()` / `lattice()`


szhorvat:
To upgrade from the deprecated `igraph_lattice()` to the new `igraph_square_lattice()`, the R interface needs to adapt.

Right now `make_lattice` has a `circular` parameter that accepts a single logical. To match the capabilities of `igraph_square_lattice()`, there should be a parameter that accepts either a single logical, or a vector of logicals, one for each dimension.

It might be good to eventually rename `circular` to `periodic`, and to rename `make_lattice` to `make_square_lattice` to make room for other types of lattice generators (`igraph_triangular_lattice`, `igraph_hexagonal_lattice`).

szhorvat:
This you can move to `upgrade-2` @krlmlr

krlmlr:
We still need to deal with the `R CMD check` warning, but we can temporarily remove the "deprecated" declaration too.

szhorvat:
Can you add `-Wno-deprecated` to the build just before the submission?

Note that the C core won't compile without deprecation warnings with GCC because of [a GCC limitation](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79078). If CRAN truly wants zero warnings, the only way is to disable the warning. 

Yes, we absolutely must go through all deprecation warnings and check what the consequences of a proper fix are. I did look at this one and it is possible to make this change in the future as a non-breaking feature extension.

krlmlr:
I’m not sure if CRAN will accept a package with the switch. We can modify
our own code to remove the warnings.

Szabolcs Horvát ***@***.***> schrieb am Do. 4. Jan. 2024 um
10:49:

Issue #630, opened 2023-01-09
Split sample_bipartite() into two functions for the G(n,m) and G(n,p) case


szhorvat:
**What is the feature or improvement you would like to see?**

Separate `sample_bipartite()` into `sample_biparite_gnm()` and `sample_bipartite_gnp()`, and eventually deprecate `sample_bipartite()`.

**Motivation**

 - Consistency with the separate `sample_gnm()` and `sample_gnp()`
 - The $G(n,p)$ and $G(n,m)$ models are conceptually distinct
 - Reduces the number of optional argument
 - Explicit function names that reduce confusion
 - Already have separate `igraph_bipartite_game_gnp()` and `igraph_bipartite_game_gnm()` available in C/igraph

**References**

 - https://igraph.org/r/doc/sample_bipartite.html

szhorvat:
Consider this for 2.0 @krlmlr @maelle?

maelle:
Noting that if we do, the deprecation code for `bipartite.random.game()` will need to be updated.

maelle:
What to do regarding `bipartite()`?

szhorvat:
> What to do regarding `bipartite()`?

Same thing, split into `bipartite_gnm` and `bipartite_gnp`. Perhaps `sample_(bipartite())` should call `sample_bipartite()`, which then issues the deprecation warning?

PR #1352, opened 2024-05-02
feat: remove deprecated functions before 1.0

Antonov548:
## Deprecated

 - [x] `igraph_bipartite_game()` — see #630 - not use at all

## Updated

 - [x] `igraph_hub_score()` - replaced with `igraph_hub_and_authority_scores()`, see https://github.com/igraph/rigraph/issues/1177#issuecomment-1917980983
 - [x] `igraph_authority_score()` - replaced with `igraph_hub_and_authority_scores()`, see https://github.com/igraph/rigraph/issues/1177#issuecomment-1917980983
 - [x] `igraph_lattice()` — replace with `igraph_square_lattice()`, see #994 
 - [x] `igraph_laplacian()` — replace by `igraph_get_laplacian()`
 - [x] `igraph_erdos_renyi_game()` — not used in R.  
 - [x] `igraph_random_edge_walk()` — functionality merged into `igraph_random_walk()`.
 
 ## R side
 
 - [x] `hub_score()` - added `hits_scores`. `hub_score` is deprecated and use `hits_scores` under the hood.
 - [x] `authority_score()` - added `hits_scores`. `authority_score` is deprecated and use `hits_scores` under the hood.
 - [x] `make_lattice()` - use `square_lattice()`. `circular` parameter is deprecated. Added `periodic` parameter. `periodic` should be logical vector with same length as `dim` or single value which will be converted to vector.
 - [x] `laplacian_matrix` - use generated `get_laplacian_sparse_impl` and `get_laplacian_impl` instead of `R_igraph_laplacian`. `normalized` argument is deprecated with new `normalization` argument.
  - [x] `igraph_erdos_renyi_game()` — not used in R.  
  - [x] `igraph_random_edge_walk()` — updated `random_edge_walk` to use `random_walk_impl` with `edges` return.


Antonov548:
@krlmlr 
This is related to https://github.com/igraph/igraph/issues/2575#issuecomment-2083059712

Is it right that we should just remove deprecated functions?

szhorvat:
I edited the top post and added information on what can / should be done.

NAMESPACE
```diff
@@ -483,8 +481,7 @@ export(hrg.fit)
 export(hrg.game)
 export(hrg.predict)
 export(hrg_tree)
-export(hub.score)
-export(hub_score)
+export(hub_and_authority_scores)
```
szhorvat, line 484: I suggest `hits_scores()` here, see comment I linked

Antonov548:
> I edited the top post and added information on what can / should be done.

Thanks!

szhorvat:
Let me know if you need input here, for example on what parameters to set in the new function to get the behaviour of the old one.

----

Regarding hub and authority scores, I expect that removing these without deprecation would cause quite the uproar. What could be done is to implement `hub_score()` and `authority_score()` in pure R in terms of `hits_scores()` and keep them deprecated for a while.

There will also be another change in these two functions, as well as `eigenvector_centrality()` in 1.0: the `scale` parameter will be removed, and the function will always behave as if `scale=TRUE` was set. I'll open a new issue for preparing for this change.

Antonov548:
@szhorvat 

- I think we still need to remove `hub_score()` and `authority_score()` because implementing same functions, but using `hits_scores()` still break interface on R side, so maybe it worth just to remove them.

- Should then functions which are using `igraph_get_laplacian` to be extended with additional argument `mode`?

szhorvat:
> * I think we still need to remove `hub_score()` and `authority_score()` because implementing same functions, but using `hits_scores()` still break interface on R side, so maybe it worth just to remove them.

Why would it break the interface? `hits_scores()` returns both scores. Just drop one and rewrite the result to the format that `hub_score()` used to return.

As for the `scale` parameter, allow the parameter but ignore its value.

Antonov548:
> > * I think we still need to remove `hub_score()` and `authority_score()` because implementing same functions, but using `hits_scores()` still break interface on R side, so maybe it worth just to remove them.
> 
> Why would it break the interface? `hits_scores()` returns both scores. Just drop one and rewrite the result to the format that `hub_score()` used to return.
> 
> As for the `scale` parameter, allow the parameter but ignore its value. We never made any promises about the type of scaling used with `scale=FALSE`.

Because `hub_score()` and `authority score()` returns not just a vector. Of course we can maybe try to keep previous behavior by creating the similar list.
https://github.com/igraph/rigraph/pull/1352/files#diff-aa8edbd6189f5557d42166c470f065672ba52363e61d84dcd7bd3200c3b3420cL3860-L3862

Sorry, I didn't get regarding laplacian. It has also `mode=c("out", "in", "all", "total")` and `igraph_laplacian` hasn't this argument.

szhorvat:
> * Should then functions which are using `igraph_get_laplacian` to be extended with additional argument `mode`?

This will require a somewhat breaking change (which is why I kept pushing to do this for 2.0...)

The big change is not `mode` but the `normalization` parameter, which is NOT the same as `normalized`.  The old `igraph_laplacian()` normalized in an inconsistent way. You can see what it did here:

https://github.com/igraph/igraph/blob/223fd0e2c11b4c0a35452ffa68347206771dc0b4/src/properties/spectral.c#L403-L431

We need to transition from the old boolean `normalized` to the new `normalization` in some way. I'm not sure what the customary way is in R.  I guess we can temporarily have both `normalized` and `normalization` at the same time. Deprecate `normalized`, i.e. any explicit use of `normalized` should trigger a deprecation warning. I'm not sure what to do when the user passes _both_ `normalized` and `normalization` at the same time: I guess throw an error since this makes no sense?

szhorvat:
Re hub and authority scores:

Currently the function returns a list with many named elements. The main result is in `$vector`. `hits_scores()` would return the same, except that instead of `vector` it'd have both `$hub` and `$authority`. `hub_score()` would call `hits_scores()`, would remove `$authority` and would rename `$hub` to `$vector`.  The rest of the entries are the same between the two functions.

szhorvat:
I notice you were working on `make_lattice`. Please see this before going ahead: https://github.com/igraph/rigraph/issues/994

The R interface shouldn't always directly mirror the raw C interface. The `periodic` parameter should accept either a single logical value or a vector. The `circular` parameter probably shouldn't be removed immediately, breaking compatibility, but should go through a gradual deprecation.

Antonov548:
> I notice you were working on `make_lattice`. Please see this before going ahead: #994
> 
> The R interface shouldn't always directly mirror the raw C interface. The `periodic` parameter should accept either a single logical value or a vector. The `circular` parameter probably shouldn't be removed immediately, breaking compatibility, but should go through a gradual deprecation.

Sure. I think derpecation of old parameter name is quite easy.

Can you clarify please about single logical value or a vector? You mean it should be handled on R side? 
Because in core as I understood it could be `NULL` or vector with exact same length as `dimension`.

szhorvat:
> Can you clarify please about single logical value or a vector? You mean it should be handled on R side?
> Because in core as I understood it could be `NULL` or vector with exact same length as `dimension`.

Yes, it should be handled in R. I would not allow `NULL` at all. Instead: 

 - it's either a logical vector of the same length as the dimension (which can be sent directly to C)
 - or a logical scalar, which will be replicated dimension times (which can't be implemented in C, so needs to be handled in R)

The default should be `FALSE`.

R/make.R
```diff
@@ -1045,8 +1045,9 @@ full_graph <- function(...) constructor_spec(make_full_graph, ...)
 #' @param directed Whether to create a directed lattice.
 #' @param mutual Logical, if `TRUE` directed lattices will be
 #'   mutually connected.
-#' @param circular Logical, if `TRUE` the lattice or ring will be
-#'   circular.
+#' @param periodic Logical vector, Boolean vector, defines whether the generated lattice is
+#'   periodic along each dimension. This parameter may also be `NULL`, which
+#'   implies that the lattice will not be periodic.
```
szhorvat, line 1050: The docs need an update to match the new interface. The new interface looks good.

Antonov548: @szhorvat Do you mean description in the `make.R` of this function? Is current desciption is not valid anymore?
```R
#' `make_lattice()` is a flexible function, it can create lattices of
#' arbitrary dimensions, periodic or aperiodic ones. It has two
#' forms. In the first form you only supply `dimvector`, but not
#' `length` and `dim`. In the second form you omit
#' `dimvector` and supply `length` and `dim`.
```

szhorvat: Specifically the change I am commenting on is not valid anymore. It says that the parameter value has to be a logical vector or NULL, but in fact it is either a logical vector or a logical scalar, with specific interpretation for each.

Antonov548:
@maelle 
May I ask you to help with documentation in this PR? I trying to update documentation and have error:
```
✖ In topic 'authority_score.Rd': Skipping; no name and/or title.
✖ In topic 'hub_score.Rd': Skipping; no name and/or title.
```

I don't get what is missing in documentation, mentioned functions have `rdname`.

maelle:
I'll come back to this next week!

maelle:
Do we really want to _remove_ the functions though? Shouldn't they be deprecated first?

NAMESPACE
```diff
@@ -160,7 +160,6 @@ export(assortativity_nominal)
 export(asym_pref)
 export(asymmetric.preference.game)
 export(atlas)
-export(authority.score)
```
maelle, line 163: why remove them now?

R/centrality.R
```diff
@@ -1132,28 +1102,43 @@ diversity <- diversity_impl
 #' hub_score(g2)$vector
 #' authority_score(g2)$vector
 #' @family centrality
-hub_score <- function(graph, scale=TRUE, weights=NULL, options=arpack_defaults()) {
+hits_scores <- function(graph, scale=TRUE, weights=NULL, options=arpack_defaults()) {
```
maelle, line 1105: should this get the ... argument then checked with `rlang::check_dots_empty()`?
```suggestion
hits_scores <- function(graph, ..., scale=TRUE, weights=NULL, options=arpack_defaults()) {
```

R/centrality.R
```diff
@@ -1132,28 +1102,43 @@ diversity <- diversity_impl
...
+                 weights = weights,
+                 options = options)
+}
 
+#' @rdname authority_score
+#' @param options A named list, to override some ARPACK options. See
+#'   [arpack()] for details.
+#' @export
+authority_score <- function(graph, scale=TRUE, weights=NULL, options=arpack_defaults()) {
+  lifecycle::deprecate_soft("2.0.3", "authority_score()", "hits_scores()")
```
maelle, line 1117:
```suggestion
  lifecycle::deprecate_soft("2.0.4", "authority_score()", "hits_scores()")
```

R/centrality.R
```diff
@@ -1132,28 +1102,43 @@ diversity <- diversity_impl
...
 
-  hub_score_impl(graph = graph,
-                 scale = scale,
-                 weights = weights,
-                 options = options)
+  scores <- hits_scores(graph = graph,
+                  scale = scale,
+                  weights = weights,
+                  options = options)
+  scores$vector <- scores$authority
```
maelle, line 1132: and remove authority from the list?

R/centrality.R
```diff
@@ -1164,10 +1149,12 @@ authority_score <- function(graph, scale=TRUE, weights=NULL, options=arpack_defa
...
 
-  authority_score_impl(graph = graph,
-                       scale = scale,
-                       weights = weights,
-                       options = options)
+  scores <- hits_scores(graph = graph,
+                  scale = scale,
+                  weights = weights,
+                  options = options)
+  scores$vector <- scores$hub
```
maelle, line 1156: and remove hub from the list?

R/games.R
```diff
@@ -716,16 +716,19 @@ gnm <- function(...) constructor_spec(sample_gnm, ...)
...
-    "gnm" = 1
-  )
 
   on.exit(.Call(R_igraph_finalizer))
-  res <- .Call(
-    R_igraph_erdos_renyi_game, as.numeric(n), as.numeric(type1),
-    as.numeric(p.or.m), as.logical(directed), as.logical(loops)
-  )
+  if (type == "gnp") {
+    res <- res <- .Call(
```
maelle, line 722: why two `res <-`?

maelle:
Above authority_score and hub_score you need

```r
#' @title Kleinberg's hub and authority centrality scores.
```

R/games.R
Antonov548: Thanks for finding it. Just typo.

R/centrality.R
```diff
@@ -29,21 +29,6 @@ page.rank <- function(graph, algo = c("prpack", "arpack"), vids = V(graph), dire
...
-#' @description
-#' `r lifecycle::badge("deprecated")`
-#'
-#' `hub.score()` was renamed to `hub_score()` to create a more
-#' consistent API.
-#' @inheritParams hub_score
-#' @keywords internal
-#' @export
-hub.score <- function(graph, scale = TRUE, weights = NULL, options = arpack_defaults()) { # nocov start
-  lifecycle::deprecate_soft("2.0.0", "hub.score()", "hub_score()")
```
krlmlr, line 43: Keep it and direct to `hits_score()` instead?

R/centrality.R
```diff
@@ -134,21 +119,6 @@ bonpow <- function(graph, nodes = V(graph), loops = FALSE, exponent = 1, rescale
...
-#' @description
-#' `r lifecycle::badge("deprecated")`
-#'
-#' `authority.score()` was renamed to `authority_score()` to create a more
-#' consistent API.
-#' @inheritParams authority_score
-#' @keywords internal
-#' @export
-authority.score <- function(graph, scale = TRUE, weights = NULL, options = arpack_defaults()) { # nocov start
-  lifecycle::deprecate_soft("2.0.0", "authority.score()", "authority_score()")
```
krlmlr, line 148: Same?

krlmlr (commented):
Can you please review with Maëlle?

krlmlr:
For the test failure, the problem and solution pattern could look like this:

``` r
wrapped <- function(a, b, c) {
  list(a, b, c)
}

fun <- function(..., b = NULL, c = NULL) {
  wrapped(..., a = 5, b = b, c = c)
}

fun(b = 2, c = 3)
#> [[1]]
#> [1] 5
#> 
#> [[2]]
#> [1] 2
#> 
#> [[3]]
#> [1] 3
```

``` r
fun(a = 1, b = 2, c = 3)
#> Error in wrapped(..., a = 5, b = b, c = c): formal argument "a" matched by multiple actual arguments
```

``` r

fun <- function(..., a = NULL, b = NULL, c = NULL) {
  a <- if (is.null(a)) 5 else a
  wrapped(..., a = a, b = b, c = c)
}

fun(a = 1, b = 2, c = 3)
#> [[1]]
#> [1] 1
#> 
#> [[2]]
#> [1] 2
#> 
#> [[3]]
#> [1] 3
```

<sup>Created on 2024-05-30 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

maelle:
@Antonov548 I'll check out this PR locally and make the changes

- no longer removing the functions
- doing the argument adaptation mentioned above by Kirill
- ...

szhorvat:
In `sample_gnm()` / `sample_gnp()`, the `type1` and `type` variables can now be removed. `type1` is not used and the value of `type` can be inlined, simplifying code.

R/make.R
```diff
@@ -1045,8 +1045,9 @@ full_graph <- function(...) constructor_spec(make_full_graph, ...)
 #' @param directed Whether to create a directed lattice.
 #' @param mutual Logical, if `TRUE` directed lattices will be
 #'   mutually connected.
-#' @param circular Logical, if `TRUE` the lattice or ring will be
```
maelle, line 1048: note to self: re-add

szhorvat:
In `laplacian_matrix()`, the documentation need to be updated based on:

 - https://igraph.org/c/html/latest/igraph-Structural.html#igraph_get_laplacian
 - https://igraph.org/c/html/latest/igraph-Structural.html#igraph_laplacian_normalization_t

What's most important is to include the formulas for the normalization methods. The text is not sufficient to make it clear what these methods do.

szhorvat:
Can you rebase this on latest `main` so it'd be easier to test locally?

maelle:
I first want to make r cmd check pass, it's not ready yet

R/random_walk.R
```diff
@@ -64,4 +64,17 @@ random_walk <- function(
...
+    graph,
+    start,
+    steps,
+    weights = NULL,
+    mode = c("out", "in", "all", "total"),
+    stuck = c("return", "error")) {
+  mode <- match.arg(mode)
+  stuck <- match.arg(stuck)
+  out <- random_walk_impl(graph, start, steps, weights, mode, stuck)
+  # FIXME: Support returning the full structure
```
maelle, line 77: @Antonov548 what does this mean?

szhorvat:
It seems you copied the C enum constants directly ... these are not meaningful in R. What's needed is to amend the existing docs on normalization methods and explain what each one is, using the same names that the R function takes.

maelle:
Ok I'll revert and keep this for later #1384

maelle:
@Antonov548 can you please update the checklist at the very top and indicate which issues will be fixed by this PR?

R/make.R
maelle: @Antonov548 was this comment solved?

R/centrality.R
```diff
@@ -1104,7 +1104,8 @@ diversity <- diversity_impl
 #'   random surfer model, an edge with a larger weight is more likely to be
 #'   selected by the surfer.
 #' @param options A named list, to override some ARPACK options. See
-#'   [arpack()] for details.
+#'   [arpack()] for details.In packages, document ... with this standard tag:
```
szhorvat, line 1107: Is "In packages, document ... with this standard tag:" intentional here?

maelle: oops no, I'll fix it, thank you

maelle:
@Antonov548 I left some questions for you!

Antonov548:
> @Antonov548 can you please update the checklist at the very top and indicate which issues will be fixed by this PR?

All items in checklist are done. Updated list shows functions which were updated and corresponded list of R side - description what happened to deprecated function.

Antonov548:
It's green. Thanks a lot @maelle

szhorvat:
> Ok I'll revert and keep this for later #1384

Do you have a copy of that commit so I could edit it a bit and fix it up>?

krlmlr:
Running revdepchecks.

szhorvat:
Once this is done, can you please submit a PR to the C core removing all no-longer-needed deprecated entries from https://github.com/igraph/igraph/blob/master/interfaces/functions.yaml and https://github.com/igraph/igraph/blob/master/interfaces/types.yaml ?

krlmlr:
Added revdepcheck results. We have new failures, but they might come from the main branch. Double-checking.

szhorvat:
One of the issues is that in `hub_and_authority_scores`, the the generated C core, `hub_vector` is not renamed to `hub.vector`. In the R code it is. Either this should be fixed in Stimulus, or the code can be fixed manually and autogen disabled for this function.

szhorvat:
The other issue is that the `method` value renaming in `sample_degseq()` went a bit wrong and no longer accepts the old names due to the `method <- igraph.match.arg(method)`.  Example:

```
> sample_degseq(c(1,1,2,2,2),method='simple.no.multiple')
Error in `sample_degseq()`:
! `arg` must be one of "configuration", "vl", "fast.heur.simple",
  "configuration.simple", or "edge.switching.simple", not "simple.no.multiple".
Run `rlang::last_trace()` to see where the error occurred.
```

This is actually on `main`, not on this branch.

It should instead issue a deprecation warning. Can you have a look at this one @maelle ?

Antonov548:
> One of the issues is that in `hub_and_authority_scores`, the the generated C core, `hub_vector` is not renamed to `hub.vector`. In the R code it is. Either this should be fixed in Stimulus, or the code can be fixed manually and autogen disabled for this function.
> 
> I would go with a manual patch for the time being, so we can move ahead with the release. Stimulus can be fixed up with some options to control `_` vs `.` translation later. I have some vague memory about this being an issue in the past, but don't remember with which function. I think it was the opposite: we _didn't_ want the `.`-renaming, but it did it anyway for compatibility. Here we probably do want it for consistency with other functions.

Maybe it's still better to fix it in Stimulus?
I think it's not much changes in stimulus, but I see that it's this issue no only in `hit_scores()` function.

maelle:
https://github.com/igraph/rigraph/pull/1393

krlmlr:
The revdepcheck failures are a mismatch between `hub_score` (provided by `interface.c`)  and `hub.score` (expected by `aaa-auto.R`) . Can you please resolve?

Moving forward, I'd like to have more tests with the `"add.vertex.names"` option set to `TRUE` . I'll see what our tests do if we enable this globally.

maelle:
Relatedly, do we want to make the deprecation of `hub.score()` and `authority.score()` a soft deprecation instead?

maelle:
hub.vector not score (note to self)

maelle:
update hits_score() examples (they use the deprecated functions)

szhorvat:
Please make all graphs directed in the examples. This is important. These metrics make no sense for undirected graphs, and the examples confuse users.

Also, stars and rings are both bad examples. I suggest using `sample_gnm()` with at least twice as many edges as vertices, and make it directed.

szhorvat:
Or better: use `macaque` from `igraphdata`.

maelle:
@Antonov548 I do not understand why aaa-auto.R refers to `hub.vector`. How exactly is the body of `hub_and_authority_scores_impl()` created? In `functions-R.yaml` I only see

```yaml
igraph_hub_and_authority_scores:
    # This is a temporary hack; we need to find a way to handle default values
    # for dependencies. The problem is that the VERTEXINDEX type needs two
    # dependencies: the graph and the vertex set, but we only have the graph
    # in the argument list.
    DEPS: weights ON graph, hub_vector ON graph V(graph), authority_vector ON graph V(graph)
```

How do we best patch `hub_and_authority_scores_impl()`? It should refer to names with underscores not with dots. Thank you!

maelle:
The example I tried adding pairs as a reprex of what's broken @Antonov548 

``` r
library("igraphdata")
library("igraph")
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
```

``` r
data(macaque)
hits_scores(macaque)
#> This graph was created by an old(er) igraph version.
#>   Call upgrade_graph() on it to use with the current igraph version
#>   For now we convert it on the fly...
#> Error in names(res$hub.vector) <- vertex_attr(graph, "name", V(graph)): attempt to set an attribute on NULL
```

<sup>Created on 2024-06-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Antonov548:
> @Antonov548 I do not understand why aaa-auto.R refers to `hub.vector`. How exactly is the body of `hub_and_authority_scores_impl()` created? In `functions-R.yaml` I only see
> 
> ```yaml
> igraph_hub_and_authority_scores:
>     # This is a temporary hack; we need to find a way to handle default values
>     # for dependencies. The problem is that the VERTEXINDEX type needs two
>     # dependencies: the graph and the vertex set, but we only have the graph
>     # in the argument list.
>     DEPS: weights ON graph, hub_vector ON graph V(graph), authority_vector ON graph V(graph)
> ```
> 
> How do we best patch `hub_and_authority_scores_impl()`? It should refer to names with underscores not with dots. Thank you!

To not spend time to make changes in stimulus I can propose following change. Please check last commit.

I found a place in stimulus where `_` is replaced by dot. Seems this logic not synchronized with C code generation.

Antonov548:
He... I'm not sure what with GHA here. Maybe something wrong with cache.

krlmlr:
I don't see a caching step, can you please be more specific?

Antonov548:
> I don't see a caching step, can you please be more specific?

I mean this caches. I already had once the problem with them - https://github.com/igraph/rigraph/actions/caches.
But as I know we don't have easy way to clear all of them.

Antonov548:
@krlmlr the issue wasn't the cache. I had some troubles with search in VS code, so that I didn't see another usage of removed function. But it's good right now and maybe it make sense to run check for packages once again.

krlmlr:
This looks green. Running revdepchecks.

krlmlr:
Looks good enough to me?

Antonov548:
> Looks good enough to me?

Also looks good to me. Thanks for checking it.

szhorvat:
I'm on my phone and can't look properly, but this caught my eye: 

```
Error in igraph::graph.lattice(length(subwords) + 1, directed = TRUE) : 
  argument "circular" is missing, with no default
```

This is changing a deprecated function, which is perhaps not a good idea? 

The circular / periodic parameters are otherwise not mandatory, right?

szhorvat:
The netropy issue is not our problem. 

Did anyone look and check that the tidygraph one also isn't?

krlmlr:
tidygraph is present in main, I think this was also something that we can/should fix, in addition to the `circular` argument.

@maelle: can you please take a closer look?

This isn't meant to be a blocker for this PR, just something that we want to do before release.

szhorvat:
> This isn't meant to be a blocker for this PR, just something that we want to do before release.

Absolutely!

szhorvat:
`make_lattice` should also be renamed `make_square_lattice` to make room for `make_hex_lattice()` and `make_tri_lattice()` (or whatever you want to name them). https://github.com/igraph/rigraph/issues/994

krlmlr:
Thanks!

PR #1356, opened 2024-05-07
fix: fixed memory leak in `sample_degseq()`

szhorvat:
CC @Antonov548


szhorvat:
@Antonov548 I'm looking at `rinterface_extra.c` and I see that several updates were made without regard to memory management. Can you please have a look at some other functions that were updated and make the necessary corrections?

Here's one example:

https://github.com/igraph/rigraph/blob/5e1c7b4f708d43c830f6f856305e9b81329428b5/src/rinterface_extra.c#L7335-L7357

Notice that two new data structures are allocated, but only one is added to the finally stack. Generlaly, when you see a different number in `IGRAPH_FINALLY_CLEAN` than the preceding number of `destroy` calls, then something's wrong:

```c
  igraph_matrix_int_destroy(&merges);
  igraph_vector_int_destroy(&membership);
  IGRAPH_FINALLY_CLEAN(1);
```

Additionally, the `IGRAPH_R_CHECK` are missing from the `R_SEXP_to_matrix_int` and `igraph_vector_int_init`

I suggest starting by searching the file for `IGRAPH_FINALLY_CLEAN` and making corrections wherever the number doesn't match the number of `destroy` calls. This won't catch all issues, but it should catch many.

We need to be careful with changes where memory management is involved because finding issues later is difficult and painful.

szhorvat:
P.S. Most such fixes are easier by moving to auto-generation. Here I'm blocked in doing that by https://github.com/igraph/rigraph/issues/876

Antonov548:
> @Antonov548 I'm looking at `rinterface_extra.c` and I see that several updates were made without regard to memory management. Can you please have a look at some other functions that were updated and make the necessary corrections?
> 
> Here's one example:
> 
> https://github.com/igraph/rigraph/blob/5e1c7b4f708d43c830f6f856305e9b81329428b5/src/rinterface_extra.c#L7335-L7357
> 
> Notice that two new data structures are allocated, but only one is added to the finally stack. Generlaly, when you see a different number in `IGRAPH_FINALLY_CLEAN` than the preceding number of `destroy` calls, then something's wrong:
> 
> ```c
>   igraph_matrix_int_destroy(&merges);
>   igraph_vector_int_destroy(&membership);
>   IGRAPH_FINALLY_CLEAN(1);
> ```
> 
> Additionally, the `IGRAPH_R_CHECK` are missing from the `R_SEXP_to_matrix_int` and `igraph_vector_int_init`
> 
> I suggest starting by searching the file for `IGRAPH_FINALLY_CLEAN` and making corrections wherever the number doesn't match the number of `destroy` calls. This won't catch all issues, but it should catch many.
> 
> We need to be careful with changes where memory management is involved because finding issues later is difficult and painful.

Thanks for finding it. Seems sanitizer is good right now.
I'm agree that it would be much easier to fix with autogeneration.

Sometime it's really hard to find them all in 9k+ lines C file.

Antonov548:
I guess we can merge it regardless `devel` and `mac-os` build.

szhorvat:
Yes, it should be merged, but I'll leave out-of-order merges to @krlmlr :-)

krlmlr:
Thanks.

@Antonov548: Can you please tweak our CI/CD to ignore the failing builds, in a separate PR? I'd very much like to keep using AppVeyor, but don't have the time to properly fix CI/CD.

krlmlr:
s/AppVeyor/Aviator/

krlmlr:
This PR still has failing checks with macOS and Ubuntu 20.04. What can we do about that?

Antonov548:
> This PR still has failing checks with macOS and Ubuntu 20.04. What can we do about that?

Merge this one when it will be green: https://github.com/igraph/rigraph/pull/1357. Will it works?

krlmlr:
Maybe, let's see.

krlmlr:
Thanks! This might be included in the main branch already.

PR #1360, opened 2024-05-12
try to debug macos ci failure

szhorvat:


krlmlr:
@szhorvat: I'm not sure `R CMD build` gets any more verbose than that. But `R CMD INSTALL .` gives you the output of the `configure` script.

You can also get the contents of the `igraph_*.Rcheck` folder from the artifact: https://github.com/igraph/rigraph/actions/runs/9050298368/job/24865573278?pr=1360#step:9:360 . It might contain what you're looking for.

Try also `set -ex` in the `configure` script.

@Antonov548: Can you please assist?

PR #1365, opened 2024-05-14
refactor: use `check_string()` instead of `as.character()`

maelle:


szhorvat:
I'm not experienced enough with R to be able to follow the code without considerable time investment, but I wanted to point out that in many cases, numerical vertex IDs or vertex names are auto-converted to characters. This is convenient and probably desirable. I do not know if this change breaks any of the relevant use cases, but it's good to pay attention here.

R/make.R
```diff
@@ -645,10 +645,11 @@ make_graph <- function(edges, ..., n = max(edges), isolates = NULL,
 }
 
 make_famous_graph <- function(name) {
+  assert_character(name)
   name <- gsub("\\s", "_", name)
 
   on.exit(.Call(R_igraph_finalizer))
-  res <- .Call(R_igraph_famous, as.character(name))
```
maelle, line 651: what could name be given it's been fed to `gsub()`

maelle:
@szhorvat thanks! Yes, I'm hoping the revdeps checks would help with that. If this breaks too many things, I won't merge it.

szhorvat:
And just another reminder that our users are not just dependent packages, but also individuals who use igraph for research, often interactively.  Unfortunately it is difficult to track this second kind of user, even though IMO they are the majority.

schochastics:
I think this PR is a good idea, because it forces user to be rational on naming attributes.
But given that the reprex below currently works, I am not sure if we are opening some pandoras box with it?
``` r
library(igraph)
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
g <- sample_gnp(10,0.2)
g$`5` <- 4
g
#> IGRAPH 9a520e7 U--- 10 3 -- Erdos-Renyi (gnp) graph
#> + attr: name (g/c), type (g/c), loops (g/l), p (g/n), 5 (g/n)
#> + edges from 9a520e7:
#> [1] 2--3 1--4 1--8
graph_attr(g,5)
#> [1] 4
delete_graph_attr(g,5)
#> IGRAPH 9a520e7 U--- 10 3 -- Erdos-Renyi (gnp) graph
#> + attr: name (g/c), type (g/c), loops (g/l), p (g/n)
#> + edges from 9a520e7:
#> [1] 2--3 1--4 1--8
```

<sup>Created on 2025-03-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

maelle:
set_graph_attr should also use the assert function

krlmlr (commented):
Thanks. I'd prefer using the rlang standalone for checking, wanted to add it for some time now: https://github.com/r-lib/rlang/blob/main/R/standalone-types-check.R .

krlmlr:
Crazy idea: add a lifecycle compatibility layer that doesn't break existing code (which I'd consider off-label use, but keeping compat is a gesture of goodwill)?

R/igraph-package.R
```diff
@@ -10,6 +9,7 @@
 #' @importFrom rlang inject
 #' @importFrom rlang warn
 #' @importFrom rlang %||%
+#' @import rlang
```
maelle, line 12: needed for the standalone files (unless we prefer I try to find all function names that need to be namespaced in these files) but I didn't want to remove the lines before as I like the clarity... but maybe I should.

R/utils-assert-character.R
```diff
@@ -0,0 +1,7 @@
+assert_character <- function(x, name = "name", call = rlang::caller_env()) {
```
maelle, line 1: this uses the rlang thing now

maelle:
> Crazy idea: add a lifecycle compatibility layer that doesn't break existing code (which I'd consider off-label use, but keeping compat is a gesture of goodwill)?

can we decide after we've seen the revdeps results?

krlmlr (commented):
Thanks. Can you please add a test? Do we really need `assert_character()`, or does the default arg of `check_string()` do the right thing?

PR #1368, opened 2024-05-20
More auto-generation

szhorvat:
This increases auto-generation, as well as the robustness of some error checks.


R/games.R
```diff
@@ -971,19 +971,7 @@ degseq <- function(..., deterministic = FALSE) {
...
-    as.logical(directed), as.logical(citation)
-  )
-  if (igraph_opt("add.params")) {
-    res$name <- "Growing random graph"
-    res$m <- m
-    res$citation <- citation
-  }
-  res
-}
+sample_growing <- growing_random_game_impl
```
krlmlr, line 974: Moving forward, I'd rather have these functions _wrap_ than alias the impl functions. This will add flexibility to change the C layer. Also, ellipsis (but I forget if we need the `rlang::` prefix here). @maelle?

```suggestion
sample_growing <- function(n, m = 1, ..., directed = TRUE, citation = FALSE) {
  check_dots_empty()
  growing_random_game_impl(n, m, directed, citation)
}
```

R/structural.properties.R
```diff
@@ -1540,12 +1540,7 @@ reciprocity <- reciprocity_impl
 #' edge_density(g, loops = TRUE) # this is right!!!
 #' edge_density(simplify(g), loops = FALSE) # this is also right, but different
 #'
-edge_density <- function(graph, loops = FALSE) {
-  ensure_igraph(graph)
-
-  on.exit(.Call(R_igraph_finalizer))
-  .Call(R_igraph_density, graph, as.logical(loops))
-}
+edge_density <- density_impl
```
krlmlr, line 1543: Same.

krlmlr (approved):
Thanks. What's the motivation to add `RC`?

Moving forward, can you please tag @maelle for review first?

Antonov548 (approved):

R/games.R
szhorvat: Well, the original didn't have the ellipsis, and adding it is a breaking change, right?

That said, I agree with adding the ellipsis, and we can easily auto-generate it, since I had to add the parameter list manually on the R side anyway to correct the default for `directed`.

I'll go ahead  with the ellipsis if I don't hear further concerns from you.

R/structural.properties.R
szhorvat: Not sure I understand: do you want the following?

```R
edge_density <- function (graph, loops=FALSE) {
    density_impl(graph, loops)
} 
```

R/games.R
szhorvat: I pushed a commit that includes the `...` and `check_dots_empty()` in the auto-generated `growing_random_game_impl()`. Let me know if you still want the wrapper.

If we go with the wrapper, then default parameter values are specified purely in R instead of through auto-generation. (This is not good or bad, just pointing it out.)

szhorvat: Keep in mind that if we do wrappers, we will need to manually call `match.arg` in the wrapper (at least that's how I understand it). This function doesn't require `match.arg` so it's not affected.

R/structural.properties.R
maelle: I think so but with the ...?

szhorvat: OK. Do we want to do it with a wrapper or auto-generation?

krlmlr:
Thanks!

maelle:
Mmmh the covr error seems to imply the exclusions... aren't excluded (same in #1378)

PR #1371, opened 2024-05-23
test: refactor test-assortativity for readability

maelle:
@szhorvat does the name `reference_assortativity()` make sense for that test helper function?

### DIFF
--- a/tests/testthat/test-assortativity.R
+++ b/tests/testthat/test-assortativity.R
@@ -1,7 +1,7 @@
 test_that("assortativity works", {
   g <- read_graph(f <- gzfile("celegansneural.gml.gz"), format = "gml")
 
-  assR <- function(graph) {
+  reference_assortativity <- function(graph) {
     indeg <- degree(graph, mode = "in")
     outdeg <- degree(graph, mode = "out")
     el <- as_edgelist(graph, names = FALSE)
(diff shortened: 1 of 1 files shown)


tests/testthat/test-assortativity.R
```diff
@@ -13,29 +13,27 @@ test_that("assortativity works", {
...
-  as2 <- assR(g)
+  assortativity_igraph <- assortativity(
+    g,
+    values = degree(g, mode = "out"),
+    values.in = degree(g, mode = "in")
+  )
 
-  expect_that(asd, equals(as))
-  expect_that(asd, equals(as2))
+  expect_equal(assortativity_degree(g), assortativity_igraph)
```
maelle, line 22: is it bad to run `assortativity_degree(g)` twice? I find that more readable than giving it a name, that the reader has to keep track of.

szhorvat:
That name is no worse or better than the previous to me, use whatever you like. The function computes degree assortativity in a different way—it's not a baseline or a reference, it's just a "different" way (but really the same as what the C core does). Perhaps this sort of renaming is not the best use of our time. This is a trivial test that just repeats what's already being done in the C core. Users would be much better served by exposing new functionality, fixing bugs, and cleaning up the interface.

maelle:
I am going through all test files as part of the test refactoring efforts. :wink:

tests/testthat/test-assortativity.R
szhorvat: Running twice is fine.

szhorvat (approved):

Issue #1317, opened 2024-03-22
link C docs from R docs


maelle:
in each manual page.

Find a way to make that automatic.

cc @szhorvat

maelle:
It would not need to be automatic.

We could add a special tag automatically for already exposed functions, and manually for new exposed functions.

maelle:
for each function that is defined as or wraps an `_impl` function we can find what is `.Call` inside it.

Now from say `R_igraph_empty` we can know we want to link to `igraph_empty`. How do we find the link to https://igraph.org/c/doc/igraph-Basic.html#igraph_empty?

We could parse the docs folder of the C library every time we update the vendored file. In each xxml file there are links like

https://github.com/igraph/igraph/blob/03760a09cc7195dfaad3a1e60cff1632aa6a6118/doc/basicigraph.xxml#L176

so for each xxml file we can know what C things are documented in it.

- the URL fragment is the name of the C thing e.g. `#igraph_empty`
- the base URL is `https://igraph.org/c/doc/`
- the path is given by the id of the chapter in the xxml file https://github.com/igraph/igraph/blob/03760a09cc7195dfaad3a1e60cff1632aa6a6118/doc/basicigraph.xxml#L7

maelle:
so the tag we'd add would be something like

```
#'  @cmethod igraph_empty
```

and we'd use an internal lookup table to know in which chapter of the C docs it is documented.

szhorvat:
> How do we find the link to https://igraph.org/c/doc/igraph-Basic.html#igraph_empty?

You can parse the index, https://igraph.org/c/html/latest/ix01.html

The name of the target to build docs for C locally is `html`.

> so for each xxml file we can know what C things are documented in it.

Trying to get this from xxml files isn't going to be reliable.

----

Be careful, some R functions map to multiple C functions! Example `distances()`.

maelle:
then on how we make the special tag work I'll mimic https://github.com/moodymudskipper/devtag

maelle:
> Be careful, some R functions map to multiple C functions! Example distances().

Ah then they'd need multiple tags.

Great to know reg https://igraph.org/c/html/latest/ix01.html thank you!

maelle:
starting the R package over at https://github.com/igraph/igraph.r2cdocs, I parsed the index in https://github.com/igraph/igraph.r2cdocs/blob/main/data-raw/c-links.R

maelle:
[...]

PR #1376, opened 2024-05-23
feat: rename sample_degseq() method and add the edge.switching.simple method

### LINKED ISSUE #876: sample_degseq() method parameter values should be renamed to match C core

maelle:
Fix  #876

See @szhorvat, a new feature exposed after all the boring refactoring PRs. :grin: :innocent: Some questions

- where do the numbers corresponding to the method come from? I assumed the new one was 4.
- what could be a better test for the new method?

### DIFF
--- a/R/games.R
+++ b/R/games.R
@@ -752,37 +752,45 @@ random.graph.game <- erdos.renyi.game
 #' It is often useful to create a graph with given vertex degrees. This function
 #' creates such a graph in a randomized manner.
 #'
-#' The \dQuote{configuration} method (formerly called "simple") connects the out-stubs of the edges (undirected
-#' graphs) or the out-stubs and in-stubs (directed graphs) together. This way
-#' loop edges and also multiple edges may be generated. This method is not
-#' adequate if one needs to generate simple graphs with a given degree
-#' sequence. The multiple and loop edges can be deleted, but then the degree
-#' sequence is distorted and there is nothing to ensure that the graphs are
-#' sampled uniformly.
+#' The \dQuote{configuration} method (formerly called "simple") implements the
+#' configuration model. For undirected graphs, it puts all vertex IDs in a bag
+#' such that the multiplicity of a vertex in the bag is the same as its degree.
+#' Then it draws pairs from the bag until the bag becomes empty. This method may
+#'  generate both loop (self) edges and multiple edges. For directed graphs,
+#'  the algorithm is basically the same, but two separate bags are used
+#'  for the in- and out-degrees. Undirected graphs are generated
+#'  with probability proportional to \eqn{(\prod_{i<j} A_{ij} ! \prod_i A_{ii} !!)^{-1}},
+#'  where A denotes the adjacency matrix and !! denotes the double factorial.
+#'  Here A is assumed to have twice the number of self-loops on its diagonal.
+#'  The corresponding expression for directed graphs is \eqn{(\prod_{i,j} A_{ij}!)^{-1}}.
+#'   Thus the probability of all simple graphs
+#'   (which only have 0s and 1s in the adjacency matrix)
+#'   is the same, while that of non-simple ones depends on their edge and
+#'   self-loop multiplicities.
 #'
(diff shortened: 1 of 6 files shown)


R/games.R
```diff
@@ -909,13 +912,30 @@ random.graph.game <- erdos.renyi.game
...
+
   method1 <- switch(method,
-    "simple" = 0,
+    "configuration" = 0,
     "vl" = 1,
-    "simple.no.multiple" = 2,
-    "simple.no.multiple.uniform" = 3
+    "fast.heur.simple" = 2,
+    "configuration.simple" = 3,
+    "edge.switching.simple" = 4 #TODO: check number and see where it comes from
```
maelle, line 938: ideally we should make a note of where the numbers come from, for the next time someone needs to touch this part of the file.

szhorvat:
> where do the numbers corresponding to the method come from? I assumed the new one was 4.

For enum values in general, you need to look in `include/igraph_constants.h` in the C core:

https://github.com/igraph/igraph/blob/master/include/igraph_constants.h#L94

Enums in C are "named values" represented internally as integers. The numbers are integers. Notice that the first value is explicitly assigned `0`. Anything after that that does not have an explicit assignment increments by 1. So yes, the new values is `4`.

In the majority of cases we start at `0`, but there are a handful of exceptions, so it's good to have a look at the sources.

 > what could be a better test for the new method?

Checking that the degree sequence of the result is what we input is the best we can do here. But your test has a problem: you generate an undirected graph then you use both in- and out-degrees, which are for directed graphs. Add `directed=T` in `sample_gnm`.

It might be useful, but not essential, to also test the undirected case, with an undirected starting point and undirected degrees.

maelle:
> It might be useful, but not essential, to also test the undirected case, with an undirected starting point and undirected degrees.

Do you have an example at hand?

maelle:
I made edits @szhorvat. Thanks for your feedback!

szhorvat:
> Do you have an example at hand?

 - Generate a `directed=FALSE` graph
 - Take `degrees(graph, mode="all")`
 - Use this as a single input to `sample_degseq()`
 - Check that the result has the same degrees, again with `degrees(graph, mode="all")`

szhorvat:
Could you please update the docs by lifting the descriptions of the five methods from the C docs here? https://igraph.org/c/html/latest/igraph-Generators.html#igraph_degree_sequence_game  The current R docs have misleading statements.  It's quickest to copy-paste the method description from C, and fix up the math so it actually renders are proper math.

szhorvat:
I improved the C docs a bit in this commit: 563d594a0bc5aca462a2df64a968fa0cb2bd9bcd

szhorvat:
I made some more (independent) doc updates, so I'll just leave this link here instead of linking to comments:

https://github.com/igraph/igraph/blob/master/src/games/degree_sequence.c#L671

These updates did not affect the text I suggested you copy, but there's additional information there that might be worth copying (such as the "see also" section).

maelle:
Is this PR good as is? I should put https://github.com/igraph/rigraph/issues/1317 nearer the top of my list :wink:

szhorvat:
Yes, let's merge this. Further doc improvements are always possible, and should not block this PR.

PR #1377, opened 2024-05-24
feat: update cigraph

Antonov548:


Antonov548:
@szhorvat update of igraph/C. But also have to fix compilation problems.

Issue #977, opened 2023-12-01
graph()/make_graph() error in rev deps


maelle:
Taking part of the concerns of #970 to here.

In revdeps such as cran/BASiNETEntropy we see the error

```
Error in make_graph(edges = edges, n = n, isolates = isolates, directed = directed,  : 
  'simplify' should only be used for graph literals
```

for code such as `graph(edges = edgestofilter[[t]], directed = FALSE)`

Should the default value of the simplify argument (which is TRUE for both `graph()` and `make_graph()`) be changed? What about `graph.famous()`?

I wonder why the error didn't happen before.

maelle:
oooh that's due to how we pass parameters of the call to `graph()`, to `make_graph()`!

maelle:
preparing a PR. I see it's also problematic to pass both dir and directed.

maelle:
https://github.com/maelle/rigraph/pull/5

Issue #697, opened 2023-03-06
Use {lifecycle} instead of zzz-deprecate.R


maelle:
- Keep "redirecting" calls, but with lifecycle syntax.
- Completely remove most functions?

Apart from [automorphism](https://github.com/igraph/rigraph/commit/1c8d344e264f0346c699ffa7697554df68817c24) and [page.rank.old](https://github.com/igraph/rigraph/commit/b93a8681762ecbe9f94b9b7501a551dfe5a4e13e#diff-861f2ed6503102e90f22cc3d1169a8a001ef4364e066977f3a88ef06b180ed6a) most deprecations seem to be from 2015.

ntamas:
It seems to me that even though these functions are marked as deprecated "unofficially", they do not print any warnings to the user yet. `zzz-deprecate.R` seems to be a nice central place to add warnings to all the "old" dotted-style functions in the next version (say, igraph 1.5.0) so we can schedule their removal for igraph 1.6.0 or 2.0.0, whichever comes first.

PR #1390, opened 2024-06-04
refactor!: change make.R

maelle:
Fix https://github.com/igraph/rigraph/issues/697
Fix https://github.com/igraph/rigraph/issues/977
Fix #1014 

@krlmlr this simply inlines all the new function bodies inside the old ones. It should be more robust. Can you please run revdeps?


krlmlr:
Thanks! Let's clean up later.

PR #1397, opened 2024-06-07
refactor: use `lifecycle::deprecated()` now that we can

maelle:


R/conversion.R
```diff
@@ -238,10 +238,7 @@ get.adjacency.sparse <- function(graph, type = c("both", "upper", "lower"),
   el <- as_edgelist(graph, names = FALSE)
   use.last.ij <- FALSE
 
-  if (edges) {
```
maelle, line 241: since we've only been passing it FALSE anyway

krlmlr (commented):
Thanks!

PR #1398, opened 2024-06-07
chore!: Hard-deprecate `get.edge()` and `layout.grid.3d()` which have been deprecated for 10 years

maelle:


maelle:
worth running revdeps checks for I guess, but I think it's about time. :sweat_smile:

krlmlr (approved):
Thanks!

PR #1414, opened 2024-06-21
fix: `subgraph_centrality()` now ignores edge directions

szhorvat:
The method that this function uses does not work in general for directed graphs, and the concept of subgraph centrality is normally defined for undirected graphs only.

----

I consider this a bugfix, so I don't think any lifecycle work is necessary. A revdepcheck will be useful, as some users may have tried to use this for directed graphs (in which case the function wouldn't return meaningful results, as already indicated in its documentation).

Please consider including this simple fix in the next release.


szhorvat:
@maelle Please have a look at how I access functions from the `Matrix` library in the test. Is this the customary way in tests?

R/centrality.R
```diff
@@ -799,7 +798,11 @@ subgraph_centrality <- function(graph, diag = FALSE) {
   if (!diag) {
     diag(A) <- 0
   }
-  eig <- eigen(A)
+  # Ignore edge directions in directed graphs
+  if (is_directed(graph)) {
+    A <- A + Matrix::t(A)
+  }
+  eig <- eigen(A, symmetric = TRUE)
```
krlmlr, line 805:
```suggestion
  # This calls lapack and creates a dense matrix, but accepts the sparse matrix A
  # We can choose to convert A to a dense matrix right away, but it doesn't matter
  eig <- eigen(A, symmetric = TRUE)
```

krlmlr (commented):
@maelle, can you please take another look and approve if good?

tests/testthat/test-subgraph_centrality.R
```diff
@@ -0,0 +1,24 @@
+test_that("subgraph_centrality works", {
+  g <- make_graph("Frucht")
```
maelle, line 2: I'll make a commit to rename those to not re-use names which is a pet peeve of mine :innocent: https://masalmon.eu/2024/05/23/refactoring-tests/

maelle (approved):
made a few mostly cosmetic changes :nail_care:

tests/testthat/test-subgraph_centrality.R
szhorvat: I don't understand what you mean.

maelle: In the test two objects were created with the name "g". I prefer the two names to be different.

szhorvat: I see what you mean after looking at your commits, but keep in mind that there is no special significance to this being the Frucht graph here. If all you need is to not reassign `g`, I suggest `g1` and `g2` for similar situations in the future.

The reason why mathematicians don't spell out variable names is  not that they don't know any better. It's because sometimes it is useful to be generic and not be distracted by things that are irrelevant to the work at hand.

maelle: Fair enough, but with g1 and g2 I'd easily get confused. They can be uninformative but they have to be really different for me :grin:

PR #1416, opened 2024-06-22
feat: `sample_chung_lu()`

### LINKED ISSUE #1369: Add `igraph_chung_lu_game()` to R
This is a well-know random graph models that will likely be appreciated by our users. It should be named `sample_chung_lu()`.

szhorvat:
Closes #1369.

Translating the documentation was a lot of work. If there are minor stylistic issues to fix, I'd appreciate it if someone could take this over and fix those. Let's try to have this in the next release please.


R/aaa-auto.R
```diff
@@ -391,6 +391,25 @@ simple_interconnected_islands_game_impl <- function(islands.n, islands.size, isl
   res
 }
 
+chung_lu_game_impl <- function(out.weights, in.weights=NULL, loops=TRUE, variant=c("original", "maxent", "nr")) {
```
krlmlr, line 394: Do we want this?

```suggestion
chung_lu_game_impl <- function(out.weights, ..., in.weights=NULL, loops=TRUE, variant=c("original", "maxent", "nr")) {
```

R/games.R
```diff
@@ -2032,6 +2032,148 @@ sample_islands <- simple_interconnected_islands_game_impl
...
+#'   100, 
+#'   degree(sample_chung_lu(c(1, 3, 2, 1), c(2, 1, 2, 2)), mode = "out")
+#' ))
+#' 
+#' rowMeans(replicate(
+#'   100, 
+#'   degree(sample_chung_lu(c(1, 3, 2, 1), c(2, 1, 2, 2), variant = "maxent"), mode='out')
+#' ))
+#' @export
+sample_chung_lu <- chung_lu_game_impl
```
krlmlr, line 2169: Not sure if we should forward or alias. Both have advantages.

krlmlr (commented):
Thanks! I glimpsed over the text and the autogenerated code, looks good. Would you add the ellipsis?

R/aaa-auto.R
szhorvat: No, definitely not. The ellipsis should come _after_ `in.weights`, not before.

szhorvat:
> Would you add the ellipsis?

I think the ellipsis makes sense like this:

```r
function(out.weights, in.weights=NULL, ..., loops=TRUE, variant=c("original", "maxent", "nr"))
```

I can add it, but I need some help here. Notice that there are two functions on the same doc page (as with all `sample_` and `make_` functions), and one of them already has a `...`, which is already documented:

image

<br>

image

Is this going to be an issue? What should I do here? Just keep the docs as they are even though it is slightly confusing?

krlmlr:
Good point. I pushed a proposal, it's a bit of work but solves the issue at hand and also offers autocomplete for `chung_lu()` .

How is `chung_lu()` used anyway?

szhorvat:
> How is `chung_lu()` used anyway?

It's used with `sample_()`. The `make_()` doc page has examples. One can add constructor modifiers: https://r.igraph.org/reference/index.html#constructor-modifiers  I don't have much direct experience with using these.

szhorvat:
I think this is done and ready to go, right?

krlmlr:
Thanks!

szhorvat:
Thanks @krlmlr !

szhorvat:
I forgot to mark it as experimental, which is necessary since it's experimental in C. I pushed this change to `main` directly: 4a3983d97861408c329f9c2ebeb9f90bcc3811a0 Are you okay with this?

krlmlr:
Works for me, as long as the main branch stays green 🙃

szhorvat:
OK, I'll continue to make simple doc changes on `main`, for efficiency.

PR #1421, opened 2024-07-02
test: rename test file to align with script structure.info.R

maelle:
Really minimal.

@szhorvat for info part of the refactoring of tests will consist in aligning script names and test filenames because this allows to easily jump from test to script and vice versa (using `usethis::use_r()` and `usethis::use_test()`).


szhorvat:
If I understand you correctly, you are asking me to follow certain naming standards. I'm happy to do so but I have a lot of trouble keeping these in my head (even things I came up with myself). Could you please open a new [wiki page](https://github.com/igraph/rigraph/wiki) and write down  (very briefly) the standards we need to follow? I can then refer to that, and I don't need to remember anything by heart :-) 

We do the same in the C core (see e.g. [here](https://github.com/igraph/igraph/wiki/Guidelines-for-function-argument-ordering) and [here](https://github.com/igraph/igraph/wiki/Documentation-guidelines)).

PR #1433, opened 2024-07-17
fix: use V(graph) as we should

### LINKED ISSUE #1427: Use `as.numeric(seq_len(n))` in favor of `seq(1, n)`

https://github.com/igraph/rigraph/pull/1330#discussion_r1665753951

We also need the `by` argument to ensure that the output is a numeric.

Perhaps add a function `seq_numeric()` that has a special case for zero-length output?

### To reproduce

`seq(by = 1)` seems to be the fastest, look at the units. Maybe it's worth arguing with R core to support `seq(start, start - by, by)` without error and to return an empty vector in this case?

``` r
x <- seq(1, 1, 1)
class(x)
#> [1] "numeric"
```

``` r
x
#> [1] 1
```

``` r
seq(1, 0, 1)
#> Error in seq.default(1, 0, 1): wrong sign in 'by' argument
```

``` r

bench::mark(
  sum(seq(1, 100000, 1)),
  sum(as.numeric(rlang::seq2(1, 100000))),
  sum(as.numeric(seq_len(100000)))
)
#> # A tibble: 3 × 6
#>   expression                             min median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                           <bch> <bch:>     <dbl> <bch:byt>    <dbl>
#> 1 sum(seq(1, 1e+05, 1))                628µs  791µs     1233.    1.91MB     46.8
#> 2 sum(as.numeric(rlang::seq2(1, 1e+05… 861ns  984ns   898993.        0B      0  
#> 3 sum(as.numeric(seq_len(1e+05)))      164ns  246ns  3139792.        0B    314.
```

<sup>Created on 2024-07-05 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### System information

_No response_

### LINKED ISSUE #1432: V(g) not the same as 1:vcount
thanks @szhorvat 

Relevant for

#1330
#1326 (looks ok actually)
#1427

maelle:
Fix #1432 
Fix #1427

cc @krlmlr @clpippel

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -520,7 +520,7 @@ vertex.attributes <- function(graph, index = V(graph)) {
 
   if (!missing(index)) {
     index_is_natural_sequence <- (length(index) == vcount(graph) &&
-        identical(index, seq(1, vcount(graph))))
+        identical(index, V(graph)))
     if  (!index_is_natural_sequence) {
       for (i in seq_along(res)) {
         res[[i]] <- res[[i]][index]


R/attributes.R
```diff
@@ -520,7 +520,7 @@ vertex.attributes <- function(graph, index = V(graph)) {
 
   if (!missing(index)) {
     index_is_natural_sequence <- (length(index) == vcount(graph) &&
-        identical(index, seq(1, vcount(graph))))
+        identical(index, V(graph)))
```
krlmlr, line 523: I still think this should be

```suggestion
        identical(index, as.numeric(seq_len(vcount(graph)))))
```

We want to cater for cases where `!missing(index)` but `index` is still essentially `V(g)` . I see:

``` r
library(igraph)
g <- make_ring(10)

identical(V(g), V(g))
#> [1] FALSE
```

<sup>Created on 2024-07-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Similarly to https://github.com/igraph/rigraph/issues/1432#issuecomment-2233902313.

clpippel:
```
graph <- make_ring(5)
index <- V(graph)
identical(index, as.numeric(seq_len(vcount(graph))))
# returns FALSE.
```
Try `all(index == V(graph))`

krlmlr: @clpippel: `all()` is challenging if the vectors are of different length? We'd still want to check the lengths first.

Would the `all()` route be faster?

krlmlr: Perhaps we also want to check `isTRUE(attr(g, "is_all"))` ?

``` r

library(igraph)
g <- make_ring(10)
dput(V(g))
#> structure(1:10, class = "igraph.vs", is_all = TRUE, env = <weak reference>, graph = "516d6a01-f041-48ae-bbe3-60dd751a8f4b")
```

<sup>Created on 2024-08-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Need to check if this other code path is ever touched -- easiest way by breaking it and running tests or revdeps.

krlmlr:
Continuing in #1427.

PR #1436, opened 2024-07-18
feat: export `.from()` etc. with behavior similar to `dplyr::across()`

### LINKED ISSUE #1435: Export `.from` etc. with behavior similar to `dplyr::across()`
and link to help pages.

Motivation:

https://github.com/maelle/simcausal/blob/b8dcf741e9388e48f1220b335add26fb073f93a1/R/simcausal.r#L422-L423

``` r
dplyr::across()
#> Error in `dplyr::across()`:
#> ! Must only be used inside data-masking verbs like `mutate()`,
#>   `filter()`, and `group_by()`.
igraph::.from()
#> Error: '.from' is not an exported object from 'namespace:igraph'
```

<sup>Created on 2024-07-18 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

maelle:
Fix #1435

A drawback: the tree "nei" functions are actually only for _index_ sequences but I let them output the same message. Is that bad?

Also, the help link does not work for me but it looks correct, maybe just a problem with my IDE version? I'd be grateful if you could try it on your side.


krlmlr:
Would you like to add examples of correct usage, or a link to the help topic where they are described? Perhaps even merge with that existing help topic (if it, well, exists)?

maelle:
> These days, I'd use generated code instead of clever code like this. One drawback is that environment(.from) is not the igraph environment; to achieve this, we'd need even cleverer code.

Do we need to tackle this here?

maelle:
Ooooh https://roxygen2.r-lib.org/articles/rd-formatting.html?q=link#operators

maelle:
> Would you like to add examples of correct usage, or a link to the help topic where they are described? Perhaps even merge with that existing help topic (if it, well, exists)?

Painfully added the links. :grin:

krlmlr (approved):
Let's remove the `keywords internal` at some point, shouldn't be a blocker for now.

maelle:
I don't think they should be external though. Users should only get to them from the error message.

PR #1442, opened 2024-08-13
refactor: stop passing nei &co as using the nei() &co functions will …

maelle:
…already have triggered an error


maelle:
ah no this does not make any sense.

PR #1444, opened 2024-08-13
test: add test for hits_scores() directly

maelle:


szhorvat:
A good test graph for HITS is the following:

```
make_graph(c(1, 3, 1, 6, 1, 10, 2, 1, 3, 1, 4, 2, 4, 7, 4, 9, 5, 4, 5, 6, 5, 8, 6, 3, 7, 1, 7, 5, 7, 6, 7, 10, 8, 4, 9, 6, 10, 5, 10, 7), directed = TRUE)
```

An issue with HITS is that often there are multiple valid solutions. I constructed this graph so that there would be precisely one valid solution.

The criterion used is that the graphs corresponding to $AA^T$ and $A^TA$ are connected, with $A$ the adjacency matrix.

PR #1445, opened 2024-08-13
refactor: remove lazyeval usage in `[.igraph.vs`

### LINKED ISSUE #1426: replace vendored lazyeval with rlang
from #1419

maelle:
Fix #1426 

I'm less confident about this one.


src/cpp11.cpp
```diff
@@ -479,10 +479,6 @@ extern SEXP R_igraph_write_graph_lgl(SEXP, SEXP, SEXP, SEXP, SEXP);
 extern SEXP R_igraph_write_graph_ncol(SEXP, SEXP, SEXP, SEXP);
 extern SEXP R_igraph_write_graph_pajek(SEXP, SEXP);
 extern SEXP UUID_gen(SEXP);
-extern SEXP make_lazy(SEXP, SEXP, SEXP);
```
maelle, line 482: this file is read-only, how was I supposed to remove these lines then? :sweat_smile:

szhorvat: Regenerate using `R -e 'cpp11::cpp_register()'`. This is why I keep insisting that these procedures must be documented in wiki pages.

maelle: oh right, thanks! or the contributing guide, a more usual place for R packages. :smile_cat:

R/iterators.R
```diff
@@ -517,30 +517,33 @@ simple_vs_index <- function(x, i, na_ok = FALSE) {
...
+  if (length(args) == 0) {
+    return(x)
+  }
 
-  ## Special case, no argument (but we might get an artificial
-  ## empty one
-  if (length(args) < 1 ||
-    (length(args) == 1 && inherits(args[[1]]$expr, "name") &&
-      as.character(args[[1]]$expr) == "")) {
+  ## or empty (what is the test case for this??)
```
krlmlr, line 527: `g[,]` perhaps? Or `g[, na_ok = FALSE]` ?

R/iterators.R
```diff
@@ -517,30 +517,33 @@ simple_vs_index <- function(x, i, na_ok = FALSE) {
...
+  ## try() because of .env$something would otherwise error
+  first_is_empty <- !nzchar(paste(rlang::quo_get_expr(args[[1]]), collapse = ""))
```
krlmlr, line 529: `rlang::is_missing(args[[1]])` ? But we need a test case...

The `.ignore_empty` argument to `rlang::enquos()` might take care of that.

R/iterators.R
```diff
@@ -517,30 +517,33 @@ simple_vs_index <- function(x, i, na_ok = FALSE) {
...
+  if (length(args) == 1 && first_is_empty) {
     return(x)
   }
 
   ## Special case: single numeric argument
-  if (length(args) == 1 && inherits(args[[1]]$expr, "numeric")) {
```
krlmlr, line 534: This case is for literal numeric values and can be kept as is? What about integer values?

``` r
rlang::expr(5L) |> inherits("numeric")
#> [1] FALSE
```

<sup>Created on 2024-08-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

We can keep most of this logic as is also when using rlang:

``` r
library(rlang)

tester <- function(...) {
  quos <- enquos(...)

  exprs <- purrr::map(quos, quo_get_expr)
  purrr::map(exprs, class)
}

tester(5, symbol, x + 1)
#> [[1]]
#> [1] "numeric"
#> 
#> [[2]]
#> [1] "name"
#> 
#> [[3]]
#> [1] "call"
```

<sup>Created on 2024-08-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

krlmlr (commented):
Nice!

maelle:
Does this make any sense:

``` r
library("igraph")
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
g <- make_ring(10)
V(g)[,]
#> + 10/10 vertices, from 847d690:
#>  [1]  1  2  3  4  5  6  7  8  9 10
V(g)[, na_ok = FALSE]
#> + 10/10 vertices, from 847d690:
#>  [1]  1  2  3  4  5  6  7  8  9 10
E(g)[,]
#> + 10/10 edges from 847d690:
#>  [1] 1-- 2 2-- 3 3-- 4 4-- 5 5-- 6 6-- 7 7-- 8 8-- 9 9--10 1--10
E(g)[, na_ok = FALSE]
#> + 0/10 edges from 847d690:
```

<sup>Created on 2024-08-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

maelle:
with the version of main it's actually worse

``` r
library("igraph")
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
g <- make_ring(10)
V(g)[,]
#> Error in eval(x$expr, data, x$env): argument is missing, with no default
V(g)[, na_ok = FALSE]
#> + 10/10 vertices, from 2584d97:
#>  [1]  1  2  3  4  5  6  7  8  9 10
E(g)[,]
#> Error in eval(x$expr, data, x$env): argument is missing, with no default
E(g)[, na_ok = FALSE]
#> Error in eval(x$expr, data, x$env): argument is missing, with no default
```

<sup>Created on 2024-08-22 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

krlmlr:
The `E(g)[, na_ok = FALSE]` case looks odd, can we clean that up as part of this PR? Looks good otherwise.

maelle:
I thought `V(g)[, na_ok]`  was the special case of an empty argument but it's apparently not.

maelle:
Actually `[.igraph.es` has no `na_ok` argument! :woman_facepalming:

PR #1450, opened 2024-08-20
chore: fix TODO related to missing citation

maelle:


szhorvat:
When editing references, let's try to add DOI links as well. In this case,  https://doi.org/10.1111/0081-1750.00098

PR #1452, opened 2024-08-20
refactor: drop attributes in a more straightforward fashion in `graphlet_basis()`

maelle:
I actually wonder what the original TODO meant. ("Going through all TODO comments" is an item in the planned ISC work)


szhorvat:
I'm not immediately sure what this is about either.



> ("Going through all TODO comments" is an item in the planned ISC work)

I just wanted to note that there's a chance that there are some TODOs left in there that really concern the C core. In the early days the separation between the C library and the R interface was not very strong.

maelle:
I'll ping you when needed then, thank you! For this one if no one remembers what it was about... maybe this refactor is enough as it does the same thing as before just cleaner.

szhorvat:
I took another look, and I just don't see how dropping attributes makes any difference. I suggest removing the code that drops attributes.

maelle:
Ok, let's see how it goes :smile_cat:

maelle:
argh the sanitizer is failing yet again, but the rest looks ok (excluding revdeps not checked here :sweat_smile: )

krlmlr:
Thanks!

Issue #1447, opened 2024-08-18
Add `igraph_feedback_vertex_set()`


szhorvat:
When writing tests, the same things apply that I wrote for _exact methods_ for the feedback arc set function in #1446

PR #1470, opened 2024-08-22
chore: Update vendored sources to igraph/igraph@03760a09cc7195dfaad3a1e60cff1632aa6a6118

szhorvat:
@krlmlr For revdepcheck purposes, as agreed. Somehow I don't see any deprecation warnings, despite `igraph_minimum_spanning_tree_*` functions being deprecated. @Antonov548, are deprecation warning disabled now? I don't see where.


szhorvat:
Why are checks not running? Do I need to mark as ready-for-review?

R/aaa-auto.R
```diff
@@ -852,7 +852,7 @@ distances_johnson_impl <- function(graph, from=V(graph), to=V(graph), weights) {
   res
 }
 
-distances_floyd_warshall_impl <- function(graph, from=V(graph), to=V(graph), weights=NULL, mode=c("out", "in", "all", "total"), method) {
+distances_floyd_warshall_impl <- function(graph, from=V(graph), to=V(graph), weights=NULL, mode=c("out", "in", "all", "total"), method=AUTOMATIC) {
```
szhorvat, line 855: @Antonov548 This will need fixing (although currently it is unused). Translation needs to be set up for `igraph_floyd_warshall_algorithm_t`, or code generation needs to be disabled for this function (which is what I recommend now).

R/aaa-auto.R
```diff
@@ -1370,6 +1370,27 @@ feedback_arc_set_impl <- function(graph, weights=NULL, algo=c("approx_eades", "e
   res
 }
 
+feedback_vertex_set_impl <- function(graph, weights=NULL, algo=EXACT_IP) {
```
szhorvat, line 1373: @Antonov548 Again, `algo=EXACT_IP` is invalid code, which needs to be dealt with, either by setting up translation for the new type, or removing code generation for this function temporarily. Exposing this function should be really easy, as it's near-identical to the feedback arc set one (so it's a copy and paste job). Ref: https://github.com/igraph/rigraph/issues/1447

R/aaa-auto.R
szhorvat: I disabled codegen for this function, see last commit.

szhorvat:
Disabled codegen for all `distances` functions, since these are exposed manually, and moving them to autogen is unlikely to happen in the near future due to the custom interface. The extra commit finally allows checks to run.

Antonov548:
> @Antonov548, are deprecation warning disabled now? I don't see where.

@szhorvat deprecation warnings are disabled in [igraph_config.h](https://github.com/igraph/rigraph/blob/main/src/vendor/igraph_config.h) in order to pass CI and CRAN. You can define `IGRAPH_DEPRECATED_ENUMVAL` to see deprecation warnings.

szhorvat:
> @szhorvat deprecation warnings are disabled in [igraph_config.h](https://github.com/igraph/rigraph/blob/main/src/vendor/igraph_config.h) in order to pass CI and CRAN. You can define `IGRAPH_DEPRECATED_ENUMVAL` to see deprecation warnings.

I think this is perfect for now.

krlmlr:
Checks run in #1472. I haven no idea what's going on.

krlmlr:
No, they are not!

szhorvat:
Tests pass, if revdepchecks are fine, this can be merged @krlmlr

krlmlr:
I currently can't run revdepchecks...

szhorvat:
> I currently can't run revdepchecks...

That's alright. Can we merge this though, even without doing the revdepchecks within the PR? We'd need to deal with revdepcheck breakages either way. Merging this now opens the door for exposing new functionality.

krlmlr:
What functionality would you like to expose? Can we do it inside this PR? As long as it's only new features, it shouldn't make a difference for the revdepchecks.

It's easier from an organizational perspective to delay the merge until we can run the revdepchecks.

szhorvat:
> Can we do it inside this PR?

Yes.

szhorvat:
I'm going to have a look at these now.

szhorvat:
I want to refactor `R_check_int_scalar()` and similar functions to be macros instead. This will allow error messages to point to the exact location of the failure in the sources. Any objection, @Antonov548 @krlmlr ?

krlmlr:
Good idea re macros, no objections from me. What's the impact on the size of the compiled library?

szhorvat:
Convenience link to revdepchecks README:

https://github.com/igraph/rigraph/blob/e04c35515df8ebfdfb7be49f94ee85d3f00d8b48/revdep/README.md

### manynet

They pass a vector for the `islands.size` parameter of `sample_islands()`. This is invalid: it must be a scalar integer, and this has always been so. 

I don't recall any recent changes in igraph that would cause this failure. We did add more stringent checks in 2.0.0, but that should have caught this misuse much earlier.

_The fix needs to be done in the other package._

### MetaNet

???

Doesn't seem related to C core changes, I won't dig deeper at this time.

### netropy

They are passing NaN (or NA) values as edge weights when finding shortest paths. This has always been invalid.

Again, as I recall, the check for this was added a long time ago, not recently.

_The fix needs to be done in the other package._

### priorCON

Might need more diggint, but overall it looks like they expect a specific result for some sort of community detection (didn't check which one).

I did make a fix to `igraph_community_multilevel()` (`cluster_louvain()` in R) which causes it to return a different result for the same random seed.  We do not guarantee the same result for the same seed across bugfix versions—this would prevent many kinds of bugfixes. We only guarantee the same statistical properties.

_The fix needs to be done in the other package:_ Do not expect a specific output (even for the same seed) when writing tests that use stochastic igraph functions.

### R6causal

???

```
Found the following significant warnings:
  Note: possible error in '.to(l[i])': unused argument (l[i]) 
  Note: possible error in '.from(l[i])': unused argument (l[i]) 
```

Does this ring a bell @maelle ? You were modifying these functions, as I recall.

### simcausal

???

Doesn't look igraph related.

### skynet

Looks like this is due to a deliberate deprecation, fix needs to be done in other package. CC @maelle

szhorvat:
> What's the impact on the size of the compiled library?

Probably very small, but needs to be measured, which should be easy to do. The checks themselves should add very little code. The large part is the error string, but the compiler consolidates duplicate strings.

szhorvat:
@krlmlr I opened issues for two affected packages. Can you email netropy? They don't seem to have a GitHub repo. I'll leave the rest to the R team 😊

szhorvat:
Actually, we can probably ping @schochastics here for `netropy`—please see above.

maelle:
Reg skynet https://github.com/ropensci/skynet/pull/11, will also email a reminder

krlmlr:
Only priorCON seems to be a new problem in this branch: https://github.com/igraph/rigraph/compare/f-revdep-2...try/c-core-2024-08-22

krlmlr:
The other packages are also a problem in the main branch.

schochastics:
@szhorvat we are on it and will report back

schochastics:
Turns out there was a call to `graph_from_adjacency`  where the input matrix contained NA weights on the lower.tri to indicate symmetry. This is being fixed in a pull request (https://github.com/termehs/netropy/pull/4)

szhorvat:
Thanks so much @schochastics and @termehs!

termehs:
Thanks all! New netropy version on its way to CRAN 😊

szhorvat:
@krlmlr I had to make a fairly crude change to the C core to deal with potentially incorrect results (see https://github.com/igraph/igraph/pull/2674). I would like to request another round of revdepchecks for this PR. I'll talk to packages that are affected by this change. Hopefully there are not many.

szhorvat:
We still have the weird issue that CI tests don't get run for C core updates.

szhorvat:
Can we merge this please (and have another revdepcheck round)?

krlmlr:
Running checks now.

krlmlr:
No regressions, good to go. Thanks!

szhorvat:
Thanks @krlmlr! I am almost surprised that the last C core change didn't cause revdepcheck failures, but I won't complain about good news :-)

PR #1476, opened 2024-08-27
docs: improve cross-links from make_() manual page

### LINKED ISSUE #1417: Fix `@seealso` for `make_()`
The "see also" section for `make_()` does not have links.

It should list all functions from the `constructor modifier` family (without making `make_()` itself a part of this family). I'm not sure if there's a simple way to do this (other than manual listing). Can you help, @maelle ?

The same should also be listed as "see also" for `sample_()`.

maelle:
Fix #1417

To me making `make_()` part of both families has no real downside, it'll be easier to maintain the cross-links this way, and the text in the rest of the manual page is clear enough.

### DIFF
--- a/R/make.R
+++ b/R/make.R
@@ -790,8 +790,6 @@ graph.atlas <- function(n) { # nocov start
 #'
 #' @param ... Parameters, see details below.
 #'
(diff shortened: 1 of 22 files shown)


maelle:
Hey I'll tweak the name of the section so it does not say "Other".

PR #1478, opened 2024-08-27
chore: add some argument checking to `add_shape()`

maelle:
Just removing a TODO :-)

### DIFF
--- a/R/plot.shapes.R
+++ b/R/plot.shapes.R
@@ -358,11 +358,35 @@ shape_noplot <- function(coords, v = NULL, params) {
 add_shape <- function(shape, clip = shape_noclip,
                       plot = shape_noplot,
                       parameters = list()) {
-  ## TODO
-  ## checkScalarString(shape)
-  ## checkFunction(clip)
-  ## checkFunction(plot)
-  ## checkList(parameters, named=TRUE)
+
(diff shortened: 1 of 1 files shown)


R/plot.shapes.R
```diff
@@ -358,11 +358,34 @@ shape_noplot <- function(coords, v = NULL, params) {
 add_shape <- function(shape, clip = shape_noclip,
                       plot = shape_noplot,
                       parameters = list()) {
-  ## TODO
-  ## checkScalarString(shape)
-  ## checkFunction(clip)
-  ## checkFunction(plot)
-  ## checkList(parameters, named=TRUE)
+
+  if (!is.character(shape)) {
```
szhorvat, line 362: Doesn't this check only that `shape` is a character, but not that it has length 1?

I don't feel very qualified to comment on such changes, since I'm not that knowledgeable about R 🙈 It's the sort of thing I'd ask you about @maelle 

Other than the above, nothing else stands out for me.

maelle: thank you!

krlmlr (approved):
Thanks, nice! We saw that we need `{.help igraph::add_shape}`, also elsewhere?

krlmlr:
Checks still fail.

PR #1480, opened 2024-08-29
ci: Bring back macOS and R devel

krlmlr:


krlmlr:
Need to install GLPK on macOS with Homebrew.

PR #1482, opened 2024-08-30
docs: add a custom roxygen2 cdocs tag

### LINKED ISSUE #1317: link C docs from R docs

maelle:
Fix #1317

- actually not sure it's better to define the tag inside the package... but at the same time it was easier.
- I need to fix the way it behaves when there are several cdocs tags for a single topic.
- Then a remaining TODO will be to add all the tags automatically for functions that are already in the package!


maelle:
although it is more complicated I think it's necessary to have a separate package with the functionality so as not to import roxygen2. I'll fix the loop here then work on getting this to work using an external package.

maelle:
Closing in favor of #1484

PR #1491, opened 2024-09-06
docs: add more cdocs tags

maelle:
Part of #1317 

It's taking a while to run so I might have to start again on Monday but it's looking good so far.

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -95,7 +95,7 @@ jobs:
         shell: bash
 
       - name: Upload R source package
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@main
         with:
           name: rigraph
           path: igraph_*.tar.gz
--- a/R/hrg.R
+++ b/R/hrg.R
@@ -305,6 +305,7 @@ hrg_tree <- function(hrg) {
 #'
 #' @family hierarchical random graph functions
 #' @export
+#' @cdocs igraph_hrg_game
 sample_hrg <- hrg_game_impl
 
 #' Predict edges based on a hierarchical random graph model
--- a/R/interface.R
+++ b/R/interface.R
@@ -554,6 +554,7 @@ get.edge.ids <- function(
 #' g <- make_ring(10)
 #' gorder(g)
 #' vcount(g)
+#' @cdocs igraph_vcount
 vcount <- function(graph) {
   as.numeric(vcount_impl(graph))
 }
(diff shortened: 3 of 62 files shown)


maelle:
mmh why can't it find a igraph entry for "igraph_eigen_adjacency"? It's listed in the YAML files.

maelle:
oooh https://github.com/igraph/igraph/issues/2142

szhorvat:
C🐶 🏷️ ?

szhorvat:
I don't have much to add here. I think this should be merged, and improvements can be done later.

Some very minor things:

 - My personal preference would be to format C function names as `igraph_foo()` instead of `igraph_foo`. But either is fine, really.
 - There are a few prominent functions that seem to be missing, such as `betweenness()`.
 - Be aware that sometimes the link goes to a generic version of a C function, while some of the documentation is present in a more specific version. Examples are e.g. `igraph_harmonic_centrality_cutoff()` vs `igraph_harmonic_centrality()`, or `igraph_personalized_pagerank()` vs `igraph_pagerank()`. This is a necessary limitation of this imperfect approach, and I wouldn't worry about it now. A practical solution is to do more cross-referencing in the C docs. Overall, I expect that one major positive effect of this work will be a lot more feedback from users about the C docs

szhorvat (approved):

maelle:
Thank you!

> My personal preference would be to format C function names as igraph_foo() instead of igraph_foo. But either is fine, really.

That's something I can fix!

> There are quite a few prominent functions that seem to be missing, such as betweenness(). When I browse functions randomly, most of the time I see no link to C docs.

`betweenness()` does not wrap an `_impl()` function hence its exclusion from this PR.

> Do you prefer the links to C docs to be in a bullet list (as they are now), or on a single compact line, similar to the "See Also" section? I am noting the difference between the two, but I don't have a clear preference.

Good idea, I'll look into this.

> Would it make sense to have this section directly above "See Also" instead of above "References"? Again, no strong preference from me.

I'm not sure this is something we can control.

> Be aware that sometimes the link goes to a generic version of a C function, while some of the documentation is present in a more specific version

It still sounds weird, I'll check.

maelle:
In https://roxygen2.r-lib.org/articles/rd-formatting.html#sections-and-subsections

> Top level sections are placed at a fixed position in the manual page, after the parameters and the details, but before \note{}, \seealso{} and the \examples{}. Their order will be the same as in the roxygen block.

maelle:
Regarding the "wrong" links for "igraph_harmonic_centrality" I only see:

```r
#' @cdocs igraph_harmonic_centrality_cutoff
harmonic_centrality <- harmonic_centrality_cutoff_impl
```

maelle:
I'll regenerate then merge.

PR #1493, opened 2024-09-09
chore: soft-deprecate `graph_()`

### LINKED ISSUE #1479: how to document graph_()?
it seems a bit different from `make_()` and `sample_()` but close. Any pointer @szhorvat?

maelle:
Fix #1479

### DIFF
--- a/R/make.R
+++ b/R/make.R
@@ -890,6 +890,14 @@ sample_ <- function(...) {
 #' graph_(cbind(1:5, 2:6), from_edgelist(directed = FALSE))
 #' graph_(cbind(1:5, 2:6), from_edgelist(), directed = FALSE)
 graph_ <- function(...) {
+  lifecycle::deprecate_soft(
+    "2.0.4",
+    "graph_()",
+    details = c(
+      "Please use constructors directly, for instance graph_from_edgelist().",
+      "graph_() will be removed in a future version of igraph."
+    )
+  )
   me <- attr(sys.function(), "name") %||% "construct"
   extracted <- .extract_constructor_and_modifiers(..., .operation = me, .variant = "graph")
   cons <- extracted$cons


szhorvat:
I'm only understanding what this function is for now.

We have `graph_from_xyz()` functions and `make_xyz()` functions. This seems to provide an alternative syntax for `make_(xyz())` to match the `graph_...` naming? Is `graph_()` actually the same as `make_()`, functionally speaking? If yes, should the deprecation message mention that? If not, what is the difference?

These appear to be interchangeable:

```r
make_(lattice(c(3,3))
graph_(lattice(c(3,3))

make_(from_literal(A -- B))
graph_(from_literal(A -- B))
```

szhorvat:
Would you be able to double-check if there's a difference between `make_` and `graph_`, @maelle ? I don't see one after a quick glance at the sources.

Here are the `make_` sources:

```r
make_ <- function(...) {
  me <- attr(sys.function(), "name") %||% "construct"
  extracted <- .extract_constructor_and_modifiers(..., .operation = me, .variant = "make")
  cons <- extracted$cons

  if (cons$lazy) {
    cons_args <- lapply(cons$args, rlang::quo_get_expr)
  } else {
    cons_args <- lapply(cons$args, rlang::eval_tidy)
  }

  res <- do_call(cons$fun, cons_args, extracted$args)
  .apply_modifiers(res, extracted$mods)
}
```

And `graph_`:

```r
graph_ <- function(...) {
  me <- attr(sys.function(), "name") %||% "construct"
  extracted <- .extract_constructor_and_modifiers(..., .operation = me, .variant = "graph")
  cons <- extracted$cons

  if (cons$lazy) {
    cons_args <- lapply(cons$args, rlang::quo_get_expr)
  } else {
    cons_args <- lapply(cons$args, rlang::eval_tidy)
  }

  res <- do_call(cons$fun, cons_args, extracted$args)
  .apply_modifiers(res, extracted$mods)
}
```

Here is one difference:

```
attr(make_, "name") <- "make_"
attr(sample_, "name") <- "sample_"
attr(graph_, "name") <- "graph_"
```

If they're identical, should we just reduce code duplication instead of deprecation?  If we still deprecate, should `make_` be mentioned in the deprecation message?  What difference does the `attr(graph_, "name") <- "graph_"` line create?

maelle:
Good points!

> If they're identical, should we just reduce code duplication instead of deprecation?

I'm not even sure `make_()` and `sample_()` should remain in the long term to be honest. **I'll have to look at usage again** but I feel that someone who uses them might achieve the same thing by doing their metaprogramming themselves (in base R or with rlang), and I also suspect very few people use them.

maelle:
`.extract_constructor_and_modifiers()` has a `variant` argument, whose value is different in `make_()` and `graph_()`.

Then it is intricate:

https://github.com/igraph/rigraph/blob/34e9e81fba8dc00f97c5b2871814be9931ae8cab/R/make.R#L691

maelle:
ah there are docs for `.variant`:

```r
#' @param .variant Constructor variant; must be one of \sQuote{make},
#'   \sQuote{graph} or \sQuote{sample}. Used in cases when the same constructor
#'   specification has deterministic and random variants.
```

krlmlr:
Thanks!

PR #1495, opened 2024-09-09
refactor: remove useless (?) attributes of `make_()`, `sample_()`, `graph_()`

maelle:
Let's see whether this breaks anything.


maelle:
it does break a test so they are useful. @szhorvat

Issue #1483, opened 2024-09-03
explore and replace for loops


maelle:
if any readability/speed can be gained.

Not urgent but it might be a good way to explore different corners of the codebase.

schochastics:
This is something that I noticed in a (admittedly niche) use case. Getting the dense adjacency matrix with a weight attribute is quite slow for larger graphs. I think this part of the code slows it down.
https://github.com/igraph/rigraph/blob/7e2c5febeeedf50f56454652031bc66ae80a57a1/R/conversion.R#L206-L236

I was using a hacky workaround to speed it up but I dont think that this would generalize well

``` r
library(igraph)
#>
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#>
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#>
#>     union
as_adj_weighted <- function(g, attr = NULL) {
    as.matrix(igraph::as_adj(g, attr = attr, type = "both", sparse = TRUE))
}


g <- sample_gnp(500, 0.2)
E(g)$weight <- runif(ecount(g))
microbenchmark::microbenchmark(
    as_adj_weighted(g, attr = "weight"),
    as_adj(g, attr = "weight",sparse = FALSE), times = 5
)
#> Unit: milliseconds
#>                                        expr        min         lq     mean
#>         as_adj_weighted(g, attr = "weight")   4.716238   4.894456 253.0044
#>  as_adj(g, attr = "weight", sparse = FALSE) 793.289684 825.885278 838.2382
#>      median        uq       max neval
#>    6.978351  34.61644 1213.8167     5
#>  850.559909 860.65557  860.8005     5
```

<sup>Created on 2024-09-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

maelle:
Thank you @schochastics!! If you ever had time to prepare a PR for that bit of the code, we'd be super grateful! 

Why do you think your solution wouldn't generalize well?

schochastics:
hmm thinking about it I guess it might just be a matter of properly propagating the type argument. I will check this and prepare a PR eventually

maelle:
thank you! no hurry and no pressure :smile_cat:

schochastics:
There are no more instances of for loops that could be sped up significantly

Issue #1137, opened 2024-01-22
Implement sparse in as_adjacency_matrix


maelle:
From #1071 
@krlmlr I am missing some context (maybe it'll come once I read the whole thread but I am writing down items for now)

szhorvat:
Sparse adjacency matrix conversion was implemented in pure R. R/igraph should use the functions from the C core instead. First step is to have _efficient_ bi-directional conversion routines for the sparse matrix data structures.

maelle:
Thank you! Then probably not for 2.0.0

szhorvat:
Something to pay attention to is that there should be no differences in behaviour after the update ... As far as I can recall, differences between the R and C implementations are:

 - The R implementations supported non-numerical matrices. Support for string matrices was dropped IIRC. 
 - The new C implementation allows choosing how to handle self-loops: once, twice or ignore. This will need to be exposed in R. Related: #1125, #1063
 - Handling of invalid inputs (e.g. creating undirected graph from non-symmetric matrix) should be handled identically between the two. I don't know if this was checked!

We should move away from using an edge attribute for weights when building matrices, as this is bad and limiting design (see e.g. #910 and related) We should use a separate weight parameter, as is now done in the C core, #906.

krlmlr:
From #1071.

maelle:
I stumbled upon this when adding tests and wondering why the `loops` argument wasn't exposed to users.

szhorvat:
Because it's new in C/igraph 0.10. I kept insisting that R/igraph 2.0 should have exposed this argument for all adjacency matrix functions. Doing this was going to be quite a bit of work though, and requires some internal refactoring.

Here's a first step: https://github.com/igraph/rigraph/pull/1063

PR #1518, opened 2024-09-22
refactor: remove for loops from weighted dense matrix creation (#1483)

schochastics:
This PR removes (nested) for loops for dense weighted matrices. Specifically:

- `get.adjacency.dense` when `attr != NULL`
- `graph.incidence.dense` when `weighted = TRUE`
- `get.incidence.dense` when `attr != NULL`

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -845,15 +845,14 @@ get.incidence.dense <- function(graph, types, names, attr) {
     recode[!types] <- seq_len(n1)
     recode[types] <- seq_len(n2)
 
-    for (i in seq(length.out = ecount(graph))) {
-      eo <- ends(graph, i, names = FALSE)
-      e <- recode[eo]
-      if (!types[eo[1]]) {
-        res[e[1], e[2]] <- edge_attr(graph, attr, i)
-      } else {
-        res[e[2], e[1]] <- edge_attr(graph, attr, i)
-      }
-    }
+    el <- as_edgelist(graph, names = FALSE)
+    idx <- types[el[, 1]]
+    el[, 1] <- recode[el[, 1]]
+    el[, 2] <- recode[el[, 2]]
+
+    tmp <- el[idx, 2:1]
+    el[idx, ] <- tmp
+    res[el] <- edge_attr(graph, attr)
 
     if (names && "name" %in% vertex_attr_names(graph)) {
       rownames(res) <- V(graph)$name[which(!types)]
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -187,6 +187,7 @@ get.adjacency.dense <- function(graph, type = c("both", "upper", "lower"),
       loops
     )
   } else {
+    # faster than a specialized implementation
     res <- as.matrix(get.adjacency.sparse(graph, type = type, attr = attr, names = names))
   }
 
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -334,7 +334,6 @@ as_adjacency_matrix <- function(graph, type = c("both", "upper", "lower"),
 as_adj <- function(graph, type = c("both", "upper", "lower"),
                    attr = NULL, edges = deprecated(), names = TRUE,
                    sparse = igraph_opt("sparsematrices")) {
-
   lifecycle::deprecate_soft("2.1.0", "as_adj()", "as_adjacency_matrix()")
 
   as_adjacency_matrix(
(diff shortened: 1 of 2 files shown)


maelle:
Thank you!! I'll start by adding some tests (from examples) for `as_adjacency_matrix()` and the other functions impacted by this PR in another/other PR(s), then we can rebase this. I have to work on tests anyway.

maelle:
I found a related issue https://github.com/igraph/rigraph/issues/1137 cc @szhorvat

R/conversion.R
```diff
@@ -183,61 +177,22 @@ get.adjacency.dense <- function(graph, type = c("both", "upper", "lower"),
...
-        for (i in seq(length.out = ecount(graph))) {
-          e <- ends(graph, i, names = FALSE)
-          res[e[1], e[2]] <- exattr[i]
-          if (e[1] != e[2]) {
-            res[e[2], e[1]] <- exattr[i]
-          }
-        }
-      }
-    }
+    res <- as.matrix(get.adjacency.sparse(graph, type = type, attr = attr, names = names))
```
krlmlr, line 190: Should we add a comment here that this is actually faster than hand-rolling a specialized implementation?

R/incidence.R
```diff
@@ -101,46 +100,27 @@ graph.incidence.dense <- function(incidence, directed, mode, multiple,
...
-          }
-        }
-      }
+
+    idx <- which(incidence != 0, arr.ind = TRUE)
+    el <- cbind(idx, incidence[idx])
+
+    el[, 2] <- el[, 2] + n1
+
+    if (!directed || mode == 1) {
```
krlmlr, line 109: This is hard to follow because at this stage in the code the meaning of `mode` is lost. Can/should we postpone the conversion of `mode` to numbers to a later stage so that we have more readable code here?

R/incidence.R
```diff
@@ -101,46 +100,27 @@ graph.incidence.dense <- function(incidence, directed, mode, multiple,
...
-            weight[ptr] <- incidence[i, j]
-            ptr <- ptr + 1
-          }
-        }
-      }
+
+    idx <- which(incidence != 0, arr.ind = TRUE)
+    el <- cbind(idx, incidence[idx])
+
+    el[, 2] <- el[, 2] + n1
```
krlmlr, line 107: Is this the same as

```r
el <- cbind(idx, incidence[idx] + n1)
```

?

In either case, this needs a bit of explanation in a comment.

R/conversion.R
```diff
@@ -825,15 +825,14 @@ get.incidence.dense <- function(graph, types, names, attr) {
...
-      if (!types[eo[1]]) {
-        res[e[1], e[2]] <- edge_attr(graph, attr, i)
-      } else {
-        res[e[2], e[1]] <- edge_attr(graph, attr, i)
-      }
-    }
+    el <- as_edgelist(graph, names = FALSE)
+    idx <- types[el[, 1]]
+    el[, 1] <- recode[el[, 1]]
+    el[, 2] <- recode[el[, 2]]
```
krlmlr, line 831: Can we do

```r
el[] <- recode(...)
```

or similar?

R/conversion.R
```diff
@@ -825,15 +825,14 @@ get.incidence.dense <- function(graph, types, names, attr) {
...
+
+    tmp <- el[idx, 2:1]
+    el[idx, ] <- tmp
```
krlmlr, line 834: Do we still need `tmp` ?

krlmlr (commented):
Thanks, great! The second commit was hard for me to understand, maybe we could create three PRs instead and review independently?

@maelle has #1539 that might improve coverage here. I'm rebasing, so let's check what the test coverage is, and adapt as needed.

R/incidence.R
schochastics: I had the same thought but just kept it as it was before. As far as I can see, the only need for the numeric value is for `R_igraph_biadjacency`. I think having the character value here would definitely make sense.

R/incidence.R
schochastics: Yes that is the same. I can refactor a bit and add comments

R/conversion.R
schochastics: yes I can adapt that

R/incidence.R
schochastics: sorry I was wrong here. It is not the same, which makes it obvious that comments are needed...

maelle:
The tests are erroring, is it the tests' or the code's fault? :sweat_smile:

schochastics:
I will have a look!

schochastics:
cc @maelle 

https://github.com/igraph/rigraph/blob/0d96334876472a8e311efc50297cd5de83dee975/tests/testthat/test-conversion.R#L175-L183

Here is an error in the test. The test graph has multiple edges and the edge weights are aggregated. E.g. the entry [3,3] is supposed to be 5.6+6.0=11.6 and not 6.0 and [4,2] is 6.1+ 3.3+4.3=13.7 and not 4.3. The aggregation makes sense and is invoked by `use.last.ij <- FALSE` in `get.adjacency.sparse()`

The later snapshot errors seem to be related with this error (I am not familiar with `expect_snapshot`). Do you want me to fix the tests in this PR, or do you do this in another branch?

maelle:
Thank you! It'd make sense to fix them in another PR to main to check they're passing for main, then we'll merge and rebase this PR. At least I think it makes sense :sweat_smile:

schochastics:
hmm they currently pass on main, so in the current igraph version, get.adjacency.dense does not aggregate multiple edge weights (gut get.adjacency.sparse does). So I guess if you change this test on main, it will fail. But IMHO this is a bug. dense and sparse should give the same result and both aggregate

schochastics:
closing this and splitting it into three separate PRs

PR #1519, opened 2024-09-23
test: add tests for as_adjacency_matrix() in test-conversion

maelle:
@szhorvat this is to prep for part of #1518, prior to my PR there were no direct tests of `as_adjacency_matrix()`.


szhorvat:
You are using randomly generated data, then testing for specific values that depend on the random seed. This is difficult to manage and it's hard to see what goes wrong with failures.

I suggest:

 - Create a _small_ specific non-random graph, which has self-loops and multi-edges. 3 or 4 vertices should suffice.
 - Create non-random edge weights. I suggest that these should be simple non-integers rounded to a single digit after the point (e.g. 1.2, 0.1, 2.0, etc.), for readability.
 - Test the whole matrix for exact equality. Do not just test specific elements. Do not test row or column sums for a specific value. Once the `loops` argument is sorted out, we can test that row/column sums equal the out/in degree (unweighted) and strength (weighted). That is not true right now, so don't test it yet.
 - Ideally, test this both for a directed and undirected instance.

This would make a complete test.

It is important to test self-loops and multi-edges, which is where things go wrong!

szhorvat:
For example,

```
dg <- make_graph(c(1,2, 2,1, 2,2, 3,3, 3,3, 3,4, 4,2, 4,2, 4,2), directed=T)
ug <- as.undirected(g, mode="each")
```

Then just add edge weights.

maelle:
Thank you!!

maelle:
@szhorvat I've implemented the changes, is this good to go?

szhorvat:
Based on a quick skim, yes!

PR #1520, opened 2024-09-23
test: add tests for `graph_from_biadjacency_matrix()`

maelle:
Also characterization tests as prep for #1518

### DIFF
--- a/tests/testthat/test-incidence.R
+++ b/tests/testthat/test-incidence.R
@@ -139,3 +139,16 @@ test_that("graph_from_biadjacency_matrix() works - sparse, modes, weighted", {
   expect_length(E(total_g), 14)
   expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
 })
+
+test_that("graph_from_biadjacency_matrix() errors well", {
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
+  colnames(inc) <- letters[1:5]
+  rownames(inc) <- LETTERS[1:3]
+
+  expect_snapshot(error= TRUE, {
+    (g <- graph_from_biadjacency_matrix(inc, weight = FALSE))
+  })
+  expect_snapshot(error = TRUE, {
+    (g <- graph_from_biadjacency_matrix(inc, weight = 42))
+  })
+})
--- a/tests/testthat/test-incidence.R
+++ b/tests/testthat/test-incidence.R
@@ -1,0 +1,141 @@
+test_that("graph_from_biadjacency_matrix() works -- dense", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
+  colnames(inc) <- letters[1:5]
+  rownames(inc) <- LETTERS[1:3]
+
+  expect_snapshot((g <- graph_from_biadjacency_matrix(inc)))
+  expect_false(is_weighted(g))
+
+  expect_snapshot((weighted_g <- graph_from_biadjacency_matrix(inc, weighted = TRUE)))
+  expect_true(is_weighted(weighted_g))
+})
+
+test_that("graph_from_biadjacency_matrix() works - dense, modes", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
+  colnames(inc) <- letters[1:5]
+  rownames(inc) <- LETTERS[1:3]
+
+  out_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "out")
+  expect_true(is_directed(out_g))
+  expect_length(E(out_g), 7)
+  expect_equal(as_adj_list(out_g, mode = "out")$A %>% as.numeric(), c(6, 7))
+
+  in_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "in")
+  expect_true(is_directed(in_g))
+  expect_length(E(in_g), 7)
+  expect_equal(as_adj_list(in_g, mode = "in")$A %>% as.numeric(), c(6, 7))
+
+  total_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "total")
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
+
+test_that("graph_from_biadjacency_matrix() works - dense, modes, weighted", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
+  colnames(inc) <- letters[1:5]
+  rownames(inc) <- LETTERS[1:3]
+
+  out_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "out", weighted = TRUE)
+  expect_true(is_directed(out_g))
+  expect_length(E(out_g), 7)
+  expect_equal(as_adj_list(out_g, mode = "out")$A %>% as.numeric(), c(6, 7))
+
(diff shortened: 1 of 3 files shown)


maelle:
Based on the coverage results https://app.codecov.io/gh/igraph/rigraph/pull/1520/blob/R/incidence.R

I need to add weights to the cases that I thought would cover the types.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,128 @@
+test_that("graph_from_biadjacency_matrix() works -- dense", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
```
maelle, line 5: @szhorvat any graph to recommend? the same example as in the other PR you've reviewed? Thank you :pray:

szhorvat: Anything will work here.

To test `multiple=TRUE`, we need a matrix which have entries greater than 1 (e.g. 2 creates two edges).

A change is coming to this function in C/igraph 1.0. Right now, when `multiple=T`, non-integer values are rounded up. After 1.0, they will be truncated (i.e. rounded now). Generally, matrix elements must be integers with `multiple=T`. It is not impossible that we will start issuing warnings for non-integer values in the future.

maelle:
@szhorvat do these characterization tests look ok to you? (needed for #1518)

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+
+  in_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "in")
+  expect_true(is_directed(in_g))
+  expect_length(E(in_g), 7)
+  expect_equal(as_adj_list(in_g, mode = "in")$A %>% as.numeric(), c(6, 7))
+
+  total_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "total")
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
```
szhorvat, line 37: Can you please use `mode = "all"` here, and rename `total_g` to `mutual_g`?

With `mode`, the strings `"all"` and `"total"` are always treated as aliases. But the terminology of "total" only makes sense when talking about degrees. In a context where we talk about edge directions, this word would never be used.  In igraph, we generally call `<->` edges "mutual".

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  in_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "in", weighted = TRUE)
+  expect_true(is_directed(in_g))
+  expect_length(E(in_g), 7)
+  expect_equal(as_adj_list(in_g, mode = "in")$A %>% as.numeric(), c(6, 7))
+
+  total_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "total", weighted = TRUE)
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
```
szhorvat, line 62: Same comment for total -> all.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  in_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "in")
+  expect_true(is_directed(in_g))
+  expect_length(E(in_g), 7)
+  expect_equal(as_adj_list(in_g, mode = "in")$A %>% as.numeric(), c(6, 7))
+
+  total_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "total")
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
```
szhorvat, line 116: Same comment for total -> all.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  in_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "in", weighted= TRUE)
+  expect_true(is_directed(in_g))
+  expect_length(E(in_g), 7)
+  expect_equal(as_adj_list(in_g, mode = "in")$A %>% as.numeric(), c(6, 7))
+
+  total_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "total", weighted= TRUE)
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
```
szhorvat, line 141: Same comment for total -> all.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
+  colnames(inc) <- letters[1:5]
+  rownames(inc) <- LETTERS[1:3]
+
+  out_g <- graph_from_biadjacency_matrix(inc, directed = TRUE, mode = "out", weighted = TRUE)
+  expect_true(is_directed(out_g))
+  expect_length(E(out_g), 7)
+  expect_equal(as_adj_list(out_g, mode = "out")$A %>% as.numeric(), c(6, 7))
```
szhorvat, line 51: Just wondering why you are testing in terms `as.numeric`. Do we have a good way to compare through vertex names, and test that these are `c` and `d`? If not, should such a way be added?

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
+
+test_that("graph_from_biadjacency_matrix() works - dense, modes, weighted", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
```
szhorvat, line 44: If it's weighted, then let's add diverse weights. Now all weights are just 1. Maybe just use `sample(0:2)`, or whatever way works to still have a lot of zeros, but have weights different from 1 as well.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  rownames(inc) <- LETTERS[1:3]
+
+  expect_snapshot((g <- graph_from_biadjacency_matrix(inc)))
+  expect_false(is_weighted(g))
+
+  expect_snapshot((weighted_g <- graph_from_biadjacency_matrix(inc, weighted = TRUE)))
+  expect_true(is_weighted(weighted_g))
+})
+
+test_that("graph_from_biadjacency_matrix() works -- sparse + multiple", {
```
szhorvat, line 80: Good to have a `multiple` test for the sparse case. There should be one for the dense case as well.

tests/testthat/test-incidence.R
maelle: It was just easier, while ignoring "+ 2/8 vertices, named, from". Otherwise I could have used a snapshot test but I preferred inlining.

tests/testthat/test-incidence.R
```diff
@@ -0,0 +1,154 @@
...
+  expect_true(is_directed(total_g))
+  expect_length(E(total_g), 14)
+  expect_equal(as_adj_list(total_g, mode = "all")$A %>% as.numeric(), c(6, 6, 7, 7))
+})
+
+test_that("graph_from_biadjacency_matrix() works - sparse, modes, weighted", {
+  local_igraph_options(print.id = FALSE)
+  withr::local_seed(42)
+
+  inc <- matrix(sample(0:1, 15, repl = TRUE), 3, 5)
```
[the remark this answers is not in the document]
maelle: #1544

Issue #1340, opened 2024-04-09
Custom code style


krlmlr:
For graph literals.

https://github.com/igraph/rigraph/pull/673 contains the logic that was used in a one-off run. It would be nice to have the code styled on a regular basis, e.g., in a pre-commit hook.

maelle:
In the PR, there are two scripts defining the same function `normalize_igraph_arrows()`.

maelle:
I'd suggest:

- creating a package holding the custom style (like https://github.com/ropensci-review-tools/spaceout)
- referencing it in the contributing guide. Regular contributors would install and run it locally.
- creating and tweaking a PR command workflow https://github.com/r-lib/actions/blob/473c68190595b311a74f208fba61a8d8c0d4c247/examples/pr-commands.yaml#L51

Does that sound good @krlmlr?

krlmlr:
Can that be a function in the igraph package too?

maelle:
I suppose it could, but I find it cleaner this way. https://github.com/igraph/igraph.r2cdocs is a separate package for instance.

If it is a function in the igraph package, could we export it? Styling graphs might be relevant for igraph users too.

krlmlr:
Yeah, let's start with a separate package. Merging is easier than tearing apart.

We have infrastructure to run styler on every PR, need to understand how to make it work with that new package or style guide. No PR command workflow needed.

maelle:
@krlmlr are all these snaps correct? https://github.com/igraph/igraph.style/blob/main/tests/testthat/_snaps/igraph-style.md

In the PR you had linked to, I think all versions of `normalize_igraph_arrows()` were the same but they contained leftover `browser()`.

krlmlr:
- Needs a space before the RHS: `A +-+ B` etc.
- Also: `A -+ B:C`

maelle:
Right, so does it mean you had used other code in your PR? I'm confused.

maelle:
oh maybe the functions aren't the same (I had tried to compare them with version control, maybe I pasted the wrong one :melting_face: )

maelle:
RHS problem fixed but did the styling of `styler::style_text("A - +B:C", style = igraph_style)` ever work?

krlmlr:
Might not have worked ever, or styler changed in the meantime.

maelle:
Do you still remember enough about the code to fix {igraph.style} yourself or should I try?

krlmlr:
[...]

PR #673, opened 2023-02-13
WIP: styler infrastructure

krlmlr:
This isn't meant to be merged yet. Contains code I used for formatting the R code in the repository, could be enhanced to continuous formatting.

PR #1523, opened 2024-09-23
chore: use `"parent"` instead of `"father"` in `bfs()` and `dfs()`

### LINKED ISSUE #880: Use 'parent' instead of 'father'
The `father` vector, used by the `dfs` and `bfs` functions, should be renamed to `parent` to line up with standard nomenclature. Doing so will be breaking change, so some mitigation may be necessary.

maelle:
Fix #880

It's not breaking anything as the functions still also return the "father" field.


R/structural.properties.R
```diff
@@ -293,12 +293,13 @@ graph.knn <- function(graph, vids = V(graph), mode = c("all", "out", "in", "tota
 #'
 #' `graph.dfs()` was renamed to `dfs()` to create a more
 #' consistent API.
+#' @param father Logical scalar, whether to return the father of the vertices.
```
maelle, line 296: we can't inherit the function definition anymore as we are not adding the new argument to the old function. 🤪

szhorvat (approved):
This just renames `father` to `parent`, right? I don't think my review is needed here (I wanted to do this for a long time).

maelle:
This way you saw it's not even a breaking change :smile_cat:

R/structural.properties.R
```diff
@@ -2214,6 +2226,9 @@ bfs <- function(
     res$dist[is.nan(res$dist)] <- -3
   }
 
+  # Remove this later?
```
krlmlr, line 2229: Reference: #1576 .

R/structural.properties.R
```diff
@@ -2310,9 +2328,11 @@ bfs <- function(
 #'
 dfs <- function(graph, root, mode = c("out", "in", "all", "total"),
```
krlmlr, line 2329: Use the opportunity to add an ellipsis after `mode` ? Also with `bfs()` ?

R/structural.properties.R
```diff
@@ -2150,6 +2155,13 @@ bfs <- function(
     }
   }
 
+  if (lifecycle::is_present(father)) {
+    lifecycle::deprecate_warn("2.0.4", "bfs(father)", "bfs(parent)")
```
krlmlr, line 2159: The version could use an update.

krlmlr, line 2159: Surprisingly, this code works, but the following is a bit more correct because `parent` and `father` are argument names:

```suggestion
    lifecycle::deprecate_warn("2.0.4", "bfs(father = )", "bfs(parent = )")
```

tests/testthat/test-structural.properties.R
```diff
@@ -188,15 +189,15 @@ test_that("bfs() works", {
...
       pred = TRUE,
       succ = TRUE,
       dist = TRUE
     )
   })
 })
 
-test_that("bfs() deprecated argument", {
+test_that("bfs() deprecated arguments", {
   g <- graph_from_literal(a - +b - +c, z - +a, d)
```
krlmlr, line 201: #673 would take care of the style here. Can we somehow add this to the package (with `.Rbuildignore`) and configure/document the usage?

krlmlr: https://github.com/igraph/rigraph/issues/1340

krlmlr (commented):
Thanks!

R/structural.properties.R
szhorvat: makes sense!

krlmlr:
Running checks now.

maelle:
https://github.com/mhahsler/pomdp/pull/18

krlmlr:
@Antonov548: can you please resolve the merge conflicts in `src/rin*.c` ?

krlmlr:
Resolved conflicts.

maelle:
I quickly "resolved the conflict" with the goal to then really fix it by running tests locally.

Now I cannot load the package any more?!

maelle:
I think we could merge this.

maelle:
I guess we can merge?

krlmlr:
Thanks!

PR #1527, opened 2024-09-24
chore!: deprecation of `scale` param of `eigen_centrality()`

### LINKED ISSUE #1353: Remove of `scale` parameter of `eigen_centrality()`
In C/igraph 1.0, the `scale` parameter of `eigen_centrality()` will be removed and the function will always behave as if `scale=TRUE` were used. Details are at https://github.com/igraph/igraph/issues/2446

This issue is to start preparing for this change on the R side, probably by deprecating the `scale` parameter.

Note that no real functionality is lost: `scale=FALSE` doesn't guarantee anything about any kind of scaling/normalization, while `scale=TRUE` uses a specific normalization. People should never have relied on `scale=FALSE`, so warning those who do soon is a good thing.

CC @maelle

maelle:
Fix #1353

But waiting on feedback there.

I also used this as an opportunity to align test files with scripts, but in separate commits. :innocent:


szhorvat:
To clarify, what will happen is that this parameter will be removed in C/igraph 1.0.

A more friendly way to deal with this in R is a multi-stage deprecation:

 - In the next R/igraph version based on C/igraph 0.10, people who use `scale=F` should get a warning, but people who use `scale=T` shouldn't.
 - In the first R/igraph version based on C/igraph 1.0, people who use `scale=F` should get an error. People who use `scale=T` should get a warning.
 - In a subsequent version, `scale` should be removed entirely.
 
Maybe this is difficult and not worth the time, so it also works that any explicit use of `scale` should get a warning right away, and in the first R/igraph based on C/igraph 1.0, we remove `scale` entirely.  This second option is what you seem to be going with, which is probably the best choice (saves us time).

I just wanted to be explicit about the options. I leave the choice to you :-)

R/centrality.R
```diff
@@ -952,9 +952,8 @@ eigen_defaults <- function() {
 #' @param graph Graph to be analyzed.
 #' @param directed Logical scalar, whether to consider direction of the edges
 #'   in directed graphs. It is ignored for undirected graphs.
-#' @param scale Logical scalar, whether to scale the result to have a maximum
-#'   score of one. If no scaling is used then the result vector has unit length
-#'   in the Euclidean norm.
+#' @param scale `lifecycle::badge("deprecated")` Normalization will always take
+#' place.
```
szhorvat, line 956: There is more than one way to normalize, so let's not simply call this normalization.

```suggestion
#' @param scale `lifecycle::badge("deprecated")` The result is always scaled to have
#' a maximum score of one.
```

I only see the comment "If no scaling is used then the result vector has unit length in the Euclidean norm." This wasn't actually true, so it's good to remove it ASAP.

PR #1539, opened 2024-09-30
test: test-conversion

maelle:


tests/testthat/test-conversion.R
```diff
@@ -255,3 +255,151 @@ test_that("as_adjacency_matrix() works -- dense + not both", {
...
+
+  expect_true(is_named(g))
+
+  I2 <- as_biadjacency_matrix(g)
+  expect_equal(I, I2, ignore_attr = TRUE)
+  expect_identical(rownames(I2), c("a", "b", "c", "d", "e", "f", "g"))
+  expect_identical(colnames(I2), c("h", "i", "j", "k", "l"))
+})
+
+test_that("as_biadjacency_matrix() works -- dense + attribute", {
```
maelle, line 381: @szhorvat does this test make any sense? It's also needed for #1518

maelle:
See also #1540

tests/testthat/test-conversion.R
```diff
@@ -255,3 +255,151 @@ test_that("as_adjacency_matrix() works -- dense + not both", {
...
+    V(ring)$score <- LETTERS[1:3]
+    as_long_data_frame(ring)
+
+    E(ring)$info <- 3:1
+    as_long_data_frame(ring)
+  })
+})
+
+test_that("as_biadjacency_matrix() works -- dense", {
+  I <- matrix(sample(0:1, 35, replace = TRUE, prob = c(3, 1)), ncol = 5)
```
szhorvat, line 360: Minor nitpick: Let's not call this `I`, as that's the standard name for the identity matrix. (It was probably called `I` because of "incidence" matrix.)

maelle: and good to merge once I fix that?

tests/testthat/test-conversion.R
szhorvat: The C core will not support adjacency matrices that contain other values than numbers. This means that we either need to drop support for character adjacency matrices, or we need to implement this behaviour in pure R while making sure to match the C core.

There are several things that don't make much sense with non-numerical adjacency matrices. For example, what do you do with multi-edges? How do you combine their character properties? For plain adjacency matrices, what do you do with self-loops in undirected graphs (whose value is normally doubled in the matrix)?

I'm sure we had a discussion about this in the past, and even reached out to someone who used character matrices (very few people do). But I don't remember the decision. We need to search the issue tracker.

tests/testthat/test-conversion.R
szhorvat: This part is good, and this is just an aesthetic nitpick that doesn't necessarily need to be fixed.  My main comments are in the other thread in this PR, addressing string matrices. Will open an issue for that soon.

tests/testthat/test-conversion.R
szhorvat: I suggest aiming to remove string matrix support, and not producing characterization tests for it. Issue: #1542

maelle: @schochastics for info if you touch these test files too

schochastics: Thanks, I unfortunately did not see this PR before because the tests were changed in #1654 and #1701. But I will get back to this specific issue here when dealing with #1542. Then we will need to add these type of tests

maelle: sorry!!

schochastics:
closing because this has been included in #1654 and #1701. Remaining issue will be solved with #1542

PR #1541, opened 2024-09-30
test: test-cliques

maelle:
moving all related tests into test-cliques and refactoring just a bit at the same time. not ready.


szhorvat:
I didn't look at any code here, I just wanted to note that the independent vertex sets of $G$ are the same as the cliques of the complement of $G$. See `complementer()`. This may be an extra way to test.

Issue #1158, opened 2024-01-23
p-value missing for `fit_power_law()`


krlmlr:
### What happens?

`fit_power_law()` no longer returns a `KS.p` component.

## igraph 1.6.0

``` r
options(conflicts.policy = list(warn = FALSE))
library(igraph)
g <- sample_pa(1000) # increase this number to have a better estimate
d <- degree(g, mode = "in")
fit_power_law(d + 1, 10)
#> $continuous
#> [1] FALSE
#> 
#> $alpha
#> [1] 2.738748
#> 
#> $xmin
#> [1] 10
#> 
#> $logLik
#> [1] -62.25711
#> 
#> $KS.stat
#> [1] 0.1037705
#> 
#> $KS.p
#> [1] 0.986668
```

<sup>Created on 2024-01-23 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

## igraph 2.0.0

``` r
options(conflicts.policy = list(warn = FALSE))
library(igraph)
g <- sample_pa(1000) # increase this number to have a better estimate
d <- degree(g, mode = "in")
fit_power_law(d + 1, 10)
#> $continuous
#> [1] FALSE
#> 
#> $alpha
#> [1] 3.226469
#> 
#> $xmin
#> [1] 10
#> 
#> $logLik
#> [1] -63.90509
#> 
#> $KS.stat
#> [1] 0.07306791
```

<sup>Created on 2024-01-23 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

bockthom:
Are there any news on this issue? I have seen that you declared this bug as "breaking change" of igraph 2.0.0 in your `NEWS.md` ("`fit_power_law()` no longer returns a `KS.p` component.")

However, without `KS.p` component, the function `fit_power_law()` is useless, at least, in the situations in which I need to use this function for power-law fitting. So, declaring this a "breaking change" is not a solution to the problem. Did you already investigate what had happened and why the component is missing now? Could you please fix this problem as soon as possible in your next release? I guess I am not the only user who needs the `KS.p` component of power-law fitting. 
Thanks in advance!

krlmlr:
The underlying C core function no longer returns this component. My attempts to use the new `igraph_plfit_result_calculate_p_value()` didn't give satisfactory results, so I gave up.

@jhollway: You mentioned in https://github.com/stocnet/migraph/issues/284 that you could compute the p-value yourself. Would you like to share a pointer?

[...]

PR #1546, opened 2024-10-04
feat: Add `p.value` and `p.precision` arguments to `fit_power_law()` to control the computation of the p-value

szhorvat:
…led in `fit_power_law()` with the plfit method

This PR is to help make quicker progress on #1158. It is not finished: a decision must be made on the interface of requesting a p-value calculation. Please see https://github.com/igraph/rigraph/issues/1158#issuecomment-2393117859

Can someone please take this over from here?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,6 @@ igraph.Rcheck/
 inst/doc
 cran
 /CRAN-SUBMISSION
-docs
+/docs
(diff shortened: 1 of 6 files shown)


szhorvat:
@krlmlr Thanks for taking this over! Let me know if you need any more help.

PR #1556, opened 2024-10-16
Prepare for libxml2 depending on bcrypt, use pkg-config.

kalibera:
Some upcoming version of Rtools will ship with libxml2 which will depend on bcrypt. This patch updates linking on Windows for that. Also, the patch conditionally uses pkg-config (when available, which it is in current version of Rtools) to establish the dependencies so that such upgrades in the future should not require changes in the package.


krlmlr:
Thanks!

PR #1559, opened 2024-10-21
chore: update arpack to arpack-ng 3.9.1

szhorvat:
@krlmlr You asked to update ARPACK to the latest ARPACK-NG. This PR attempts to do that. It was quite painful: first we need to rename functions to have an igraph prefix, and avoid conflicts. This can be done automatically. But then the line lengths exceed 72 characters, which is the maximum for fixed-form Fortran, and we need to manually reformat them, with the appropriate continuation character. I hope I didn't mess this up.

The changes to snapshots are fine. But there's a failure with spectral clustering, and it's unclear to me whether that's a real failure or not. I'll have another look at this when teaching is over (mid-November).


szhorvat:
@krlmlr I am quite certain that this is _better_ than the old version, despite the test failures. That said, the tests need to be dealt with carefully before going ahead, and I won't have time for this before 2.2.0. 

However, running revdepchecks for this would be quite useful, to test if anything in the real world was actually broken by this change. I'm not sure if that's possible before fixing _our_ tests @krlmlr .

krlmlr:
Running revdepchecks now.

krlmlr:
The results don't look terrible, comparing with the main branch now.

schochastics:
> Are you okay with merging this, assuming we can fix the tests?

Just checked locally. The error occurs at random. Need to have a closer look, but I think it should be fine to merge?

schochastics:
I can give it a try

schochastics:
@szhorvat [errr.csv](https://github.com/user-attachments/files/21962392/errr.csv)

for reference (and if we need more):

```r
gList <- list()
k <- 0
for (i in 1:5000) {
  g_rand <- sample_gnm(sample(20:40,1), sample(10:40, 1))
  lec <- tryCatch(cluster_leading_eigen(g_rand), error = function(e) NULL)
  if(is.null(lec)){
    k <- k+1
    gList[[k]] <- g_rand
  }
}
el <- lapply(gList, as_edgelist)

outfile <- "errr.csv"

con <- file(outfile, open = "wt")

for (i in seq_along(el)) {
  write.table(
    el[[i]],
    file = con,
    sep = ",",
    row.names = FALSE,
    col.names = FALSE
  )

  # Add a blank line between entries (except after last one)
  if (i < length(el)) {
    writeLines("", con)
  }
}

close(con)

```

szhorvat:
Thanks @schochastics, I am able to repro the issues with `cluster_leading_eigen()` within the C core. This also gives me confidence that this PR is fine.

Do you thin you could look at a _different_ issue in #2092 and give me examples there as well?

schochastics:
oh boy, this looks wild. but I THINK it should be correct now (waiting for checks) but needs to be reviewed

PR #1565, opened 2024-10-30
fix: Fix undefined behavior in `sample_motifs(sample = NULL)`

krlmlr:
Closes #1564.

PR #1567, opened 2024-10-31
chore: Adapt handling of optional parameters to interface definition changes in the C core

szhorvat:
This PR contains two changes:

 - General cleanup in the auto-generated code
 - Adapt to changed in this PR on the C side: https://github.com/igraph/igraph/pull/2690

Both PRs should be merged at the same time.

There is some duplicate NULL-check ugliness that is not harmful, but should be cleaned up. I think this needs to be fixed directly in Stimulus. Can you help with this @Antonov548 ?  Example: `(Rf_isNull(weights) ? 0 : (Rf_isNull(weights) ? NULL : &c_weights))`

I'm marking this as draft so it doesn't get accidentally merged before we all agree, but from my side it's good to go.

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -793,7 +793,7 @@ get_k_shortest_paths_impl <- function(graph, from, to, ..., k, weights=NULL, mod
   res
 }
 
-get_widest_path_impl <- function(graph, from, to, weights, mode=c("out", "in", "all", "total")) {
+get_widest_path_impl <- function(graph, from, to, weights=NULL, mode=c("out", "in", "all", "total")) {
   # Argument checks
   ensure_igraph(graph)
   from <- as_igraph_vs(graph, from)
@@ -826,7 +826,7 @@ get_widest_path_impl <- function(graph, from, to, weights, mode=c("out", "in", "
   res
 }
 
-get_widest_paths_impl <- function(graph, from, to=V(graph), weights, mode=c("out", "in", "all", "total")) {
+get_widest_paths_impl <- function(graph, from, to=V(graph), weights=NULL, mode=c("out", "in", "all", "total")) {
   # Argument checks
   ensure_igraph(graph)
   from <- as_igraph_vs(graph, from)
@@ -856,7 +856,7 @@ get_widest_paths_impl <- function(graph, from, to=V(graph), weights, mode=c("out
   res
 }
 
-widest_path_widths_dijkstra_impl <- function(graph, from=V(graph), to=V(graph), weights, mode=c("out", "in", "all", "total")) {
+widest_path_widths_dijkstra_impl <- function(graph, from=V(graph), to=V(graph), weights=NULL, mode=c("out", "in", "all", "total")) {
   # Argument checks
   ensure_igraph(graph)
   from <- as_igraph_vs(graph, from)
@@ -878,7 +878,7 @@ widest_path_widths_dijkstra_impl <- function(graph, from=V(graph), to=V(graph), 
   res
 }
 
-widest_path_widths_floyd_warshall_impl <- function(graph, from=V(graph), to=V(graph), weights, mode=c("out", "in", "all", "total")) {
+widest_path_widths_floyd_warshall_impl <- function(graph, from=V(graph), to=V(graph), weights=NULL, mode=c("out", "in", "all", "total")) {
   # Argument checks
   ensure_igraph(graph)
   from <- as_igraph_vs(graph, from)
--- a/tools/stimulus/functions-R.yaml
+++ b/tools/stimulus/functions-R.yaml
@@ -400,12 +400,35 @@ igraph_get_k_shortest_paths:
         edge_paths: epaths
 
 igraph_get_widest_path:
+    PARAMS: |-
+        GRAPH graph,
+        OPTIONAL OUT VERTEX_INDICES vertices, OPTIONAL OUT EDGE_INDICES edges,
+        VERTEX from, VERTEX to, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT
+    DEPS: |-
+        from ON graph, to ON graph, weights ON graph, vertices ON graph, edges ON graph
 
 igraph_get_widest_paths:
+    PARAMS: |-
+        GRAPH graph,
+        OPTIONAL OUT VERTEXSET_LIST vertices, OPTIONAL OUT EDGESET_LIST edges,
+        VERTEX from, VERTEX_SELECTOR to=ALL, EDGEWEIGHTS weights=NULL,
+        NEIMODE mode=OUT, OPTIONAL OUT VECTOR_INT parents,
+        OPTIONAL OUT VECTOR_INT inbound_edges
+    DEPS: |-
+        from ON graph, to ON graph, weights ON graph, vertices ON graph,
+        edges ON graph
 
 igraph_widest_path_widths_dijkstra:
+    PARAMS: |-
+        GRAPH graph, OUT MATRIX res, VERTEX_SELECTOR from=ALL,
+        VERTEX_SELECTOR to=ALL, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT
+    DEPS: from ON graph, to ON graph, weights ON graph
 
 igraph_widest_path_widths_floyd_warshall:
+    PARAMS: |-
+        GRAPH graph, OUT MATRIX res, VERTEX_SELECTOR from=ALL,
+        VERTEX_SELECTOR to=ALL, EDGEWEIGHTS weights=NULL, NEIMODE mode=OUT
+    DEPS: from ON graph, to ON graph, weights ON graph
 
 igraph_spanner:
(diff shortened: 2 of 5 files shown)


szhorvat:
The CI failures are due to the issue fixed in #1565. I'll rebase this once that PR is merged.

szhorvat:
> There is some duplicate NULL-check ugliness that is not harmful, but should be cleaned up. I think this needs to be fixed directly in Stimulus. Can you help with this @Antonov548 ? Example: `(Rf_isNull(weights) ? 0 : (Rf_isNull(weights) ? NULL : &c_weights))`

I was wrong, this needed to be fixed in the yaml interface file, and is now fixed.

I think this is good to go.

szhorvat:
#1569 shows that tests will pass after merging #1568.

szhorvat:
The C side PR was merged, which means that this also needs to be merged. It is now updated to what is likely to become C/igraph 0.10.14.

I'd like to have this merged soon as it is a blocker for further improvements I want to make.

Thus, merge #1568, then merge this, then ping me.

krlmlr:
CI/CD is failing here, https://github.com/igraph/rigraph/actions/runs/11681694081/job/32527681296?pr=1567#step:9:184 :


```
  Running examples in ‘igraph-Ex.R’ failed
  The error most likely occurred in:
  
  > base::assign(".ptime", proc.time(), pos = "CheckExEnv")
  > ### Name: count_motifs
  > ### Title: Graph motifs
  > ### Aliases: count_motifs
  > 
  > ### ** Examples
  > 
  > g <- sample_pa(100)
  > motifs(g, 3)
   [1]  NA  NA 218  NA  84   0   0   0   0   0   0   0   0   0   0   0
  > count_motifs(g, 3)
  [1] 302
  > sample_motifs(g, 3)
  Error in sample_motifs(g, 3) : 
    At vendor/cigraph/src/misc/motifs.c:604 : The number of vertices to use as a sample for motif count estimation must be at least one, got 0. Invalid value
  Execution halted
```

Haven't looked closer, but is this anticipated?

szhorvat:
As I said, that's fixed in #1568.

krlmlr:
I stacked and will review later today.

R/aaa-auto.R
```diff
@@ -826,7 +826,7 @@ get_widest_path_impl <- function(graph, from, to, weights=NULL, mode=c("out", "i
   res
 }
 
-get_widest_paths_impl <- function(graph, from, to=V(graph), weights=NULL, mode=c("out", "in", "all", "total")) {
+get_widest_paths_impl <- function(graph, from, to=V(graph), weights, mode=c("out", "in", "all", "total")) {
```
krlmlr, line 829: The `weights=NULL` seems useful, let's keep it.

tools/stimulus/functions-R.yaml
```diff
@@ -572,7 +572,7 @@ igraph_hub_and_authority_scores:
     # in the argument list.
     DEPS: weights ON graph, hub ON graph V(graph), authority ON graph V(graph)
     PARAMS: GRAPH graph, OUT VERTEX_QTY hub, OUT VERTEX_QTY authority,
-        OUT REAL value, BOOLEAN scale=True, EDGEWEIGHTS weights=NULL,
+        OUT REAL value, BOOLEAN scale=True, OPTIONAL EDGEWEIGHTS weights,
```
krlmlr, line 575: Let's keep:

```suggestion
        OUT REAL value, BOOLEAN scale=True, OPTIONAL EDGEWEIGHTS weights=NULL,
```

szhorvat:
C core now updated to the very latest. This is basically at the 0.10.15 C release.

krlmlr (commented):
I'll split this PR. Do I understand correctly that the CRAN failure is fixed with #1568 and that this could wait until after the CRAN release?

R/aaa-auto.R
szhorvat: I'm going to have to check out today, but in brief:

 - This function _requires_ weights, passing `NULL` in C is not valid.
 - Because there's a high risk that people will pass `NULL`, the C function does check for it, and reports a nice error (instead of crashing).
 - In R, the default _should_ be `NULL` (R's `NULL` this time) because this signifies that weights should be taken from edge attributes. It doesn't mean "use no weights". That's `NA`.
 - This function is _not_ exposed in R at the moment.
 - When it gets exposed, the quick solution to deal with this is what you are doing here: override the interface definition in R and add that `=NULL`. This shouldn't be done in C, only in R. But a nicer solution would be to patch Stimulus to treat `weights` specially, for multiple reasons, including this one. Then it would know that R needs the default of `NULL` even when this is not an optional parameter in C.

tools/stimulus/functions-R.yaml
szhorvat: I think you should not re-add `OPTIONAL`, only `NULL`. At this moment, it won't make a difference.

szhorvat: Fixed in last commit.

szhorvat:
> I'll split this PR. Do I understand correctly that the CRAN failure is fixed with #1568 and that this could wait until after the CRAN release?

Somehow I missed this comment.

Yes, this is correct.

I still suggest including this in the release, since it's not possible to pick up the latest C core fixes without it.

krlmlr:
There now is a route towards merging this and the other open PRs.

For branches that start with `each-`, each commit is checked with CI/CD. This is important to me to support `git bisect` in case we need to later figure out if a regression happened: this really only works when each intermediate commit can be built and tested. While we *could* squash-merge this PR "as is", it's still too big for my taste.

I have started a branch with the next 20 commits: https://github.com/igraph/rigraph/compare/each-vendor . The proposed process is now:

- find the first commit where the checks fail
- port what's necessary (and only that) from this PR to the branch
- rebase-interactive to combine that change with the vendoring
- force-push, rinse and repeat

Once we're on par with the C core, I'll implement a process that vendors commits as they come, one by one.

CC @maelle @schochastics .

krlmlr:
Making progress, already using two commits from this PR by necessity: https://github.com/igraph/rigraph/compare/main...refs/heads/each-vendor .

krlmlr:
We're almost on par with the current C core, a little glitch with the plfit update: https://github.com/igraph/igraph/commit/2cdc20a37fe048143ac54a3fa8288bb017402e7c#r150944398 .

Do we plan a v0.10.16 release for the C core?

szhorvat:
> Do we plan a v0.10.16 release for the C core?

I would like to ask for one soon, after handling a couple of C core PRs, but it's @ntamas who makes the relesaes.

krlmlr:
I now expect all intermediate commits of https://github.com/igraph/rigraph/compare/each-vendor to be green. Also, igraph 2.1.3 is on CRAN.

@szhorvat: would you like to rebase this branch on `each-vendor` ? A variant of the latter will soon land in the mainline (commit by commit).

szhorvat:
@krlmlr I need some help. Is `sources.mk` generated, and if yes, how?

Antonov548:
> @krlmlr I need some help. Is `sources.mk` generated, and if yes, how?

Yes, it's generated. There is step in `Makefile-cigraph`. You can use `rconfigure.py` python script in root which generates `sources.mk`.

szhorvat:
@krlmlr The rebase is done, and the target branch in GitHub is changed. I hope I didn't mess up anything.

szhorvat:
Why are tests not running if the target branch is not `main`?

krlmlr:
Good question. I only rarely use stacked PRs like this. We can retarget after applying the vendoring branch. Do the tests work for you locally?

szhorvat:
Yes, `devtools::test()` pass.

I still have some trouble running the full check, including examples, in the same way as on CI, so I can't quite try that yet.

krlmlr:
Thanks, this seems good enough for now.

krlmlr:
The mainline is now in sync with igraph/C.

With the new history, are there now commits that you would group together?

Ideally, the tests should be representative of the health of the code -- no need to run examples etc. typically.

krlmlr:
I can't find the UI to retarget the PR, this is now good to go to the main branch.

szhorvat:
I did the retargeting just now.

szhorvat:
> With the new history, are there now commits that you would group together?

I'm sorry about this being such a mess. 

I'm thinking the whole thing might as well be squashed, as there's little value in keeping these commits separate. The changes I was making are not independent. E.g. removing some of these null checks is only possible if we use `OPTIONAL` instead of `=NULL`.

krlmlr:
Thanks, great! This PR is really nice now.

My goal is to have the mainline bisect-able -- each intermediate commit should pass checks. In case we later see that this commit has created a regression, we can still split it. The PR would now be splittable, but much less so when it was intermingled with all the changes in the C core. There's no need to invest up front. The only reason to do so is to learn more about the changes now, I trust you there.

New changes in the C core should now also be imported automatically, commit by commit. I'll see to that. And we have a way to deal with a backlog of changes, just like that we had before today. We're in good shape!

PR #1568, opened 2024-10-31
fix: fix the incorrect handling of the `sample` parameter in `sample_motifs()` and ensure that the default `sample.size` is integer

szhorvat:
@krlmlr Please use this fix instead of the one in #1565, which doesn't seem correct. This also fixes a bad default value for the `sample.size` parameter. Recent igraph errors for non-integer values.

### DIFF
--- a/R/motifs.R
+++ b/R/motifs.R
@@ -201,6 +201,7 @@ count_motifs <- function(graph, size = 3, cut.prob = rep(0, size)) {
 #'   of the motif (the `size` argument). By default no cuts are made.
 #' @param sample.size The number of vertices to use as a starting point for
 #'   finding motifs. Only used if the `sample` argument is `NULL`.
+#'   The default is `ceiling(vcount(graph) / 10)` .
 #' @param sample If not `NULL` then it specifies the vertices to use as a
 #'   starting point for finding motifs.
 #' @return A numeric scalar, an estimate for the total number of motifs in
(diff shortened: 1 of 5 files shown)


tests/testthat/test-motifs.R
```diff
@@ -44,3 +44,19 @@ test_that("motif finding works", {
...
   expect_equal(m5 / m, c(NA, NA, 0.439985332979302, NA, 0.440288166730411, 0.346938775510204, 0.44159753136382, 0.452054794520548, NaN, 0.323076923076923, NaN, 0.347826086956522, NaN, NaN, NaN, NaN))
 })
+
+test_that("sample_motifs works", {
+  withr::local_seed(20041103)
+
+  g <- make_graph(~ A-B-C-A-D-E-F-D-C-F)
+  n <- vcount(g)
+
+  mc <- sample_motifs(g)
```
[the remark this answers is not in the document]
szhorvat: `motif_count`

krlmlr:
Thanks!

R/motifs.R
```diff
@@ -215,8 +216,13 @@ count_motifs <- function(graph, size = 3, cut.prob = rep(0, size)) {
 #' motifs(g, 3)
 #' count_motifs(g, 3)
 #' sample_motifs(g, 3)
-sample_motifs <- function(graph, size = 3, cut.prob = rep(0, size),
-                          sample.size = vcount(graph) / 10, sample = NULL) {
+sample_motifs <- function(
+  graph,
+  size = 3,
+  cut.prob = rep(0, size),
+  sample.size = NULL,
```
szhorvat, line 223: I'm a bit uncomfortable about the default being hidden behind `NULL`, as the value is actually very relevant during usage, and this default value is completely arbitrary.

IMO the ideal solution would be to force users to provide either `sample.size` or `sample` explicitly instead of relying on the default. However, that is a breaking change not suitable for this release.

szhorvat:
The CI failure is related to rgl. I think Aviator should be overridden and this should be merged.

Either there's a real memory leak in rgl (in which case this should be reported to them, but it's too much work to create a minimal example right now), or it's a LeakSanitizer false positive (in which case we need to add a suppression, which is also annoying and time consuming).

R/motifs.R
maelle: could you please open an issue for this?

krlmlr: Actually, the rationale is documented, the defaults are documented, and this will end up having a very low priority for a long time.

https://design.tidyverse.org/defaults-short-and-sweet.html#null-default

krlmlr: We can change the interface to make arguments required, even in a lifecycle-friendly way. I think I misread this as a request to put back the inline default.

We could start with a message that is shown if the argument is omitted but would have been relevant.

PR #1574, opened 2024-11-07
refactor: adapt to cut.prob's new handling of NULL in the C core (sim…

### LINKED ISSUE #1570: Change default value `cut.prob` to `NULL` in motif finding functions
The default value for the `cut.prob` parameter of motif finding functions should have a default value of `NULL` instead of `rep(0, size)`. The documentation should be changed to say that `NULL` indicates that "no cuts are made".

This does not change behaviour. It makes use of a corresponding change in the C core (https://github.com/igraph/igraph/issues/2555) and simplifies things in the long term.

Can you have a look at this @maelle, since you are doing cleanup refactoring anyway?

maelle:
…pler default for the R interface)

Fix  #1570

Work needed in the tests.


maelle:
@szhorvat does this need a new version of the C sources?

``` r
withr::local_seed(123)

b <- igraph::sample_gnp(10000, 4 / 10000, directed = TRUE)
igraph::sample_motifs(b, 3)
#> Error in igraph::sample_motifs(b, 3): At vendor/cigraph/src/misc/motifs.c:569 : Cut probability vector size (0) must agree with motif size (3). Invalid value
```

<sup>Created on 2024-11-07 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

maelle:
mmh this does not work at all currently.

maelle:
@szhorvat actually, I think things are fine. What do you think of the tests https://github.com/igraph/rigraph/blob/30518c2fa2705567202993171761c195a77a349f/tests/testthat/test-motifs.R#L1?

They're failing for small differences. Furthermore they do not make any sense to me, why are we testing for the value of the divisions?

```
── Failure ('test-motifs.R:11:3'): motif finding works ─────────────────────────
c(mno0/mno, mno1/mno, mno2/mno) (`actual`) not equal to c(0.654821903845065, 0.666289144345659, 0.668393831285275) (`expected`).

  `actual`: 0.67454 0.66614 0.66597
`expected`: 0.65482 0.66629 0.66839
```

szhorvat:
I'm really tired today ... could you please help me by showing me a specific before/after example that changes output?  Passing `c(0,0,0)` vs `NULL` should NOT change anything.

But as I'm writing this, I think I'm starting to remember what's going on:

I think passing `c(0,0,0)` vs `NULL` will cause some RNG calls to be omitted. This means that _later_ calls that use different values than `c(0,0,0)`, and therefore return stochastic results, should indeed be affected. And yes, this is not a bug, don't worry.

szhorvat:
Yes, this is certainly what's going on.  If you pass `NULL` or `c(0,0,0,...)`, the result will be the exactly same. But the RNG state will be mutated differently, which means that any subsequent uses of the RNG are affected.

The results are approximately the same and everything is fine.

Adding a tolerance won't work very well here because the noise in the results is still quite high and will continue to be high unless we use large enough graphs and small enough cut probabilities that the computation time becomes too long for a test.

szhorvat:
> Furthermore they do not make any sense to me, why are we testing for the value of the divisions?

The interface is not very nice, unfortunately, but improvements are for a later version and for the C core.

If we give cut probabilities $p_1, p_2, \dots$, then only a fraction of the motifs will be sampled. This fraction is $\prod_i (1-p_i)$. This is the value you should see in the ratio of the counts obtained with a non-zero cut probability and the full counts (with no cuts). Since _some_ motifs are rare, some entries in the result vector will fluctuate wildly.

So, if you give `c(1/3, 0, 0)`, then the ratios should all be about $1-1/3 \approx 0.66$.

szhorvat (approved):

R/motifs.R
```diff
@@ -215,22 +229,30 @@ count_motifs <- function(graph, size = 3, cut.prob = rep(0, size)) {
...
+      R_igraph_motifs_randesu_estimate, graph, as.numeric(size),
+      cut.prob, as.numeric(sample.size), as.numeric(sample)
+    )
+  } else {
+    .Call(
+      R_igraph_motifs_randesu_estimate, graph, as.numeric(size),
+      as.numeric(cut.prob), as.numeric(sample.size), as.numeric(sample)
+    )
+
+  }
```
szhorvat, line 255: Instead of putting the call in a conditional, update the value of `cut.prob` in a conditional, and keep a single call to the C function.

szhorvat:
When you resolve conflicts, be sure that you don't accidentally re-add `as.numeric` to `sample`.

maelle:
I'll make a new PR, thank you!!

R/motifs.R
```diff
@@ -242,10 +250,14 @@ sample_motifs <- function(
   }
 
   on.exit(.Call(R_igraph_finalizer))
+  if (!is.null(cut.prob)) {
+    cut.prob <- as.numeric(cut.prob)
+  }
   .Call(
     R_igraph_motifs_randesu_estimate, graph, as.numeric(size),
-    as.numeric(cut.prob), as.numeric(sample.size), sample
+    cut.prob, as.numeric(sample.size), as.numeric(sample)
```
maelle, line 258: did we want the `as.numeric(sample)` here @szhorvat?

maelle:
@szhorvat would you mind re-reviewing? thank you!

maelle:
there's a second PR which I had forgotten

PR #1577, opened 2024-11-07
chore: Add checks to generated code

krlmlr:


krlmlr:
@szhorvat: Have you seen this error before? Do you think this can be patched in a minimal way?

```
REAL() can only be applied to a 'numeric', not a 'NULL'
```

https://github.com/igraph/rigraph/actions/runs/11724389703/job/32658211635?pr=1577#step:7:41

szhorvat:
This PR is broken out of #1567. The tests of #1567 pass. Why don't you just merge #1567 instead of trying to separate this out?

As I remember, removing the `is.null` checks is only valid if `OPTIONAL foo` was used, but not if `foo=NULL` was used. With `OPTIONAL`, Stimulus will generate the `is.null` checks by itself.

krlmlr:
Merging smaller PRs simplifies the review and gives greater confidence. I think we want to strive for a looser coupling between the C library and the R package to allow for more independent development of the two.

szhorvat:
I don't think it's a good use of time, but if you want to decouple this into pieces, here's what I think can be done:

 - Adding all the `IGRAPH_R_CHECK` is independent of the interface update and can come first
 - Next comes updating the C core, which brings all the OPTIONAL instead of =NULL. This will cause a lot of duplicated/nested `is.null` checks.
 - Finally, you can remove the `is.null` checks from the yaml file.

Doing it in this clean and nice order wasn't realistic for me, as I kept discovering more and more issues while working on the codebase.

maelle:
Only not draft PR without the hackathon label, so I'm putting it on the agenda :smile_cat:

PR #1578, opened 2024-11-07
ci: Work around rgl sanitizer failure

krlmlr:


szhorvat:
Would it be better to just ignore the problem until rgl fixes it? Can Aviator be told to ignore sanitizer failures for a while?

There might be many examples using rgl.

krlmlr:
Thanks, good catch. I've added a comment that helps us remember to turn this test on again.

PR #1579, opened 2024-11-10
ci: Install rgl from GitHub for Sanitizer runs

szhorvat:
Don't merge, this PR is just for testing https://github.com/dmurdoch/rgl/issues/439

### DIFF
--- a/.github/workflows/build-and-check.yml
+++ b/.github/workflows/build-and-check.yml
@@ -253,6 +253,7 @@ jobs:
           run: |
             set -e
             printenv
+            RDcsan -e 'pak::pak("dmurdoch/rgl@issue439")'
             cd rigraph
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-check.yml
```diff
@@ -253,6 +253,7 @@ jobs:
           run: |
             set -e
             printenv
+            RDcsan -e 'pak::pak("dmurdoch/rgl@issue439")'
```
krlmlr, line 256:
```suggestion
            # Memory leaks with rgl 1.3.12, https://github.com/dmurdoch/rgl/issues/439
            RDcsan -e 'pak::pak("dmurdoch/rgl")'
```

krlmlr (commented):
Thanks, looks good. I believe we can merge a variant of this.

PR #1602, opened 2024-11-28
refactor: adapt to `cut.prob`'s new handling of NULL in the C core

### LINKED ISSUE #1570: Change default value `cut.prob` to `NULL` in motif finding functions

maelle:
Simpler default for the R interface

Fix #1570 
Fix #1574

How to handle tests?


maelle:
@szhorvat for tests, is the idea simply to

- use a fixed random seed for each of the tests that's currently failing
- record the results as the new truth?

szhorvat:
> * use a fixed random seed for each of the tests that's currently failing
> 
> * record the results as the new truth?

This would make the tests more robust to changes of the sort we made. I think it's a good approach.

maelle:
@szhorvat in that case, this is now ready for review. There were actually already fixed seeds.

szhorvat:
I'm very tired, so expect some stupid comments from me, but I'll try to do a basic review.

R/motifs.R
```diff
@@ -125,10 +126,12 @@ dyad.census <- function(graph) { # nocov start
...
 #' sample_motifs(g, 3)
-motifs <- function(graph, size = 3, cut.prob = rep(0, size)) {
+motifs <- function(graph, size = 3, cut.prob = NULL) {
   ensure_igraph(graph)
-  cut.prob <- as.numeric(cut.prob)
-  if (length(cut.prob) != size) {
+
+  if (!is.null(cut.prob)) cut.prob <- as.numeric(cut.prob)
+
+  if (!is.null(cut.prob) && length(cut.prob) != size) {
```
szhorvat, line 134: Please help me understand the code below. Is it repeating the last entry to fill up the vector until it reaches `size`?

What if it is longer than `size`? That should be an error (and I think the C core will take care of that, but I'd need to check).

szhorvat: This just errors:

```r
motifs(make_ring(10), cut.prob=0.1)
Error in motifs(make_ring(10), cut.prob = 0.1) : 
  At vendor/cigraph/src/misc/motifs.c:156 : Cut probability vector size (0) must agree with motif size (3). Invalid value
```

Why does it not trigger the `length(cut.prob) != size` code path?

I think I'm just too tired for this right now :(

R/motifs.R
```diff
@@ -39,7 +39,7 @@ graph.motifs.no <- function(graph, size = 3, cut.prob = rep(0, size)) { # nocov
 #' @inheritParams sample_motifs
 #' @keywords internal
 #' @export
-graph.motifs.est <- function(graph, size = 3, cut.prob = rep(0, size), sample.size = vcount(graph) / 10, sample = NULL) { # nocov start
+graph.motifs.est <- function(graph, size = 3, cut.prob = NULL, sample.size = vcount(graph) / 10, sample = NULL) { # nocov start
```
szhorvat, line 42: I think `sample.size` should default to `NULL` in this deprecated function as well, to match `sample_motifs`.

Do what you think is best, just drawing attention to this.

R/motifs.R
szhorvat: This code path _is_ triggered, but it produces an incorrect vector.

I know this is not your code, but can you fix it @maelle ?

My recommendation is:

 1. Either go with what this code seems to intend, i.e. repeat the last element as many times as necessary to reach `size`.
 2. Or keep things simple and not repeat anything _except_ a scalar. A scalar should be treated as a vector of identical elements of length `size`.

No preference as to which from my size (maybe 1 is a tiny bit nicer, but also overly complicated?)

Whatever you choose, can you document it? Also, can you make sure it's implemented in _all_ motif functions, not just this one?

szhorvat:
> * use a fixed random seed for each of the tests that's currently failing

I meant, the only way this issue could have been avoided is if a seed is set _before every call_ to motif functions. We have the unusual situation here where an internal change did not affect the output, but it did affect how the function propagates the random state.

Actually setting the seed before every call is overkill, and I do NOT recommend it. It's good as it is IMO.

szhorvat:
I looked at the changes in this PR and I think they're good. Thank you @maelle !

R/motifs.R
maelle: So do you mean

- error if `cut.prob` is not of length either 1 or the same as size?
- if it is of length 1, repeat it size times?

szhorvat: That is option 2.

Let me give examples for option 1 and option 2, assuming `size == 4`.

Option 1:

Input: `0.3`, interpreted as `0.3 0.3 0.3 0.3`

Input: `0.3 0.4`, interpreted as `0.3 0.4 0.4 0.4`

Input: `0.1 0.2 0.3 0.4 0.5` (longer than size 4), should trigger an error, which can be achieved by passing to the C core as-is

So, repeat the last entry to fill to size.

Option 2:

Input: `0.3`, interpreted as `0.3 0.3 0.3 0.3`

Input: `0.3 0.4`, passed to the C core as-is, which will trigger an error.

maelle: @krlmlr any opinion here?

krlmlr:
Checks fail: Codoc mismatches.

https://github.com/igraph/rigraph/actions/runs/12687999617/job/35363595837?pr=1602#step:17:240

R/motifs.R
krlmlr: This whole block just looks wrong, can we get rid of it and replace it with an error?

R/motifs.R
```diff
@@ -224,8 +231,10 @@ sample_motifs <- function(
   sample = NULL
 ) {
   ensure_igraph(graph)
-  cut.prob <- as.numeric(cut.prob)
-  if (length(cut.prob) != size) {
+
+  if (!is.null(cut.prob)) cut.prob <- as.numeric(cut.prob)
+
+  if (!is.null(cut.prob) && length(cut.prob) != size) {
```
krlmlr, line 237: Same here.

R/motifs.R
krlmlr: I propose to leave deprecated functions unchanged.

krlmlr (commented):
Do we need/want a test for the deprecated functions?

maelle:
@schochastics are you surprised that I needed to tweak expectations?

schochastics:
> @schochastics are you surprised that I needed to tweak expectations?

Yes, but it is probably due to changes in the source code? I will have a look

schochastics:
@maelle I am still confused but given that the cut off introduces non-determinism, I assume that the changes in the source code just randomly lead to a different result with the same seed. There is certainly a computer sciency explanation, but I think this intuition should suffice to trust the change in snapshot :D

szhorvat:
> I assume that the changes in the source code just randomly lead to a different result with the same seed.

Yes, that happened, it is expected.

maelle:
Thanks both! Ok to merge?

szhorvat:
@krlmlr Should we have this merged before revdepchecks?

R/motifs.R
krlmlr: Should we leave deprecated functions unchanged?

R/motifs.R
```diff
@@ -53,7 +54,7 @@ graph.motifs.est <- function(graph, size = 3, cut.prob = rep(0, size), sample.si
 #' @inheritParams motifs
 #' @keywords internal
 #' @export
-graph.motifs <- function(graph, size = 3, cut.prob = rep(0, size)) { # nocov start
+graph.motifs <- function(graph, size = 3, cut.prob = NULL) { # nocov start
```
krlmlr, line 57: Same here?

R/motifs.R
krlmlr: Still valid.

krlmlr (commented):
@maelle: Would you like to take another stab?

maelle:
@krlmlr the error had been added for all occurrences of `if (!is.null(cut.prob) && length(cut.prob) != size) {`

R/motifs.R
maelle: we did replace the block with an error

R/motifs.R
maelle: we also added an error here

krlmlr (approved):
Thanks, please resolve conflicts and merge.

Issue #1605, opened 2024-11-29
`adjacent_vertices()` is off by 1 when option `return.vs.es` is `FALSE`


stibu81:
### What happens, and what did you expect instead?

When `adjacent_vertices()` is run with the option `return.vs.es` set to `FALSE` (which is not the default), all the vertex indices are 1 too small.

`neighbors()`, on the other hand, returns correct results for both values of `return.vs.es`.

The reason is that in `neighbors()` the vertex-indices are always increased by 1, while in `adjacent_vertices`, this only happens when `return.vs.es` is `TRUE`.

### To reproduce

``` r
library(igraph)
#> 
#> Attaching package: 'igraph'
#> The following objects are masked from 'package:stats':
#> 
#>     decompose, spectrum
#> The following object is masked from 'package:base':
#> 
#>     union

g <- make_tree(13, children = 3)
V(g)$name <- paste0("V", 1:13)
plot(g, vertex.size = 35)
```

![](https://i.imgur.com/ja4T5uT.png)

``` r

# make sure the option is set to the default
igraph_options(return.vs.es = TRUE)

# in this case, all results are as expected
lapply(c(1, 4), neighbors, graph = g)
#> [[1]]
#> + 3/13 vertices, named, from 101b41a:
#> [1] V2 V3 V4
#> 
#> [[2]]
#> + 3/13 vertices, named, from 101b41a:
#> [1] V11 V12 V13
av <- adjacent_vertices(g, c(1, 4))
av
#> $V1
#> + 3/13 vertices, named, from 101b41a:
#> [1] V2 V3 V4
#> 
#> $V4
#> + 3/13 vertices, named, from 101b41a:
#> [1] V11 V12 V13
lapply(av, as.integer)
#> $V1
#> [1] 2 3 4
#> 
#> $V4
#> [1] 11 12 13

# with return.vs.es = FALSE, neighbors() still works, but adjacent_vertices()
# is 1 too small.
igraph_options(return.vs.es = FALSE)
lapply(c(1, 4), neighbors, graph = g)
#> [[1]]
#> [1] 2 3 4
#> 
#> [[2]]
#> [1] 11 12 13
av <- adjacent_vertices(g, c(1, 4))
av
#> $V1
#> [1] 1 2 3
#> 
#> $V4
#> [1] 10 11 12
lapply(av, \(i) V(g)[[i]])
#> $V1
#> + 3/13 vertices, named, from 101b41a:
#>   name
#> 1   V1
[...]

PR #1610, opened 2024-12-01
test: Test handling of `"return.vs.es"` in several functions

stibu81:
This completes the fix of #1605 and adds tests.

Since the tests of `neighbors()` were based on `as_adj_list()`, I have extended the test of the latter to also run with `return.vs.es = FALSE`. I extended the tests of `neighbors()` and added tests for `adjacent_vertices()` and `incident_edges()`.

I found another small issue related to this one that I think is a mistake. But I am not entirely sure that it is, so I would like to have confirmation before I fix it: `as_adj_edge_list()` does not respect the setting `return.vs.es` and always returns `igraph.es` objects. Could someone please confirm that this is indeed not the expected behaviour. I could then fix this and update the pull request. Or should I open a new issue for this?

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -448,6 +448,7 @@ references:
     email: dadler@dyncall.org
   year: '2024'
   doi: 10.32614/CRAN.package.rgl
+  version: '>= 1.3.14'
 - type: software
   title: rmarkdown
   abstract: 'rmarkdown: Dynamic Documents for R'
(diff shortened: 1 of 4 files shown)


stibu81:
Could you comment on the other issue I found? It could be fixed quickly if you agree that it is a mistake.

krlmlr:
Thanks, good catch, missed that. This is a new issue and a potentially breaking change, let's handle this separately.

stibu81:
Should I open an issue where this can be discussed?

krlmlr:
Please do, thank you. Chances are that other functions are affected here.

PR #1624, opened 2024-12-10
docs: add central note about experimental functions

maelle:
"Use lifecycle for all experimental functions, make a central note in the documentation." is part of the R Consortium proposal. 

Have we missed functions that should have the experimental badge? If so, how could I identify them?

Local preview:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ See the [Installation FAQ](https://r.igraph.org/articles/installation-troublesho
 
 See the [igraph package's website](https://r.igraph.org/) for the complete manual.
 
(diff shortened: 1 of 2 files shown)


szhorvat:
> Have we missed functions that should have the experimental badge? If so, how could I identify them?

I'm not sure. I wouldn't worry too much, I don't think we missed much.

If you'd like to make sure, the way to go is to check PRs from the past few months and see if we ever forgot the "experimental" marker. Anything that is "experimental" in the C core should also be experimental in R. Newly exposed functions that are NOT experimental in C may or may not be experimental in R, according to your and Kirill's judgement.

szhorvat:
I had a look through recent "feat:" PRs and I don't think we missed anything.

maelle:
Thanks so much @szhorvat!!

@krlmlr could we please merge this harmless PR?

krlmlr:
Thanks!

Issue #1576, opened 2024-11-07
Deprecate components of output lists


krlmlr:
Examples:

- [ ] `parent` vs. `father` in https://github.com/igraph/rigraph/pull/1523
- [ ] ...

We could put an ALTREP object that gives a warning when accessed. Only useful in the very long term.

maelle:
Is there an example of a package where this is implemented?

krlmlr:
I'm no longer sure ALTREP is a good solution here.

For affected functions, we can add an argument `format = "v2"`, with `"v1"` being an option that will be deprecated and defunct at some point.

krlmlr:
Or "editions" as in testthat?

maelle:
But what should be the default of the function, the new or old version? The new one I presume, which means it's still a breaking change?

PR #1626, opened 2024-12-10
docs: better describe output of `all_shortest_paths()`

### LINKED ISSUE #1029: `add_shortest_paths()` docs do not describe vpaths / epaths
The `all_shortest_paths()` docs describe the `vpath` and `epath` components in the return value, but this function actually returns `vpaths` and `epaths` (not plural).

I'm not sure what the best way is to deal with cases like this, when several related functions share a doc page. @maelle ?

maelle:
Fix  #1029

@szhorvat I'll need your help.

### DIFF
--- a/R/structural.properties.R
+++ b/R/structural.properties.R
@@ -796,8 +796,9 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
 #'   \item{epaths}{This is a list similar to vpaths,
 #'   but the vectors of the list contain the edge ids along the shortest paths,
 #'   instead of the vertex ids.}
-#'   \item{nrgeo}{Number of geodesics.}
-#'   \item{res}{}
+#'   \item{nrgeo}{A vector in which each element is the number of shortest paths
+#'   (geodesics) from `from` to the corresponding vertex in `to`.}
+#'   \item{res}{Deprecated}
 #'
 #'   For `mean_distance()` a single number is returned if `details=FALSE`,
 #'   or a named list with two entries: `res` is the mean distance as a numeric
--- a/man/closeness.Rd
+++ b/man/closeness.Rd
@@ -43,7 +43,7 @@ Numeric vector with the closeness values of all the vertices in
 \code{v}.
 }
 \description{
-Closeness centrality measures how many steps is required to access every other
+Closeness centrality measures how many steps are required to access every other
 vertex from a given vertex.
 }
 \details{
--- a/man/distances.Rd
+++ b/man/distances.Rd
@@ -146,14 +146,15 @@ for each vertex, or \code{NULL}, if it was not requested.}
 
 For \code{all_shortest_paths()} a list is returned:
 \item{vpaths}{This is a list. Each list element
-contains a shortest path from \code{from} to a vertex in \code{to}. The
-shortest paths to the same vertex are collected into consecutive elements
-of the list.}
+contains the vertices of a shortest path from \code{from} to a vertex in \code{to}.
+The shortest paths to the same vertex are collected
+into consecutive elements of the list.}
 \item{epaths}{This is a list similar to vpaths,
 but the vectors of the list contain the edge ids along the shortest paths,
 instead of the vertex ids.}
-\item{nrgeo}{Number of geodesics.}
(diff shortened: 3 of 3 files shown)


R/structural.properties.R
```diff
@@ -788,10 +788,16 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
 #'   was not requested.} \item{inbound_edges}{Numeric vector, the inbound edge
 #'   for each vertex, or `NULL`, if it was not requested.}
 #'
-#'   For `all_shortest_paths()` a list is returned, each list element
+#'   For `all_shortest_paths()` a list is returned:
+#'   \item{vpaths}{This is a list. Each list element
 #'   contains a shortest path from `from` to a vertex in `to`. The
 #'   shortest paths to the same vertex are collected into consecutive elements
-#'   of the list.
+#'   of the list.}
```
szhorvat, line 795: Let's write 

> ... contains **the vertices** of a shortest path ...

R/structural.properties.R
```diff
@@ -788,10 +788,16 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
...
+#'   \item{epaths}{This is a list similar to vpaths,
+#'   but the vectors of the list contain the edge ids along the shortest paths,
+#'   instead of the vertex ids.}
```
szhorvat, line 798: Nitpick: Let's decide whether we write "ID" or "id" and stick to it.

R/structural.properties.R
```diff
@@ -788,10 +788,16 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
...
+#'   \item{nrgeo}{Number of geodesics.}
+#'   \item{res}{}
```
szhorvat, line 800: > A vector in which each element is the number of shortest paths (geodesics) from `from` to the corresponding vertex in `to`.

szhorvat:
`$res` is for backwards compatibility. It is the same as `$vpath`. It is effectively deprecated. The deprecation should be made explicit, and it should be removed soon.

krlmlr:
@maelle: Should this be part of the upcoming 2.1.3?

R/structural.properties.R
maelle: Please add your preference in https://github.com/igraph/rigraph/issues/1683

maelle:
@krlmlr @szhorvat suggests deprecated `res`. For this we'd need #1576 I'd be curious to see you add this during a hackathon so I have a better chance of doing it myself another time.

R/structural.properties.R
```diff
@@ -788,10 +788,17 @@ degree_distribution <- function(graph, cumulative = FALSE, ...) {
 #'   was not requested.} \item{inbound_edges}{Numeric vector, the inbound edge
 #'   for each vertex, or `NULL`, if it was not requested.}
 #'
-#'   For `all_shortest_paths()` a list is returned, each list element
-#'   contains a shortest path from `from` to a vertex in `to`. The
-#'   shortest paths to the same vertex are collected into consecutive elements
-#'   of the list.
+#'   For `all_shortest_paths()` a list is returned:
+#'   \item{vpaths}{This is a list. Each list element
```
krlmlr, line 792: Do we need `\describe{}` here? See https://developer.r-project.org/parseRd.pdf .

Yes, in **R documentation (.Rd) files**, you need to use `\describe{}` if you want to use `\item{}{}`.

### Explanation:
- `\describe{}` is used to create a **description list** in R documentation.
- `\item{}{}` defines individual list items, where the **first argument** is the term being described, and the **second argument** is its description.

### Example:
```r
\describe{
  \item{param1}{This is the first parameter description.}
  \item{param2}{This is the second parameter description.}
}
```

### Alternative:
- If you don’t need a **term-description** list and just want bullet points, you can use `\itemize{}` instead:
  ```r
  \itemize{
    \item First bullet point.
    \item Second bullet point.
  }
  ```

So, **yes,** `\describe{}` is required for `\item{}{}` in `.Rd` files.

krlmlr (approved):
Thanks!

krlmlr:
@maelle: Do you want to take another stab?

Issue #1613, opened 2024-12-02
`IGRAPH_FINALLY()` and `IGRAPH_FINALLY_REAL()`


krlmlr:
CSan checks complain about our hack in `IGRAPH_FINALLY()` calling `IGRAPH_FINALLY_REAL()` with a cast function pointer: https://github.com/igraph/rigraph/actions/runs/12108577215/job/33756736066#step:6:10130 . My most recent submission got rejected from CRAN because of this.

Propose to only use `IGRAPH_FINALLY_REAL()` instead of `IGRAPH_FINALLY()` in the R interface, and create wrapper functions for each function called:

```c
# Bad:
IGRAPH_FINALLY(igraph_vector_int_destroy, &news);

# Good:
IGRAPH_FINALLY_REAL(igraph_vector_int_destroy_pv, &news);

void igraph_vector_int_destroy_pv(void* v) {
  igraph_vector_int_destroy((magic_cast*)v);
}
```

(For some `magic_cast` we'd have to figure out. I can't quite parse `igraph_vector.h` . These days, I'd write a code generator and commit the generated code to version control, instead of this complex preprocessor logic because I find that easier to read.)

@szhorvat @ntamas: What do you think?

szhorvat:
I think it is highly unreasonable to try to work around this warning, which can easily be disabled: `-fno-sanitize=function`.

We already knew that this warning existed, understand the reasons, and know that it is harmless.

I can also say right now that this is not the only UBSan warning we have.

krlmlr:
Right, I hear you. How about:

```c
    do { \
        /* the following branch makes the compiler check the compatibility of \
         * func and ptr to detect cases when we are accidentally invoking an \
         * incorrect destructor function with the pointer */ \
        if (0) { func(ptr); } \
        IGRAPH_FINALLY_REAL((func##_pv), (ptr)); \
    } while (0)
```

Note the `##_pv` in the call. Could these variants be provided by our macros?

Arguing with CRAN is an upstream battle.

szhorvat:
Let's hear from @ntamas as well.

At this point my opinion is that we need to try to talk reason into CRAN, but let's discuss the objective part below.

The change you propose (which preserves useful compiler warnings) is to duplicate every single destructor and provide a version (`pv`) that takes a `void *` pointer, then change `IGRAPH_FINALLY` to call the `pv` version.

 - This is a lot of work and a significant change.
[...]

PR #1630, opened 2024-12-11
feat: Use `_pv` destroy functions to satisfy ASAN checks

krlmlr:
Part of #1613.


krlmlr:
Current ASAN run: https://github.com/igraph/rigraph/actions/runs/12274005767/job/34246019696

krlmlr:
Both checks still fail, it's probably a different `IGRAPH_FINALLY()` we need to change. The failure is located at line 10243 of the output.

src/rinterface_extra.c
```diff
@@ -620,7 +650,7 @@ igraph_error_t R_igraph_attribute_add_vertices(igraph_t *graph, igraph_integer_t
     val=VECTOR_ELT(attr, 2);
     UNPROTECT(9);
   }
-  igraph_vector_int_destroy(&news);
+  igraph_vector_int_destroy_pv(&news);
```
krlmlr, line 653: I don't think this is necessary, the `igraph_vector_int_destroy()` function is best here. I'll revert that part.

krlmlr (commented):
Thanks! You converted many more instances than I expected, but that's fine.

krlmlr:
Thanks!

PR #1633, opened 2024-12-11
feat: Use safe accessor to vector elements

### LINKED ISSUE #1631: Replace `REAL()[0]` by `REAL_0()`, and similar
Flagged by CRAN:

```patch
diff -Nru orig/igraph/src/rinterface_extra.c patches/igraph/src/rinterface_extra.c
--- orig/igraph/src/rinterface_extra.c	2024-10-18 07:23:06.000000000 -0400
+++ patches/igraph/src/rinterface_extra.c	2024-12-03 11:51:27.514912912 -0500
@@ -4584,6 +4584,8 @@
                                  SEXP palgo) {
 
   igraph_t g;
+  if (Rf_length(pfrom) < 1)
+    Rf_error("empty pfrom");
   igraph_integer_t from=(igraph_integer_t) REAL(pfrom)[0];
   igraph_vs_t to;
   igraph_vector_int_t to_data;
```

Antonov548:
For #1631  .

### DIFF
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -128,6 +128,6 @@ void R_check_int_scalar(SEXP value);
 void R_check_real_scalar(SEXP value);
 void R_check_bool_scalar(SEXP value);
 
-igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index = 0);
-igraph_real_t R_get_real_scalar(SEXP sexp, R_xlen_t index = 0);
-igraph_bool_t R_get_bool_scalar(SEXP sexp, R_xlen_t index = 0);
+igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index);
+igraph_real_t R_get_real_scalar(SEXP sexp, R_xlen_t index);
+igraph_bool_t R_get_bool_scalar(SEXP sexp, R_xlen_t index);
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -128,6 +128,6 @@ void R_check_int_scalar(SEXP value);
 void R_check_real_scalar(SEXP value);
 void R_check_bool_scalar(SEXP value);
 
-igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index);
-igraph_real_t R_get_real_scalar(SEXP sexp, R_xlen_t index);
-igraph_bool_t R_get_bool_scalar(SEXP sexp, R_xlen_t index);
+igraph_error_t R_get_int_scalar(SEXP sexp, R_xlen_t index, igraph_integer_t *res);
+igraph_error_t R_get_real_scalar(SEXP sexp, R_xlen_t index, igraph_real_t *res);
+igraph_error_t R_get_bool_scalar(SEXP sexp, R_xlen_t index, igraph_bool_t *res);
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -127,3 +127,7 @@ igraph_error_t R_SEXP_to_attr_comb(SEXP input, igraph_attribute_combination_t *c
 void R_check_int_scalar(SEXP value);
 void R_check_real_scalar(SEXP value);
 void R_check_bool_scalar(SEXP value);
+
+igraph_integer_t R_get_int_scalar(SEXP sexp, int index);
+igraph_real_t R_get_real_scalar(SEXP sexp, int index);
+igraph_bool_t R_get_bool_scalar(SEXP sexp, int index);
--- a/src/rinterface.h
+++ b/src/rinterface.h
@@ -128,6 +128,6 @@ void R_check_int_scalar(SEXP value);
 void R_check_real_scalar(SEXP value);
 void R_check_bool_scalar(SEXP value);
 
-igraph_integer_t R_get_int_scalar(SEXP sexp, int index);
-igraph_real_t R_get_real_scalar(SEXP sexp, int index);
-igraph_bool_t R_get_bool_scalar(SEXP sexp, int index);
+igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index = 0);
+igraph_real_t R_get_real_scalar(SEXP sexp, R_xlen_t index = 0);
+igraph_bool_t R_get_bool_scalar(SEXP sexp, R_xlen_t index = 0);
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -99,7 +99,7 @@ void R_check_bool_scalar(SEXP value)
   }
 }
 
-igraph_integer_t R_get_int_scalar(SEXP sexp, int index)
+igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index)
 {
   if (Rf_length(sexp) < index + 1)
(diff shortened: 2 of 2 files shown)


szhorvat:
There are two things that bother me in the implementation:

 - We should avoid using `int` for indexing. The R API has a different type for this (I forget the name) that resolves to either a 32-bit or a 64-bit integer depending on configuration.
 - Naked calls to `igraph_errorf()` should be avoided. The behaviour depends on context. Normally, error checks are done through return codes using either `IGRAPH_CHECK` (usual) or `IGRAPH_R_CHECK` (top level). If we allow such naked calls, then we must keep in mind in what context functions like `R_get_int_scalar` are or aren't allowed to be called, which is a bit too much mental burden. It is always possible to have a function that takes a pointer to the variable it is going to set, and returns an error code.

I just wanted to note these potential issues. Proceed as you see fit.

src/rinterface_extra.c
```diff
@@ -99,6 +99,30 @@ void R_check_bool_scalar(SEXP value)
   }
 }
 
+igraph_integer_t R_get_int_scalar(SEXP sexp, int index)
```
krlmlr, line 102:
```suggestion
igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index)
```

Or perhaps `idx_t` or similar, if it exists in C?

src/rinterface.h
```diff
@@ -127,3 +127,7 @@ igraph_error_t R_SEXP_to_attr_comb(SEXP input, igraph_attribute_combination_t *c
 void R_check_int_scalar(SEXP value);
 void R_check_real_scalar(SEXP value);
 void R_check_bool_scalar(SEXP value);
+
+igraph_integer_t R_get_int_scalar(SEXP sexp, int index);
+igraph_real_t R_get_real_scalar(SEXP sexp, int index);
+igraph_bool_t R_get_bool_scalar(SEXP sexp, int index);
```
krlmlr, line 133:
```suggestion
igraph_integer_t R_get_int_scalar(SEXP sexp, R_xlen_t index = 0);
igraph_real_t R_get_real_scalar(SEXP sexp, R_xlen_t index = 0);
igraph_bool_t R_get_bool_scalar(SEXP sexp, R_xlen_t index = 0);
```

src/rinterface_extra.c
```diff
@@ -99,6 +99,30 @@ void R_check_bool_scalar(SEXP value)
...
+{
+  if (Rf_length(sexp) < index + 1)
+    igraph_errorf("Wrong index. Attempt to get element with index %" PRIuPTR " from vector of length %" PRIuPTR ".",
+                    __FILE__, __LINE__, IGRAPH_EINVAL, (uintptr_t) index, (uintptr_t) Rf_xlength(sexp));
```
krlmlr, line 106: Perhaps:

```suggestion
    Rf_error("Wrong index. Attempt to get element with index %" PRIuPTR " from vector of length %" PRIuPTR ".",
                    file, line, IGRAPH_EINVAL, (uintptr_t) index, (uintptr_t) Rf_xlength(sexp));
```

with `file` and `line` passed from the invocation point (e.g., https://github.com/igraph/rigraph/pull/1633/files#diff-44d6b09dbc4b7f024f1b2df8dcc51a6e3d489408a6f82def8dd49ffb1bcd5dc4R4610) .

szhorvat: This won't properly unwind the "finally" stack.

krlmlr: This is still better than undefined behavior.

I don't have a good answer here. Let's draft and iterate.

Antonov548: > This won't properly unwind the "finally" stack.

Is it not only the way to properlly unwind the `finally` stack is to call `igraph_error()` directly?

szhorvat: In this C core we use this approach:

```c
igraph_error_t R_get_int_scalar(SEXP sexp, R_xlen_t index, igraph_integer_t *res) {
    if (Rf_xlength(sexp) <= index) {
        IGRAPH_ERRORF("Wrong index. Attempt to get element with index %" PRIuPTR " from vector of length %" PRIuPTR ".", IGRAPH_EINVAL, (uintptr_t) index, (uintptr_t) Rf_xlength(sexp));
    }
    *res = (igraph_integer_t)REAL(sexp)[index];
    return IGRAPH_SUCCESS;
}
```

Then `R_get_int_scalar` returns an error code as usual.

Would this work for you?

Antonov548: > In the C core we use this approach:
> 
> ```c
> igraph_error_t R_get_int_scalar(SEXP sexp, R_xlen_t index, igraph_integer_t *res) {
>     if (Rf_xlength(sexp) <= index) {
>         IGRAPH_ERRORF("Wrong index. Attempt to get element with index %" PRIuPTR " from vector of length %" PRIuPTR ".", IGRAPH_EINVAL, (uintptr_t) index, (uintptr_t) Rf_xlength(sexp));
>     }
>     *res = (igraph_integer_t)REAL(sexp)[index];
>     return IGRAPH_SUCCESS;
> }
> ```
> 
> Then `R_get_int_scalar` returns an error code as usual, and can be wrapped in `IGRAPH_CHECK` or `IGRAPH_R_CHECK`.
> 
> Would this work for you?

Yes, this is what I was thinking about. Thanks.
I'll try this way.

szhorvat: > Is it not only the way to properly unwind the `finally` stack is to call `igraph_error()` directly?

Yes in the sense that `igraph_error()` should be used, but normally it is called through `IGRAPH_ERROR`, which ensures returning from the function with an error code, which can then be processed by `IGRAPH_CHECK`.

In igraph 0.10, the unwinding process can potentially have multiple stages, each of which need to be triggered appropriately. This means that `igraph_error()` should not blindly longjmp. That might leave some stages untriggered.

This is why in the R interface we have `IGRAPH_R_CHECK` which must only be used at the top level. Only at this level is it safe to longjmp from the error handler, i.e. to call `Rf_error()`. `IGRAPH_R_CHECK` controls whether `Rf_error()` is called by `igraph_error()` by setting a global flag.

Now one might argue that `R_get_int_scalar()` is (or should be) only called at the top level anyway, so we can forget all about this. (Note that I am not actually sure whether this is true, there might be some non-top-level functions where things like REAL(sexp)[0] are still used!) If you decide to not worry about this, there's a risk that someone will use `R_get_int_scalar()` in an inappropriate context in the future, which causes a bug that's only triggered in an error condition, and will be very painful to track down. Thus if you go this route, I recommend clearly documenting what check function is allowed to be called in what context, and perhaps even naming these functions in such a way that any misuse is immediately apparent.

Antonov548: > Thus if you go this route, I recommend clearly documenting what check function is allowed to be called in what context, and perhaps even naming these functions in such a way that any misuse is immediately apparent.

Not sure right now how to manage this, but I would think that functions which are have `SEXP` type as argument should be used only in top level functions. Which is also the case for `R_get_int_scalar()`. Maybe it make sense.

krlmlr: I agree that `R_get_int_scalar()` and friends should only be called at the top level and before any finalizers are set up. Also agree to document the purpose and the constraints.

krlmlr:
Thanks!

PR #1654, opened 2025-01-09
refactor: consolidate graph.incidence.* (#1483)

schochastics:
~~This PR removes (nested) for loops for `graph.incidence.dense()` when `weighted = TRUE`~~

~~(cherry picked commits from https://github.com/igraph/rigraph/pull/1518)~~

This PR removes `graph.incidence.sparse()` and `graph.incidence.dense()` and replaces it with a general `graph.incidence.build()` and more specialized helper functions.

### DIFF
--- a/R/incidence.R
+++ b/R/incidence.R
@@ -62,17 +62,11 @@ graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
   num_rows <- nrow(incidence)
   num_cols <- ncol(incidence)
 
-  if (inherits(incidence, "Matrix")) {
-    # General Sparse matrix processing
-    el <- process.sparse(incidence, num_rows)
-  } else if (!is.null(weighted)) {
-    # Dense weighted matrix processing (convert to sparse matrix first)
-    el <- process.sparse(as(incidence, "dgCMatrix"), num_rows)
-  } else {
+  # Handle dense unweighted matrices first
+  if (!inherits(incidence, "Matrix") && is.null(weighted)) {
     mode(incidence) <- "double"
     on.exit(.Call(R_igraph_finalizer))
 
-    # Dense unweighted matrix (potentially with multiple edges)
     mode_num <- switch(mode,
       "out" = 1,
       "in" = 2,
@@ -83,24 +77,32 @@ graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
     return(set_vertex_attr(res$graph, "type", value = res$types))
   }
 
+  # Convert to sparse matrix if not already sparse
+  if (!inherits(incidence, "Matrix")) {
+    incidence <- as(incidence, "dgCMatrix")
+  }
+
+  # Process sparse matrix to edgelist
+  el <- process.sparse(incidence, num_rows)
   # Adjust edgelist for directionality and mode
   el <- adjust.directionality(el, mode, directed)
 
-  # Handle weights or replicate rows for multiple edges
+  # Construct the graph object from processed edgelist
   if (!is.null(weighted)) {
+    # Handle weighted edges
     res <- make_empty_graph(n = num_rows + num_cols, directed = directed)
     weight_attr <- list(el[, 3])
     names(weight_attr) <- weighted
     res <- add_edges(res, edges = t(el[, 1:2]), attr = weight_attr)
   } else {
-    # create multiple edges according to the third column
+    # Handle unweighted edges, replicating rows for multiple edges
     el <- el[rep(seq_len(nrow(el)), times = el[, 3]), 1:2]
     res <- make_graph(n = num_rows + num_cols, c(t(el)), directed = directed)
   }
 
-  # Set vertex attributes and return
   set_vertex_attr(res, "type", value = c(rep(FALSE, num_rows), rep(TRUE, num_cols)))
 }
+
 
 
 
--- a/R/incidence.R
+++ b/R/incidence.R
@@ -46,7 +46,6 @@ process.sparse <- function(incidence, num_rows) {
 # Helper function to process dense matrices (matrix to edgelist)
 process.dense <- function(incidence, num_rows) {
   nz_ids <- which(incidence != 0, arr.ind = TRUE)
-  nz_ids <- nz_ids[order(nz_ids[, 1], nz_ids[, 2]), , drop = FALSE]
   el <- cbind(nz_ids, incidence[nz_ids])
   el[, 2] <- el[, 2] + num_rows
   el
--- a/R/incidence.R
+++ b/R/incidence.R
@@ -36,14 +36,14 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out", 
 ##
 ## -----------------------------------------------------------------
 
-# Helper function to process sparse matrices
+# Helper function to process sparse matrices (matrix to edgelist)
 process.sparse <- function(incidence, num_rows) {
   el <- mysummary(incidence)
   el[, 2] <- el[, 2] + num_rows
   as.matrix(el)
 }
 
-# Helper function to process dense matrices
+# Helper function to process dense matrices (matrix to edgelist)
 process.dense <- function(incidence, num_rows) {
   nz_ids <- which(incidence != 0, arr.ind = TRUE)
   nz_ids <- nz_ids[order(nz_ids[, 1], nz_ids[, 2]), , drop = FALSE]
@@ -52,6 +52,7 @@ process.dense <- function(incidence, num_rows) {
   el
 }
 
+# adjust edgelist according to directionality of edges
 adjust.directionality <- function(el, mode, directed) {
   if (!directed || mode == "out") {
     # No adjustment needed
@@ -79,6 +80,9 @@ graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
     # Dense weighted matrix processing
     el <- process.dense(incidence, num_rows)
   } else {
+    mode(incidence) <- "double"
+    on.exit(.Call(R_igraph_finalizer))
+
     # Dense unweighted matrix (potentially with multiple edges)
     mode_num <- switch(mode,
       "out" = 1,
@@ -86,7 +90,6 @@ graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
       "all" = 3,
       "total" = 3
     )
-    incidence <- incidence * 1.0 # TODO: check why integer fails below
     res <- .Call(R_igraph_biadjacency, incidence, directed, mode_num, multiple)
     return(set_vertex_attr(res$graph, "type", value = res$types))
   }
--- a/R/incidence.R
+++ b/R/incidence.R
@@ -36,16 +36,9 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out", 
 ##
 ## -----------------------------------------------------------------
 
-# Helper function to process sparse matrices (matrix to edgelist)
-process.sparse <- function(incidence, num_rows) {
-  el <- mysummary(incidence)
-  # adjust indices for second column to create a second mode
-  el[, 2] <- el[, 2] + num_rows
-  as.matrix(el)
-}
 
 # adjust edgelist according to directionality of edges
-adjust.directionality <- function(el, mode, directed) {
+modify_edgelist <- function(el, mode, directed) {
   if (!directed || mode == "out") {
     # No adjustment needed
     return(el)
@@ -57,7 +50,7 @@ adjust.directionality <- function(el, mode, directed) {
   rbind(el, reversed_edges)
 }
 
-graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
+graph_incidence_build <- function(incidence, directed = FALSE, mode = "out",
                                   multiple = FALSE, weighted = NULL) {
   num_rows <- nrow(incidence)
   num_cols <- ncol(incidence)
@@ -82,10 +75,11 @@ graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
     incidence <- as(incidence, "dgCMatrix")
   }
 
-  # Process sparse matrix to edgelist
-  el <- process.sparse(incidence, num_rows)
-  # Adjust edgelist for directionality and mode
-  el <- adjust.directionality(el, mode, directed)
+  el <- mysummary(incidence)
+  el[, 2] <- el[, 2] + num_rows
+  el <- as.matrix(el)
+
+  el <- modify_edgelist(el, mode, directed)
 
   # Construct the graph object from processed edgelist
   if (!is.null(weighted)) {
@@ -202,7 +196,7 @@ graph_from_biadjacency_matrix <- function(incidence, directed = FALSE,
     }
   }
 
-  res <- graph.incidence.build(incidence,
+  res <- graph_incidence_build(incidence,
     directed = directed,
     mode = mode, multiple = multiple,
     weighted = weighted
--- a/R/incidence.R
+++ b/R/incidence.R
@@ -49,15 +49,15 @@ adjust.directionality <- function(el, mode, directed) {
   if (!directed || mode == "out") {
     # No adjustment needed
     return(el)
-  } else if (mode == "in") {
-    # Reverse the edges
-    el[, 1:2] <- el[, c(2, 1)]
-  } else if (mode %in% c("all", "total")) {
-    # Add reversed edges
-    reversed_edges <- el[, c(2, 1, 3)]
-    el <- rbind(el, reversed_edges)
   }
-  el
+  reversed_edges <- el[, c(2, 1, 3)]
+  if (mode == "in") {
+    return(reversed_edges)
+  }
+  if (mode %in% c("all", "total")) {
+    # Add reversed edges
+    rbind(el, reversed_edges)
+  }
 }
 
 graph.incidence.build <- function(incidence, directed = FALSE, mode = "out",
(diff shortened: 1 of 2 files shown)


schochastics:
not sure why it fails but I am checking it

R/incidence.R
```diff
@@ -86,60 +84,53 @@ graph.incidence.sparse <- function(incidence, directed, mode, multiple,
...
-            ptr <- ptr + 1
-          }
-        }
-      }
+
+    # create an edgelist from the nonzero entries of the
+    # incidence matrix
+    idx <- which(incidence != 0, arr.ind = TRUE)
+    # convert to row-first order
+    idx <- idx[order(idx[, 1], idx[, 2]), ]
```
krlmlr, line 94: Do we need `order(c(...))` here?

``` r
order(3, 2, 1)
#> [1] 1
```

<sup>Created on 2025-01-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/incidence.R
```diff
@@ -86,60 +84,53 @@ graph.incidence.sparse <- function(incidence, directed, mode, multiple,
...
+    # add the value of the matrix. So a row is [s,t,incidence[s,t]]
+    el <- cbind(idx, incidence[idx])
+
+    # move from separate row/col indexing to 1..n1+n2 indexing
+    el[, 2] <- el[, 2] + n1
+
+    if (!directed || mode == "out") {
+      ## nothing do to
+    } else if (mode == "in") {
+      el[, 1:2] <- el[, c(2, 1)]
```
krlmlr, line 104: For symmetry?

```suggestion
      el[, 1:2] <- el[, 2:1]
```

R/incidence.R
```diff
@@ -86,60 +84,53 @@ graph.incidence.sparse <- function(incidence, directed, mode, multiple,
...
+    } else if (mode %in% c("all", "total")) {
+      reversed_el <- el[, c(2, 1, 3)]
+      names(reversed_el) <- names(el)
```
krlmlr, line 107: Is this `names()` or `rownames()` ? Do we even need it?

R/incidence.R
```diff
@@ -86,60 +84,53 @@ graph.incidence.sparse <- function(incidence, directed, mode, multiple,
...
+
     res <- make_empty_graph(n = n1 + n2, directed = directed)
-    weight <- list(weight)
+    weight <- list(el[, 3])
     names(weight) <- weighted
-    res <- add_edges(res, edges, attr = weight)
-    res <- set_vertex_attr(res, "type",
-      value = c(rep(FALSE, n1), rep(TRUE, n2))
-    )
+    res <- add_edges(res, edges = t(as.matrix(el[, 1:2])), attr = weight)
```
krlmlr, line 114: Why do we need `as.matrix()` here?

krlmlr (commented):
Thanks. Overall, the code doesn't seem to be well-prepared for the edge case of one-row matrices, we might need `drop = TRUE` in several places.

Could we perhaps forward to `graph.incidence.sparse()` in this case, perhaps even change `graph.incidence.sparse()` to accept both sparse and dense matrices? A lot of code there looks similar to what we have here.

schochastics:
Yes I now realize that I borrowed the logic from the sparse function and simply adapted where needed for the dense case. Thats not optimal. I think I can refactor both `*.sparse()` and `*.dense()` into one function.

schochastics:
The logic is now completely rewritten. There are no dedicated `graph.incidence.sparse` and `graph.incidence.dense` anymore since there was a lot of code duplication. Helper functions are now disjoint and there is no duplication of code anymore

szhorvat:
@schochastics I haven't been able to follow what's going on with the "incidence" (bipartite adjacency) matrix work, as I'm overwhelmed with other work at the moment. Do you think it would be good to have a short video meeting to discuss how the C library handled bipartite incidence matrices, so that when the R package finally starts using this functionality, the behaviour wouldn't change? I.e., just make sure that whatever's implemented in pure R now matches the C behaviour.

schochastics:
@szhorvat Happy to do so. I sent you an email

schochastics:
This PR can be reviewed now

R/incidence.R
```diff
@@ -37,109 +36,85 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out",
...
-  el[, 2] <- el[, 2] + n1
+  el[, 2] <- el[, 2] + num_rows
+  as.matrix(el)
+}
 
-  if (!is.null(weighted)) {
+# Helper function to process dense matrices (matrix to edgelist)
+process.dense <- function(incidence, num_rows) {
+  nz_ids <- which(incidence != 0, arr.ind = TRUE)
+  nz_ids <- nz_ids[order(nz_ids[, 1], nz_ids[, 2]), , drop = FALSE]
```
krlmlr, line 49: I still don't understand this idiom.

``` r
order(1:2, 3:4, 5:6)
#> [1] 1 2
order(1:2)
#> [1] 1 2
order(c(1:2, 3:4, 5:6))
#> [1] 1 2 3 4 5 6
```

<sup>Created on 2025-01-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Ideally, we'd lose the entire function.

R/incidence.R
```diff
@@ -37,109 +36,85 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out",
...
-      reversed_el <- el[, c(2, 1, 3)]
-      names(reversed_el) <- names(el)
-      el <- rbind(el, reversed_el)
-    }
+# adjust edgelist according to directionality of edges
+adjust.directionality <- function(el, mode, directed) {
+  if (!directed || mode == "out") {
+    # No adjustment needed
+    return(el)
+  } else if (mode == "in") {
```
krlmlr, line 60: Can this be simplified further?

```suggestion
  }
  reversed_edges <- ...
  if (mode == "in") {
```

R/incidence.R
```diff
@@ -37,109 +36,85 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out",
...
-      el[, 1:2] <- el[, c(2, 1)]
-    } else if (mode == 3) {
-      el <- rbind(el, el[, c(2, 1, 3)])
-    }
+  if (inherits(incidence, "Matrix")) {
+    # General Sparse matrix processing
+    el <- process.sparse(incidence, num_rows)
+  } else if (!is.null(weighted)) {
+    # Dense weighted matrix processing
+    el <- process.dense(incidence, num_rows)
```
krlmlr, line 81: Can we instead change `mysummary()` so that it also works for dense matrices?

krlmlr (commented):
Thanks, this might need yet another iteration or two.

R/incidence.R
schochastics: The function orders `el` according to the first and then second column. Works also if the matrix has only one row.
``` r
set.seed(12)
mat <- matrix(sample(1:5,20,replace = TRUE),10,2)
mat
#>       [,1] [,2]
#>  [1,]    2    2
#>  [2,]    2    1
#>  [3,]    3    4
#>  [4,]    5    2
#>  [5,]    5    2
#>  [6,]    4    5
#>  [7,]    2    4
#>  [8,]    3    1
#>  [9,]    2    2
#> [10,]    5    3
mat[order(mat[, 1],mat[ , 2]),]
#>       [,1] [,2]
#>  [1,]    2    1
#>  [2,]    2    2
#>  [3,]    2    2
#>  [4,]    2    4
#>  [5,]    3    1
#>  [6,]    3    4
#>  [7,]    4    5
#>  [8,]    5    2
#>  [9,]    5    2
#> [10,]    5    3

mat1 <- mat[1,,drop=FALSE]
mat1[order(mat1[, 1],mat1[ , 2]), ,drop=FALSE]
#>      [,1] [,2]
#> [1,]    2    2
```
this ordering is needed because what I can see is that edges are returned in row-first order hence we also need to add them in that order as `mysummary` also does.

``` r
set.seed(13)
mat2 <- matrix(sample(0:1,25,replace = TRUE),5,5)
mat2 <- Matrix::Matrix(mat2,sparse=TRUE)
igraph:::mysummary(mat2)
#> 5 x 5 sparse Matrix of class "dgCMatrix", with 14 entries
#>    i j x
#> 1  1 1 1
#> 2  3 1 1
#> 3  5 1 1
#> 4  2 2 1
#> 5  3 2 1
#> 6  4 2 1
#> 7  5 2 1
#> 8  1 3 1
#> 9  3 3 1
#> 10 4 3 1
#> 11 2 4 1
#> 12 3 4 1
#> 13 3 5 1
#> 14 5 5 1
```
so we need a function that efficiently does for dense matrices what `mysummary` does for sparse. So far, this seemed to be the most efficient approach but I will check if there is a way to integrate this in `mysummary` as you propose below

schochastics: I have to double check this row-first order claim I made. This might be a case of erroneous tests again. What the function is supposed to do is to order the edges first row than column wise. 
``` r
set.seed(12)
mat <- matrix(sample(1:5,20,replace = TRUE),10,2)
mat
#>       [,1] [,2]
#>  [1,]    2    2
#>  [2,]    2    1
#>  [3,]    3    4
#>  [4,]    5    2
#>  [5,]    5    2
#>  [6,]    4    5
#>  [7,]    2    4
#>  [8,]    3    1
#>  [9,]    2    2
#> [10,]    5    3
mat[order(mat[, 1],mat[ , 2]),]
#>       [,1] [,2]
#>  [1,]    2    1
#>  [2,]    2    2
#>  [3,]    2    2
#>  [4,]    2    4
#>  [5,]    3    1
#>  [6,]    3    4
#>  [7,]    4    5
#>  [8,]    5    2
#>  [9,]    5    2
#> [10,]    5    3

mat1 <- mat[1,,drop=FALSE]
mat1[order(mat1[, 1],mat1[ , 2]), ,drop=FALSE]
#>      [,1] [,2]
#> [1,]    2    2
```

but I realize now that `mysummary` does not do that.
``` r
set.seed(13)
mat2 <- matrix(sample(0:1,25,replace = TRUE),5,5)
mat2 <- Matrix::Matrix(mat2,sparse=TRUE)
igraph:::mysummary(mat2)
#> 5 x 5 sparse Matrix of class "dgCMatrix", with 14 entries
#>    i j x
#> 1  1 1 1
#> 2  3 1 1
#> 3  5 1 1
#> 4  2 2 1
#> 5  3 2 1
#> 6  4 2 1
#> 7  5 2 1
#> 8  1 3 1
#> 9  3 3 1
#> 10 4 3 1
#> 11 2 4 1
#> 12 3 4 1
#> 13 3 5 1
#> 14 5 5 1
```

I have to double check the tests again. If this is the case, then this ordering stuff can go away.

R/incidence.R
schochastics: I have tried that but failed so far. What is needed is a function that gets the indices of nonzero values of dense and sparse matrices. As far as I understand, `which` doesnt work for sparse and `summary` doesnt work for dense.

R/incidence.R
schochastics: Ok looks like the snapshot test was wrong (ie the old implementation had a bug). the `order` idiom is gone now. The open question is if this functionality can be merged into `mysummary`

R/incidence.R
schochastics: Do you see any issues with doing `as(incidence,"dgCMatrix")`?
It does not seem to have too much overhead, but not sure if it can have unpleasant side effects 
```r
n1 <- 5000
n2 <- 3000
incidence <- matrix(sample(0:1, n1*n2, replace = TRUE, c(0.5,0.5)), n1, n2)
inc_sparse <- Matrix::Matrix(incidence, sparse = TRUE)

bench::mark(check = FALSE,
  sparse_true = process.sparse(inc_sparse, n1),
  sparse_fake = process.sparse(as(incidence, "dgCMatrix"), n1),
  dense = process.dense(incidence, n1)
)
#># A tibble: 3 × 13
#>  expression       min   median `itr/sec` mem_alloc `gc/sec`
#>  <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#>1 sparse_true   70.7ms   78.2ms      6.15     315MB     7.69
#>2 sparse_fake  139.2ms  141.1ms      6.99     515MB     8.73
#>3 dense        317.4ms  483.2ms      2.07    1001MB     6.21
#># ℹ 7 more variables: n_itr <int>, n_gc <dbl>,
#>#   total_time <bch:tm>, result <list>, memory <list>,
#>#   time <list>, gc <list>
```

that way we can remove `process.dense()`

krlmlr: I'm in favor of outsourcing as much as possible to the Matrix package. May need `requireNamespace("Matrix", quietly = TRUE)` before `as(...)` .

R/incidence.R
krlmlr: Now I see what `order()` with two arguments is doing:

``` r
order(c(1, 1), 2:1)
#> [1] 2 1
```

<sup>Created on 2025-01-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I might have been wrong all the way, please double-check.

schochastics: the order part was scraped so we do not need to worry about potential undesired behaviour anymore

R/incidence.R
schochastics: Yes, I think this is probably the simplest solution?
```r
adjust.directionality <- function(el, mode, directed) {
  if (!directed || mode == "out") {
    # No adjustment needed
    return(el)
  }
  reversed_edges <- el[, c(2, 1, 3)]
  if (mode == "in") {
    return(reversed_edges)
  }
  if (mode %in% c("all", "total")) {
    # Add reversed edges
    return(rbind(el, reversed_edges))
  }
}
```

tests/testthat/_snaps/incidence.md
```diff
@@ -16,7 +16,7 @@
       IGRAPH UNWB 8 7 -- 
       + attr: type (v/l), name (v/c), weight (e/n)
       + edges (vertex names):
-      [1] A--c A--d B--b B--c B--e C--b C--d
+      [1] B--b C--b A--c B--c A--d C--d B--e
```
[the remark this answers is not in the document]
schochastics: This was the test I "fixed" If you think of this in matrix terms (rows:A-D, cols a-d), the ordering was rowwise (first all edges of A, then B, then C). Generally, igraph follows column first and this was the only case where this was not happening.

R/incidence.R
krlmlr: Yeah, the second and third `return()` aren't even needed.

https://style.tidyverse.org/functions.html#return

schochastics: First and second are needed, otherwise nothing is returned for `mode == "in"`
``` r
el <- matrix(
  c(1, 3, 1, 3, 1, 2, 3, 4, 4, 5, 5, 8, 8, 8, 1, 2, 1, 2, 2, 2, 1),
  nrow = 7L,
  ncol = 3L,
  dimnames = list(NULL, c("i", "j", "x"))
)

el
#>      i j x
#> [1,] 1 4 1
#> [2,] 3 4 2
#> [3,] 1 5 1
#> [4,] 3 5 2
#> [5,] 1 8 2
#> [6,] 2 8 2
#> [7,] 3 8 1

adjust.directionality <- function(el, mode, directed) {
  if (!directed || mode == "out") {
    # No adjustment needed
    return(el)
  }
  reversed_edges <- el[, c(2, 1, 3)]
  if (mode == "in") {
    reversed_edges
  }
  if (mode %in% c("all", "total")) {
    # Add reversed edges
    rbind(el, reversed_edges)
  }
}

adjust.directionality(el,mode = "in",directed = TRUE)
adjust.directionality(el,mode = "out",directed = TRUE)
#>      i j x
#> [1,] 1 4 1
#> [2,] 3 4 2
#> [3,] 1 5 1
#> [4,] 3 5 2
#> [5,] 1 8 2
#> [6,] 2 8 2
#> [7,] 3 8 1
adjust.directionality(el,mode = "all",directed = TRUE)
#>       i j x
#>  [1,] 1 4 1
#>  [2,] 3 4 2
#>  [3,] 1 5 1
#>  [4,] 3 5 2
#>  [5,] 1 8 2
#>  [6,] 2 8 2
#>  [7,] 3 8 1
#>  [8,] 4 1 1
#>  [9,] 4 3 2
#> [10,] 5 1 1
#> [11,] 5 3 2
#> [12,] 8 1 2
#> [13,] 8 2 2
#> [14,] 8 3 1
```

<sup>Created on 2025-01-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
@krlmlr ready to be checked again

krlmlr:
Can you please tweak and auto-merge?

R/incidence.R
```diff
@@ -37,109 +36,76 @@ graph.incidence <- function(incidence, directed = FALSE, mode = c("all", "out",
...
-      el <- rbind(el, el[, c(2, 1, 3)])
-    }
+    mode_num <- switch(mode,
+      "out" = 1,
+      "in" = 2,
+      "all" = 3,
+      "total" = 3
+    )
+    res <- .Call(R_igraph_biadjacency, incidence, directed, mode_num, multiple)
+    return(set_vertex_attr(res$graph, "type", value = res$types))
```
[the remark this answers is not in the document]
schochastics: @krlmlr there is an error in `biadjacency_impl()`
https://github.com/igraph/rigraph/blob/0d33d45e0d238911108a0f333c52750edcfd4c78/R/aaa-auto.R#L2051-L2055

should be something like
```r
res <- .Call(R_igraph_biadjacency, incidence, directed, mode_num, multiple)
set_vertex_attr(res$graph, "type", value = res$types)
```

because the actual created graph is in `res$graph`

PR #1655, opened 2025-01-09
refactor: removed `for` loops from `get.incidence.dense()` (#1483)

schochastics:
This PR removes (nested) for loops `get.incidence.dense()` when `attr != NULL`

(cherry picked commits from https://github.com/igraph/rigraph/pull/1518)


R/conversion.R
```diff
@@ -887,18 +886,22 @@ get.incidence.dense <- function(graph, types, names, attr) {
     res <- matrix(0, n1, n2)
```
krlmlr, line 886: What happens with non-numeric attributes?

krlmlr (approved):
Thanks, nice!

PR #1658, opened 2025-01-17
perf: Faster single bracket querying of a graph

### LINKED ISSUE #1465: can indexing be made faster
or should we delete these two lines

https://github.com/igraph/rigraph/blob/0e612b1169df68bbfeaa8b1679a41e5d7f946b00/R/indexing.R#L159

https://github.com/igraph/rigraph/blob/0e612b1169df68bbfeaa8b1679a41e5d7f946b00/R/indexing.R#L338

schochastics:
This PR refactors single bracket querying of a graph (`g[1:3,4:6]`) (closes #1465). 

### `[.igraph`
In the old version, the complete adjacency matrix was computed and then a subset created. The refactored function now builds the submatrix directly. This leads to a little speedup and a lower memory footprint.

``` r
set.seed(411)
g <- sample_gnp(5000,0.1)
bench::mark(
  check = FALSE,
  new = igraph:::get_adjacency_submatrix(g,1:100,1:100),
  old = as_adjacency_matrix(g)[1:100,1:100]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new          1.84ms   1.94ms    452.      4.26MB     24.0
#> 2 old         61.36ms 118.39ms      9.11   210.4MB     38.3

bench::mark(
  check = FALSE,
  new = igraph:::get_adjacency_submatrix(g,i = 1:100,j = 1:100,sparse = FALSE),
  old = as_adjacency_matrix(g,sparse = FALSE)[1:100,1:100]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new          1.69ms   1.81ms     462.     3.31MB     5.99
#> 2 old         59.95ms   61.1ms      16.0  190.81MB    16.0

E(g)$weight <- runif(ecount(g))
bench::mark(
  check = FALSE,
  new = igraph:::get_adjacency_submatrix(g,1:100,1:100),
  old = as_adjacency_matrix(g)[1:100,1:100]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new          1.88ms   1.99ms     451.      3.2MB     7.98
#> 2 old         55.34ms  62.09ms      14.5   209.7MB    23.6

bench::mark(
  check = FALSE,
  new = igraph:::get_adjacency_submatrix(g,i = 1:100,j = 1:100,sparse = FALSE),
  old = as_adjacency_matrix(g,sparse = FALSE)[1:100,1:100]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 new           1.7ms    1.8ms     432.     3.31MB     8.00
#> 2 old          56.7ms   57.8ms      13.8  190.81MB    13.8
```

<sup>Created on 2025-01-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -169,24 +169,6 @@ get_submatrix <- function(x, i, j, attr = NULL, sparse = TRUE) {
   res
 }
 
-clean_indices <- function(x, indices) {
-  if (is.character(indices)) {
-    return(match(indices, V(x)$name))
-  }
-  if (is.logical(indices)) {
-    if (length(indices) != vcount(x)) {
-      indices <- which(rep(indices, length.out = vcount(x)))
-    } else {
-      indices <- which(indices)
-    }
-    return(indices)
-  }
-  if (all(indices < 0)) {
-    return(seq_len(vcount(x))[indices])
-  }
-  indices
-}
-
 #' Query and manipulate a graph as it were an adjacency matrix
 #'
 #' @details
@@ -328,10 +310,10 @@ clean_indices <- function(x, indices) {
 
   # convert logical, character or negative i/j to proper vertex ids
   if (!missing(i)) {
-    i <- clean_indices(x, i)
+    i <- as_igraph_vs(x, i)
   }
   if (!missing(j)) {
-    j <- clean_indices(x, j)
+    j <- as_igraph_vs(x, j)
   }
 
   if (missing(i) && missing(j)) {
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -53,31 +53,24 @@
 # - G[1:3,2,eid=TRUE]
 #               create an edge sequence
 
-get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse = TRUE) {
+get_adjacency_submatrix <- function(x, i, j, attr = NULL, sparse = TRUE) {
   # If i or j is NULL, assume all nodes
   # if not NULL make sure to handle duplicates correctly
-  if (is.null(i)) {
-    i <- i_unique <- i_map <- seq_len(vcount(x))
-  } else {
-    i_unique <- unique(i)
-    i_map <- match(i, i_unique)
+  if (missing(i)) {
+    i <- seq_len(vcount(x))
+  }
+  if (missing(j)) {
+    j <- seq_len(vcount(x))
   }
 
-  if (is.null(j)) {
-    j <- j_unique <- j_map <- seq_len(vcount(x))
-  } else {
-    j_unique <- unique(j)
-    j_map <- match(j, j_unique)
-  }
-
-  adj <- adjacent_vertices(x, i_unique, mode = "out")
+  adj <- adjacent_vertices(x, i, mode = "out")
   i_degree <- map_int(adj, length)
 
-  from_id <- rep(i_unique, i_degree)
+  from_id <- rep(i, i_degree)
   to_id <- unlist(adj)
 
   edge_list <- cbind(from_id, to_id)
-  edge_list <- edge_list[edge_list[, 2] %in% j_unique, , drop = FALSE]
+  edge_list <- edge_list[edge_list[, 2] %in% j, , drop = FALSE]
 
   row_indices <- edge_list[, 1]
   col_indices <- edge_list[, 2]
@@ -85,21 +78,17 @@ get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse =
   values <- if (is.null(attr)) {
     1
   } else {
-    # get edge ids to locate edge attribute values
     valid_edges <- get_edge_ids(x, c(t(edge_list)))
     edge_attr(x, attr, valid_edges)
   }
 
 
-  unique_res <- Matrix::sparseMatrix(
-    i = match(row_indices, i_unique),
-    j = match(col_indices, j_unique),
+  res <- Matrix::sparseMatrix(
+    i = match(row_indices, i),
+    j = match(col_indices, j),
     x = values,
-    dims = c(length(i_unique), length(j_unique))
+    dims = c(length(i), length(j))
   )
-
-  # Expand to handle duplicated entries in i and j
-  res <- unique_res[i_map, j_map, drop = TRUE]
 
   if (!sparse) {
     res <- as.matrix(res)
@@ -110,7 +99,7 @@ get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse =
     colnames(res) <- vertex_attr(x, "name", j)
   }
 
-  return(res)
+  return(res[, , drop = FALSE])
 }
 
 
@@ -211,7 +200,8 @@ get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse =
 #'
 #' @method [ igraph
 #' @export
-`[.igraph` <- function(x, i, j, ..., from, to,
+`[.igraph` <- function(
+    x, i, j, ..., from, to,
     sparse = igraph_opt("sparsematrices"),
     edges = FALSE, drop = TRUE,
     attr = if (is_weighted(x)) "weight" else NULL) {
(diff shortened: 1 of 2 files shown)


R/indexing.R
```diff
@@ -54,6 +53,63 @@
...
 #               create an edge sequence
 
+get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse = TRUE) {
+  # If i or j is NULL, assume all nodes
+  if (is.null(i)) i <- seq_len(vcount(x))
+  if (is.null(j)) j <- seq_len(vcount(x))
+
+  # Handle duplicates
+  unique_i <- unique(i)
+  unique_j <- unique(j)
```
krlmlr, line 63: We need this only for the non-`NULL` case?

R/indexing.R
```diff
@@ -54,6 +53,63 @@
...
+
+  # Create mapping between to unique values
+  i_map <- match(i, unique_i)
```
krlmlr, line 66: And the mapping is trivial in the `NULL` or the "no duplicates" case?

R/indexing.R
```diff
@@ -54,6 +53,63 @@
...
+  j_map <- match(j, unique_j)
+
+  adj <- adjacent_vertices(x, unique_i, mode = "out")
+
+  edge_list <- cbind(rep(unique_i, sapply(adj, length)), unlist(adj))
```
krlmlr, line 71: This is a mouthful, can we extract the `rep()` call to a variable?

```suggestion
  edge_list <- cbind(rep(unique_i, map_int(adj, length)), unlist(adj))
```

krlmlr (commented):
Thanks. I see how duplicate `i` and `j` indexes add complexity to the new `get_adjacency_submatrix()` routine. How about the following logic:

- we don't compute `unique()`
- instead, we compute `adj_out <- adjacent_vertices(x, unique_i, mode = "out")` if `i` is given, and `adj_in <- adjacent_vertices(x, unique_j, mode = "in")` if `j` is given
- if none are given, we forward to a different existing routine
- if only one of `i` or `j` is given, we're done
- if both are given, we compute `vctrs::vec_set_intersect(adj_in, adj_out)`

How is the test coverage for this code?

I'd appreciate it if all changes that do not rely on `get_adjacency_submatrix()` came in one or several separate PRs. I'd like to do a few more iterations here.

schochastics:
Manipulating a graph via this logic was moved to #1661

maelle:
[`devtools::test_coverage_active_file()`](https://devtools.r-lib.org/reference/test.html)

krlmlr (commented):
`get.adjacency.sparse()` with a directed graph should involve only one or two copies. We could use that result with regular matrix subsetting and then tweak for the directed case. While this is not ideal, it may well be faster than anything we can come up in R land. Further optimizations are then possible by adding a `from` or `to` argument to `as_edgelist()` (which is called by `get.adjacency.sparse()`).

schochastics:
I am surprised myself, but the difference between the submatrix routine and `get.adjacency.sparse()` is quite big. I will try a bit more though to simplify/split the submatrix routine to make it more readable (without too much loss of performance).

``` r
pkgload::load_all("~/git/R_packages/rigraph/")
#> ℹ Loading igraph
g <- sample_gnp(5000,0.05, directed = FALSE)

bench::mark(check = FALSE,
  sub_sparse = get_adjacency_submatrix(g,1:100,sparse = TRUE),
  sub_dense = get_adjacency_submatrix(g,1:100,sparse = FALSE),
  full_sparse = as_adjacency_matrix(g,sparse = TRUE)[1:100,]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 sub_sparse    2.01ms   2.13ms     401.     5.05MB     53.9
#> 2 sub_dense     2.22ms   2.44ms     299.     8.43MB     53.8
#> 3 full_sparse  31.91ms  33.76ms      22.9  105.47MB     66.8

g <- sample_gnp(5000,0.05, directed = TRUE)

bench::mark(check = FALSE,
  sub_sparse = get_adjacency_submatrix(g,1:100,sparse = TRUE),
  sub_dense = get_adjacency_submatrix(g,1:100,sparse = FALSE),
  full_sparse = as_adjacency_matrix(g, sparse = TRUE)[1:100,]
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 3 × 6
#>   expression       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 sub_sparse    2.03ms   2.15ms     415.     3.97MB     24.0
#> 2 sub_dense     2.21ms   2.39ms     340.     7.79MB     45.8
#> 3 full_sparse  42.55ms   49.4ms      15.1   129.2MB     35.3
```

<sup>Created on 2025-01-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
I am now convinced that this is as good as it gets. Any other approach seems to loose too mach performance

R/indexing.R
```diff
@@ -53,6 +53,66 @@
 # - G[1:3,2,eid=TRUE]
 #               create an edge sequence
 
+get_adjacency_submatrix <- function(x, i = NULL, j = NULL, attr = NULL, sparse = TRUE) {
```
krlmlr, line 56: As agreed: this should be 

```suggestion
get_adjacency_submatrix_unique <- function(x, i = NULL, j = NULL, attr = NULL, sparse = TRUE) {
```

with the precondition that `i` and `j` are unique (or `NULL`), and we add a new `get_adjacency_submatrix()` that checks `anyDuplicated(...)` and blows up the resulting matrix with

```r
unique_res[i_map, j_map, drop = FALSE]
```

but only if needed.

krlmlr (commented):
Great progress, we're getting there!

schochastics:
Non unique i/j are now handled outside of `get_adjacency_submatrix()`. I think this is indeed very much cleaner.

R/indexing.R
```diff
@@ -192,32 +240,45 @@
...
     }
   }
+
+  sub_adjmat <- get_adjacency_submatrix(x, i = i, j = j, attr = attr, sparse = sparse)
+  if (i_has_dupes) {
+    sub_adjmat <- sub_adjmat[i_map, , drop = FALSE]
+  }
+  if (j_has_dupes) {
+    sub_adjmat <- sub_adjmat[, j_map, drop = FALSE]
+  }
```
krlmlr, line 280: I'd suspect double indexing to be faster and use less intermediate memory:

```suggestion
  if (i_has_dupes && j_has_dupes) {
    sub_adjmat <- sub_adjmat[i_map, j_map, drop = FALSE]
  } else if (i_has_dupes) {
    sub_adjmat <- sub_adjmat[i_map, , drop = FALSE]
  } else if (j_has_dupes) {
    sub_adjmat <- sub_adjmat[, j_map, drop = FALSE]
  }
```

R/indexing.R
```diff
@@ -53,6 +53,55 @@
...
+  if (!sparse) {
+    res <- as.matrix(res)
+  }
+
+  if ("name" %in% vertex_attr_names(x) && !is.null(dim(res))) {
+    rownames(res) <- vertex_attr(x, "name", i)
+    colnames(res) <- vertex_attr(x, "name", j)
+  }
+
+  return(res[, , drop = FALSE])
```
krlmlr, line 102: Do we need that?

```suggestion
  res
```

R/indexing.R
```diff
@@ -53,6 +53,55 @@
...
+  res <- Matrix::sparseMatrix(
+    i = match(row_indices, i),
+    j = match(col_indices, j),
+    x = values,
+    dims = c(length(i), length(j))
+  )
+
+  if (!sparse) {
+    res <- as.matrix(res)
+  }
```
krlmlr, line 95: Move this (and the arg) out of this function, and un-sparsify after the telescope?

R/indexing.R
```diff
@@ -53,6 +53,55 @@
...
 
+get_adjacency_submatrix <- function(x, i, j, attr = NULL, sparse = TRUE) {
+  # If i or j is NULL, assume all nodes
+  # if not NULL make sure to handle duplicates correctly
+  if (missing(i)) {
+    i <- seq_len(vcount(x))
+  }
+  if (missing(j)) {
+    j <- seq_len(vcount(x))
+  }
```
krlmlr, line 64: Is this worth the effort?

```suggestion
  if (missing(i)) {
    i_seq <- seq_len(vcount(x))
    has_i <- FALSE
  } else {
    i_seq <- i
    has_i <- TRUE
  }
  has_j <- missing(j)
```

and then only using `i_seq`, `has_i` and `has_j` to conditionally call `match()` and `%in%` ?

R/indexing.R
```diff
@@ -53,6 +53,55 @@
...
+
+  adj <- adjacent_vertices(x, i, mode = "out")
+  i_degree <- map_int(adj, length)
+
+  from_id <- rep(i, i_degree)
+  to_id <- unlist(adj)
+
+  edge_list <- data.frame(from = as.integer(from_id), to = as.integer(to_id))
+  edge_list <- edge_list[edge_list$to %in% j, ]
```
krlmlr, line 73: Avoid this if `!has_j` ?

R/indexing.R
```diff
@@ -53,6 +53,55 @@
...
+    1
+  } else {
+    valid_edges <- get_edge_ids(x, edge_list)
+    edge_attr(x, attr, valid_edges)
+  }
+
+
+  res <- Matrix::sparseMatrix(
+    i = match(row_indices, i),
+    j = match(col_indices, j),
```
krlmlr, line 88:
```suggestion
    i = if (has_i) match(row_indices, i) else row_indices,
    j = if (has_j) match(col_indices, j) else col_indices,
```

krlmlr (commented):
Do another final benchmarking round and then merge?

R/indexing.R
schochastics: > Is this worth the effort?
> 
> and then only using `i_seq`, `has_i` and `has_j` to conditionally call `match()` and `%in%` ?

Seems so. We do gain some time and I think it is enough to justify the implementation

krlmlr:
Thanks!

PR #1661, opened 2025-01-19
fix!: Subset assignment of a graph avoids addition of double edges and ignores loops unless the new `loops` argument is set to `TRUE`

### LINKED ISSUE #1662: Intended behaviour of `[<-.igraph`
Trying to add some more tests for #1661, I noticed some probably not intended behaviours of the current implementation

``` r
library(igraph)
#> 
#> 
#>     decompose, spectrum
#> 
#>     union
packageVersion("igraph")
#> [1] '2.1.2'
g1 <- make_empty_graph(n = 10, directed = FALSE)
g1[1:5,] <- 1
g1[]
#> 10 x 10 sparse Matrix of class "dgCMatrix"
#>                          
#>  [1,] 1 2 2 2 2 1 1 1 1 1
#>  [2,] 2 1 2 2 2 1 1 1 1 1
#>  [3,] 2 2 1 2 2 1 1 1 1 1
#>  [4,] 2 2 2 1 2 1 1 1 1 1
#>  [5,] 2 2 2 2 1 1 1 1 1 1
#>  [6,] 1 1 1 1 1 . . . . .
#>  [7,] 1 1 1 1 1 . . . . .
#>  [8,] 1 1 1 1 1 . . . . .
#>  [9,] 1 1 1 1 1 . . . . .
#> [10,] 1 1 1 1 1 . . . . .

g2 <- make_empty_graph(n = 10, directed = FALSE)
g2[1:5,1:5] <- 1
g2[]
#> 10 x 10 sparse Matrix of class "dgCMatrix"
#>                          
#>  [1,] 1 2 2 2 2 . . . . .
#>  [2,] 2 1 2 2 2 . . . . .
#>  [3,] 2 2 1 2 2 . . . . .
#>  [4,] 2 2 2 1 2 . . . . .
#>  [5,] 2 2 2 2 1 . . . . .
#>  [6,] . . . . . . . . . .
#>  [7,] . . . . . . . . . .
#>  [8,] . . . . . . . . . .
#>  [9,] . . . . . . . . . .
#> [10,] . . . . . . . . . .
```

<sup>Created on 2025-01-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In my opinion, this should rather look like this
``` r
A1 <- matrix(0,10,10)
A1[1:5,] <- 1
diag(A1) <- 0
A1
#>       [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8] [,9] [,10]
#>  [1,]    0    1    1    1    1    1    1    1    1     1
#>  [2,]    1    0    1    1    1    1    1    1    1     1
#>  [3,]    1    1    0    1    1    1    1    1    1     1
#>  [4,]    1    1    1    0    1    1    1    1    1     1
#>  [5,]    1    1    1    1    0    1    1    1    1     1
#>  [6,]    0    0    0    0    0    0    0    0    0     0
#>  [7,]    0    0    0    0    0    0    0    0    0     0
#>  [8,]    0    0    0    0    0    0    0    0    0     0
#>  [9,]    0    0    0    0    0    0    0    0    0     0
#> [10,]    0    0    0    0    0    0    0    0    0     0

A2 <- matrix(0,10,10)
A2[1:5,1:5] <- 1
diag(A2) <- 0
A2
#>       [,1] [,2] [,3] [,4] [,5] [,6] [,7] [,8] [,9] [,10]
#>  [1,]    0    1    1    1    1    0    0    0    0     0
#>  [2,]    1    0    1    1    1    0    0    0    0     0
#>  [3,]    1    1    0    1    1    0    0    0    0     0
#>  [4,]    1    1    1    0    1    0    0    0    0     0
#>  [5,]    1    1    1    1    0    0    0    0    0     0
#>  [6,]    0    0    0    0    0    0    0    0    0     0
#>  [7,]    0    0    0    0    0    0    0    0    0     0
#>  [8,]    0    0    0    0    0    0    0    0    0     0
#>  [9,]    0    0    0    0    0    0    0    0    0     0
#> [10,]    0    0    0    0    0    0    0    0    0     0
```

<sup>Created on 2025-01-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

thoughts @szhorvat @krlmlr ?

schochastics:
This PR refactors single bracket manipulating of a graph (`g[1:3,4:6] <- 2`) (#1465). 

The only real change is the use of `get_edge_ids` instead of the old `[.igraph` to get edge ids which makes the function more readable, slightly faster and a lower memory footprint.

Fixes an unintended behaviour (fix #1662)

### DIFF
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -329,18 +329,17 @@ length.igraph <- function(x) {
 }
 
 expand.grid.unordered <- function(i, j, loops = FALSE, directed = FALSE) {
-  grid <- expand.grid(i = i, j = j)
+  grid <- vctrs::vec_expand_grid(i = i, j = j)
   if (!directed) {
-    grid <- unique(data.frame(
+    grid <- vctrs::vec_unique(data.frame(
       i = pmin(grid$i, grid$j),
       j = pmax(grid$i, grid$j)
     ))
   }
   if (!loops) {
-    grid[grid[, 1] != grid[, 2], ]
-  } else {
-    grid
+    grid <- grid[grid[, 1] != grid[, 2], ]
   }
+  grid
 }
 
 #' @method [<- igraph
@@ -427,7 +426,7 @@ expand.grid.unordered <- function(i, j, loops = FALSE, directed = FALSE) {
 
       if (is.null(attr)) {
         if (value > 1) {
-          warning("value greater than one but graph is not weighted and no attribute was specified. Only unweighted edges are added.", call. = FALSE)
+          cli::cli_abort("value greater than one but graph is not weighted and no attribute was specified.")
         }
         x <- add_edges(x, toadd)
       } else {
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -1,4 +1,3 @@
-
 ## IGraph library.
 ## Copyright (C) 2010-2012  Gabor Csardi <csardi.gabor@gmail.com>
 ## 334 Harvard street, Cambridge, MA 02139 USA
@@ -391,9 +390,9 @@ length.igraph <- function(x) {
     if (missing(i) && missing(j)) {
       todel <- seq_len(ecount(x))
     } else if (missing(j)) {
-      todel <- unlist(incident(x, v = i, mode = "out"))
+      todel <- unlist(sapply(i, function(id) incident(x, v = id, mode = "out")))
     } else if (missing(i)) {
-      todel <- unlist(incident(x, v = j, mode = "in"))
+      todel <- unlist(sapply(j, function(id) incident(x, v = id, mode = "in")))
     } else {
       edge_pairs <- expand.grid(i, j)
       edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -328,11 +328,27 @@ length.igraph <- function(x) {
   vcount(x)
 }
 
+expand.grid.unordered <- function(i, j, loops = FALSE, directed = FALSE) {
+  grid <- expand.grid(i = i, j = j)
+  if (!directed) {
+    grid <- unique(data.frame(
+      i = pmin(grid$i, grid$j),
+      j = pmax(grid$i, grid$j)
+    ))
+  }
+  if (!loops) {
+    grid[grid[, 1] != grid[, 2], ]
+  } else {
+    grid
+  }
+}
+
 #' @method [<- igraph
 #' @family functions for manipulating graph structure
 #' @export
 `[<-.igraph` <- function(x, i, j, ..., from, to,
                          attr = if (is_weighted(x)) "weight" else NULL,
+                         loops = FALSE,
                          value) {
   ################################################################
   ## Argument checks
@@ -404,11 +420,15 @@ length.igraph <- function(x) {
     i <- if (missing(i)) as.numeric(V(x)) else as_igraph_vs(x, i)
     j <- if (missing(j)) as.numeric(V(x)) else as_igraph_vs(x, j)
     if (length(i) != 0 && length(j) != 0) {
-      edge_pairs <- expand.grid(i, j)
+      edge_pairs <- expand.grid.unordered(i, j, loops = loops, directed = is_directed(x))
+
       edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
       toadd <- c(rbind(edge_pairs[edge_ids == 0, 1], edge_pairs[edge_ids == 0, 2]))
 
       if (is.null(attr)) {
+        if (value > 1) {
+          warning("value greater than one but graph is not weighted and no attribute was specified. Only unweighted edges are added.", call. = FALSE)
+        }
         x <- add_edges(x, toadd)
       } else {
         x <- add_edges(x, toadd, attr = structure(list(value), names = attr))
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -390,9 +390,9 @@ length.igraph <- function(x) {
     if (missing(i) && missing(j)) {
       todel <- seq_len(ecount(x))
     } else if (missing(j)) {
-      todel <- unlist(sapply(i, function(id) incident(x, v = id, mode = "out")))
+      todel <- unlist(incident_edges(x, v = i, mode = "out"))
     } else if (missing(i)) {
-      todel <- unlist(sapply(j, function(id) incident(x, v = id, mode = "in")))
+      todel <- unlist(incident_edges(x, v = j, mode = "in"))
     } else {
       edge_pairs <- expand.grid(i, j)
       edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
--- a/R/indexing.R
+++ b/R/indexing.R
@@ -335,8 +335,6 @@ length.igraph <- function(x) {
 `[<-.igraph` <- function(x, i, j, ..., from, to,
                          attr = if (is_weighted(x)) "weight" else NULL,
                          value) {
-  ## TODO: rewrite this in C to make it faster
-
   ################################################################
   ## Argument checks
   if ((!missing(from) || !missing(to)) &&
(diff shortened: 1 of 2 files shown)


R/indexing.R
```diff
@@ -391,37 +389,31 @@ length.igraph <- function(x) {
...
     } else if (missing(j)) {
-      todel <- unlist(x[[i, , ..., edges = TRUE]])
+      todel <- unlist(incident(x, v = i, mode = "out"))
     } else if (missing(i)) {
-      todel <- unlist(x[[, j, ..., edges = TRUE]])
+      todel <- unlist(incident(x, v = j, mode = "in"))
     } else {
-      todel <- unlist(x[[i, j, ..., edges = TRUE]])
+      edge_pairs <- expand.grid(i, j)
+      edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
```
krlmlr, line 399: Is this covered by tests?

```suggestion
      edge_ids <- get_edge_ids(x, as.vector(t(edge_pairs)))
```

The interface of `get_edge_ids()` is interesting. Should we extend that to accept two-column data frames?

R/indexing.R
```diff
@@ -391,37 +389,31 @@ length.igraph <- function(x) {
     (is.null(attr) && is.numeric(value) && value == 0)) {
     ## Delete edges
     if (missing(i) && missing(j)) {
-      todel <- unlist(x[[, , ..., edges = TRUE]])
+      todel <- seq_len(ecount(x))
```
krlmlr, line 392: This looks consistent with the current implementation, but do we really expect all edges to be deleted with `g[] <- ...` ?

R/indexing.R
```diff
@@ -391,37 +389,31 @@ length.igraph <- function(x) {
...
-        to <- setdiff(j, exv[[idx]])
-        if (length(to != 0)) {
-          rbind(i[idx], setdiff(j, exv[[idx]]))
-        } else {
-          numeric()
-        }
-      }))
-      ## Do the changes
+      edge_pairs <- expand.grid(i, j)
+      edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
```
krlmlr, line 409: Same as above.

krlmlr (commented):
Thanks, I like the size and intention of this PR.

R/indexing.R
schochastics: Going through the existing tests, I realize there are some gaps. I will add a set of tests

schochastics: > Is this covered by tests?
> 
Going through the existing tests, I realize there are some gaps. I will add a set of tests for this functionality

schochastics: > The interface of `get_edge_ids()` is interesting. Should we extend that to accept two-column data frames?

It has been bothering me mildly for years as a user that edges need to be supplied as a vector (same with `add_edges()` and `delete_edges()` and probably more). However that's required by the c core. It might be too much of a fundamental change at this point?

R/indexing.R
schochastics: I think it makes sense that `g[]<-0` deletes all edges. If it is useful is another question but at least it is consistent.

R/indexing.R
krlmlr: The `c(rbind(...))` pattern is probably fine, I forgot the semantics for vectors:

``` r
c(rbind(1:3, 4:6))
#> [1] 1 4 2 5 3 6
c(t(data.frame(1:3, 4:6)))
#> [1] 1 4 2 5 3 6
as.vector(t(data.frame(1:3, 4:6)))
#> [1] 1 4 2 5 3 6
```

<sup>Created on 2025-01-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

There are two layers here: the C core and the R interface. We should provide an idiomatic R user interface that translates to what the C core needs.

R/indexing.R
krlmlr: Oh, right. Makes sense.

schochastics:
should probably be blocked until #1662 is resolved

R/indexing.R
```diff
@@ -329,14 +328,28 @@ length.igraph <- function(x) {
...
 }
 
+expand.grid.unordered <- function(i, j, loops = FALSE, directed = FALSE) {
+  grid <- expand.grid(i = i, j = j)
+  if (!directed) {
+    grid <- unique(data.frame(
+      i = pmin(grid$i, grid$j),
+      j = pmax(grid$i, grid$j)
+    ))
+  }
```
krlmlr, line 338: Faster on my machine:

``` r
grid_1 <- function(i, j) {
  grid <- expand.grid(i = i, j = j)
  grid <- unique(data.frame(
    i = pmin(grid$i, grid$j),
    j = pmax(grid$i, grid$j)
  ))
  grid
}

grid_2 <- function(i, j) {
  grid <- vctrs::vec_expand_grid(i = i, j = j)
  grid <- vctrs::vec_unique(data.frame(
    i = pmin(grid$i, grid$j),
    j = pmax(grid$i, grid$j)
  ))
  grid
}

bench::mark(
  grid_1(2:40, 1:30),
  grid_2(2:40, 1:30),
  check = FALSE
)
#> # A tibble: 2 × 6
#>   expression              min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>         <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 grid_1(2:40, 1:30)  469.6µs  518.9µs     1777.     348KB     46.6
#> 2 grid_2(2:40, 1:30)   81.8µs   91.6µs     9610.    76.4KB     28.5
```

<sup>Created on 2025-01-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/indexing.R
```diff
@@ -329,14 +328,28 @@ length.igraph <- function(x) {
...
+  if (!loops) {
+    grid[grid[, 1] != grid[, 2], ]
+  } else {
+    grid
+  }
```
krlmlr, line 343: For minimal complexity, perhaps:

```suggestion
  if (!loops) {
    grid <- grid[grid[, 1] != grid[, 2], ]
  }
  
  grid
```

R/indexing.R
krlmlr: `c(rbind(...))` is faster for data frames by an order of magnitude, but not for matrices:

``` r
df <- as.data.frame(cbind(1:30, 4:33))

bench::mark(
  c(t(df)),
  c(rbind(df[, 1], df[, 2])),
  c(rbind(df[[1]], df[[2]]))
)
#> # A tibble: 3 × 6
#>   expression                      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                 <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 c(t(df))                    12.96µs  14.51µs    57926.    74.4KB     34.8
#> 2 c(rbind(df[, 1], df[, 2]))   5.33µs   5.99µs   157537.      576B     47.3
#> 3 c(rbind(df[[1]], df[[2]]))   3.65µs   4.26µs   219083.      576B     43.8

m <- cbind(1:30, 4:33)

bench::mark(
  c(t(m)),
  c(rbind(m[, 1], m[, 2]))
)
#> # A tibble: 2 × 6
#>   expression                    min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>               <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 c(t(m))                     820ns 983.94ns   964028.      576B     96.4
#> 2 c(rbind(m[, 1], m[, 2]))    984ns   1.15µs   791706.      576B      0
```

<sup>Created on 2025-01-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Draft PR for new UI in https://github.com/igraph/rigraph/pull/1663.

R/indexing.R
```diff
@@ -391,37 +404,35 @@ length.igraph <- function(x) {
...
-      }))
-      ## Do the changes
+      edge_pairs <- expand.grid.unordered(i, j, loops = loops, directed = is_directed(x))
+
+      edge_ids <- get_edge_ids(x, c(rbind(edge_pairs[, 1], edge_pairs[, 2])))
+      toadd <- c(rbind(edge_pairs[edge_ids == 0, 1], edge_pairs[edge_ids == 0, 2]))
+
       if (is.null(attr)) {
+        if (value > 1) {
+          warning("value greater than one but graph is not weighted and no attribute was specified. Only unweighted edges are added.", call. = FALSE)
```
krlmlr, line 430: I think this should be an error. In new code, we can use `cli::cli_abort()` like elsewhere, and gradually switch the old code.

krlmlr:
Do you want to add more tests here?

schochastics:
> Do you want to add more tests here?

Yes!

krlmlr (approved):
Looks good to me after #1663. Please auto-squash or squash when good on your end.

PR #1663, opened 2025-01-20
feat: `get_edge_ids()` accepts data frames and matrices

krlmlr:
Also for `add_edges()` ?

The matrix case works already by accident, a two-row matrix will give the intended result. Should we soft-deprecate the matrix case to later support two-column matrices?

``` r
library(igraph)
set.seed(20250120)
g <- sample_gnp(20, 0.1)
g
#> IGRAPH dc0368b U--- 20 22 -- Erdos-Renyi (gnp) graph
#> + attr: name (g/c), type (g/c), loops (g/l), p (g/n)
#> + edges from dc0368b:
#>  [1]  1-- 3  1-- 5  3-- 8  7-- 8  5--11 10--11  2--12  3--14  5--14  1--15
#> [11] 10--15 11--15  4--16  6--16 12--16  1--17 15--17  3--18  4--18  8--19
#> [21] 11--19 12--19
x <- c(1, 3, 1, 5, 3, 8)
get_edge_ids(g, x)
#> [1] 1 2 3
m <- matrix(x, nrow = 2)
m
#>      [,1] [,2] [,3]
#> [1,]    1    1    3
#> [2,]    3    5    8
get_edge_ids(g, m)
#> [1] 1 2 3
```

<sup>Created on 2025-01-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/interface.R
+++ b/R/interface.R
@@ -1,4 +1,3 @@
-
 #' Check whether a graph is directed
 #'
 #' @description
@@ -467,7 +466,7 @@ get.edges <- function(graph, es) {
   ends(graph, es, names = FALSE)
 }
 
-el_to_vec <- function(x, call = caller_env()) {
+el_to_vec <- function(x, call = rlang::caller_env()) {
   if (is.data.frame(x)) {
     c(rbind(x[[1]], x[[2]]))
   } else if (inherits(x, "matrix")) {
@@ -476,7 +475,7 @@ el_to_vec <- function(x, call = caller_env()) {
   } else if (is.vector(x)) {
     x
   } else {
-    cli::cli_abort("only two-column data.frames and matrices, and vectors are allowed for vp")
+    cli::cli_abort("only two-column data.frames and matrices, and vectors are allowed for vp", call = call)
   }
 }
 
@@ -559,7 +558,6 @@ get.edge.ids <- function(graph,
                          directed = TRUE,
                          error = FALSE,
                          multi = deprecated()) {
-
   if (lifecycle::is_present(multi)) {
     if (isTRUE(multi)) {
       lifecycle::deprecate_stop("2.0.0", "get.edge.ids(multi = )")
--- a/R/interface.R
+++ b/R/interface.R
@@ -475,18 +475,21 @@ el_to_vec <- function(x, call = rlang::caller_env()) {
     }
   } else if (inherits(x, "matrix")) {
     dimx <- dim(x)
-    if (identical(dimx, c(2L, 2L))) {
+    nrow <- dimx[[1]]
+    ncol <- dimx[[2]]
+    if (nrow == 2 && ncol == 2) {
       lifecycle::deprecate_stop(
         "2.1.5",
         "get_edge_ids(vp = 'is not allowed to be a 2 times 2 matrix')"
       )
-    } else if (dimx[1] == 2L) {
+    } else if (nrow == 2) {
       lifecycle::deprecate_warn(
         "2.1.5",
-        "get_edge_ids(vp = 'supplied as a matrix should be a n times 2 matrix, not 2 times n')"
+        "get_edge_ids(vp = 'supplied as a matrix should be a n times 2 matrix, not 2 times n')",
+        details = "either transpose the matrix with t() or convert it to a data.frame with two columns."
       )
       c(x)
-    } else if (dimx[2] == 2L) {
+    } else if (ncol == 2) {
       c(t(x))
     } else {
       cli::cli_abort("{.args vp} was supplied as a {dimx[1]} times {dimx[2]} matrix. Only n times 2 matrices are allowed")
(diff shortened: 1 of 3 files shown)


schochastics:
`[.igraph` is using the two-row matrix
https://github.com/igraph/rigraph/blob/fe3b5b89ec6f7daa7f08d595f1a2a55b3cdd8eae/R/indexing.R#L186

R/interface.R
```diff
@@ -467,6 +467,19 @@ get.edges <- function(graph, es) {
   ends(graph, es, names = FALSE)
 }
 
+el_to_vec <- function(x) {
```
krlmlr, line 470: If there's an error, we want it to print like "Error in get_edge_ids()", not "Error in el_to_vec()" . Need a `call` argument for this, and then pass it to `cli_abort()` .

```suggestion
el_to_vec <- function(x, call = caller_env()) {
```

R/interface.R
```diff
@@ -467,6 +466,19 @@ get.edges <- function(graph, es) {
   ends(graph, es, names = FALSE)
 }
 
+el_to_vec <- function(x, call = rlang::caller_env()) {
+  if (is.data.frame(x)) {
+    c(rbind(x[[1]], x[[2]]))
```
krlmlr, line 471: Check if we have exactly two columns and if they are of the same type?

R/interface.R
```diff
@@ -467,6 +466,19 @@ get.edges <- function(graph, es) {
...
+  } else if (inherits(x, "matrix")) {
+    # c(t(x[,1:2])) TODO: decide on deprecation note
+    x
```
krlmlr, line 474: - if two rows, deprecate hard if also two columns, otherwise deprecate soft
- if two columns, transpose
- otherwise, reject

krlmlr (commented):
Thanks!

schochastics:
The warnings in tests are all from "wrong" use of `get_edge_ids()` with 2xn matrices. These occur all for `[.igraph` and `[<-.igraph`. So we need to merge #1658 and #1661 first and rebase this PR.

cc @maelle If you have time (no rush) can you check if my use of lifecycle (in code and tests) is ok? (never actively used it before)

R/interface.R
```diff
@@ -467,6 +466,38 @@ get.edges <- function(graph, es) {
...
+el_to_vec <- function(x, call = rlang::caller_env()) {
+  if (is.data.frame(x)) {
+    if (typeof(x[[1]]) == typeof(x[[2]])) {
+      c(rbind(x[[1]], x[[2]]))
+    } else {
+      cli::cli_abort("The columns of the data.frame are of different type ({typeof(x[[1]])} and {typeof(x[[2]])}) ")
+    }
+  } else if (inherits(x, "matrix")) {
+    dimx <- dim(x)
+    if (identical(dimx, c(2L, 2L))) {
```
krlmlr, line 478:
```suggestion
    dimx <- dim(x)
    nrow <- dimx[[1]]
    ncol <- dimx[[2]]
    if (nrow == 2 && ncol == 2) {
```

R/interface.R
```diff
@@ -467,6 +466,38 @@ get.edges <- function(graph, es) {
...
+      lifecycle::deprecate_stop(
+        "2.1.5",
+        "get_edge_ids(vp = 'is not allowed to be a 2 times 2 matrix')"
+      )
+    } else if (dimx[1] == 2L) {
+      lifecycle::deprecate_warn(
+        "2.1.5",
+        "get_edge_ids(vp = 'supplied as a matrix should be a n times 2 matrix, not 2 times n')"
```
krlmlr, line 486: Can we make concrete suggestions, e.g., to call `t()` or by using a data frame?

R/interface.R
```diff
@@ -467,6 +466,38 @@ get.edges <- function(graph, es) {
...
+      )
+      c(x)
+    } else if (dimx[2] == 2L) {
+      c(t(x))
+    } else {
+      cli::cli_abort("{.args vp} was supplied as a {dimx[1]} times {dimx[2]} matrix. Only n times 2 matrices are allowed")
+    }
+  } else if (is.vector(x)) {
+    x
```
krlmlr, line 495: Ideally, I'd like to soft-deprecate this as well. But this has other dependencies. Let's plan this work and do it later.

tests/testthat/test-interface.R
```diff
@@ -184,3 +183,42 @@ test_that("get.edge.ids() deprecation", {
...
+  V(g)$name <- letters[1:3]
+  df <- data.frame(from = c("a", "b"), to = c(1, 2))
+  expect_snapshot(error = TRUE, {
+    get_edge_ids(g, df)
+  })
+})
+
+test_that("get_edge_id() errors correctly for wrong matrices", {
+  g <- make_full_graph(10)
+  mat <- matrix(c(1, 2, 3, 4), 2, 2)
```
krlmlr, line 220:
```suggestion
  mat <- matrix(c(1, 2, 3, 4), nrow = 2, ncol = 2)
```

Reference: https://design.tidyverse.org/call-data-details.html

krlmlr (commented):
Thanks, can you please tweak and auto-merge when good?

PR #1675, opened 2025-02-08
test: merged and refactored flow.R tests

schochastics:
All tests related to functions in the file `flow.R` were moved to `test-flow.R` and refactored to fit https://masalmon.eu/2024/05/23/refactoring-tests/. 

Some additional tests were added and code cleaned in `flow.R`

cc @maelle keeping this on draft, would like to get your opinion on the structure 😁


tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  g_star_v9 <- graph_from_literal(s - +a:b - +t, a - +1:2:3:4:5 - +b)
+  all_cuts_star_v9 <- st_min_cuts(g_star_v9, source = "s", target = "t")
+  expect_equal(all_cuts_star_v9$value, 2)
+  expect_equal(unvs(all_cuts_star_v9$cuts), list(c(1, 2), c(1, 9), c(3, 9)))
+  expect_equal(unvs(all_cuts_star_v9$partition1s), list(1, c(1, 3), c(1, 3, 2, 9, 8, 7, 6, 5)))
+})
+
+test_that("st_cuts errors work", {
+  g_path <- graph_from_literal(a - +b - +c - +d - +e)
+  expect_error(st_cuts(g_path, source = "a", target = NULL))
```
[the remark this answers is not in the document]
maelle: Or, one approach I haven't actually used myself but that looks really neat: https://www.mm218.dev/posts/2023-11-07-classed-errors/

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+
+test_that("max_flow works", {
+  edge_mat <- rbind(c(1, 3, 3), c(3, 4, 1), c(4, 2, 2), c(1, 5, 1), c(5, 6, 2), c(6, 2, 10))
+  colnames(edge_mat) <- c("from", "to", "capacity")
+  g_ring_acyc <- graph_from_data_frame(as.data.frame(edge_mat))
+  flow <- max_flow(g_ring_acyc, source = "1", target = "2")
+  expect_equal(flow$value, 2)
+  expect_equal(as.vector(flow$flow), rep(1, 6))
+  expect_equal(sort(as.vector(flow$cut)), c(2, 4))
+  expect_equal(sort(as.vector(flow$partition1)), 1:2)
```
maelle, line 84:
```suggestion
  expect_setequal(as.vector(flow$partition1), 1:2)
```

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  expect_equal(sort(as.vector(flow$partition2)), 3:6)
+})
+
+test_that("vertex_connectivity works", {
```
maelle, line 88:
```suggestion
test_that("vertex_connectivity() works", {
```

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  g_disconnect <- make_graph(edges = c(1, 2, 3, 4), directed = FALSE)
+  expect_equal(vertex_connectivity(g_disconnect), 0)
+
+  g_ring <- make_ring(5, circular = TRUE)
+  expect_equal(vertex_connectivity(g_ring, source = 1, target = 4), 2)
+})
+
+test_that("vertex_connectivity error works", {
+  g_path <- make_ring(5, circular = FALSE)
+  expect_error(vertex_connectivity(g_path, source = 1))
```
maelle, line 101: snapshot or classed errors

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  expect_equal(ec1, 2)
+  ec2 <- edge_connectivity(g_kite, source = "Garth", target = "Andre")
+  expect_equal(ec2, 4)
+  ec3 <- edge_connectivity(g_kite, source = "Garth", target = "Ike")
+  expect_equal(ec3, 1)
+})
+
+test_that("edge_connectivity error works", {
+  g_path <- make_ring(5, circular = FALSE)
+  expect_error(edge_connectivity(g_path, source = 1))
```
maelle, line 139: snapshot or classed errors

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  )
+})
+
+test_that("min_st_separators works", {
+  g_zachary <- make_graph("Zachary")
+  msts <- min_st_separators(g_zachary)
+  is <- sapply(msts, is_separator, graph = g_zachary)
+  expect_equal(unique(is), TRUE)
+
+  ## TODO: check that it is minimal
```
maelle, line 228: is this hard / still valid?

maelle (commented):
:rocket:

R/flow.R
```diff
@@ -287,27 +286,21 @@ dominator.tree <- function(graph, root, mode = c("out", "in", "all", "total")) {
...
+    capacity <- E(graph)$capacity
   }
-  if (length(target) == 0) {
-    target <- NULL
-  }
-  if (is.null(source) && !is.null(target) ||
-    is.null(target) && !is.null(source)) {
-    stop("Please give both source and target or neither")
+
+  if (xor(is.null(source), is.null(target))) {
```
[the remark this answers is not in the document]
schochastics: The only reason I had to switch to it was that it is shorter and expresses what we want more clearly. 
``` r
xor
#> function (x, y) 
#> {
#>     (x | y) & !(x & y)
#> }
#> <bytecode: 0x10697add8>
#> <environment: namespace:base>
```
But now I am note sure if it is 100% ok to use, given that it uses single `|` and `&`? Could this be an issue? I think not

R/flow.R
```diff
@@ -551,22 +530,17 @@ edge_connectivity <- function(graph, source = NULL, target = NULL, checks = TRUE
...
   } else {
-    stop("either give both source and target or neither")
+    cli::cli_abort("Either give both source and target or neither")
   }
 }
 
 #' @rdname edge_connectivity
 #' @export
-edge_disjoint_paths <- function(graph, source, target) {
+edge_disjoint_paths <- function(graph, source = NULL, target = NULL) {
```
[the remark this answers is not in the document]
schochastics: Yes this was just to align the code with the rest. But I don't mind changing it back if you think it is better to leave it untouched

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  g_ring <- make_ring(5, directed = TRUE, circular = TRUE)
+  E(g_ring)$capacity <- c(1, 2, 3, 4, 5)
+  expect_error(min_cut(g_ring, source = 1))
+  expect_error(min_cut(g_ring, target = 1))
+})
+
+test_that("st_cuts works", {
+  g_path <- graph_from_literal(a - +b - +c - +d - +e)
+  all_cuts_path <- st_cuts(g_path, source = "a", target = "e")
+  expect_equal(unvs(all_cuts_path$cuts), list(1, 2, 3, 4))
```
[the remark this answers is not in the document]
schochastics: yes thats a good idea. any good suggestion (I am bad with naming things)

tests/testthat/test-flow.R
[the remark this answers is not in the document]
schochastics: Looks indeed neat, but I think I will go with snapshot for now

tests/testthat/test-flow.R
```diff
@@ -0,0 +1,265 @@
...
+  expect_error(st_min_cuts(g_path, source = "a", target = NULL))
+  expect_error(st_min_cuts(g_path, source = NULL, target = "a"))
+})
+
+test_that("max_flow works", {
+  edge_mat <- rbind(c(1, 3, 3), c(3, 4, 1), c(4, 2, 2), c(1, 5, 1), c(5, 6, 2), c(6, 2, 10))
+  colnames(edge_mat) <- c("from", "to", "capacity")
+  g_ring_acyc <- graph_from_data_frame(as.data.frame(edge_mat))
+  flow <- max_flow(g_ring_acyc, source = "1", target = "2")
+  expect_equal(flow$value, 2)
```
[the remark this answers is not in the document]
schochastics: I always felt quite ok with the dollar in igraph 🙈. But maybe we can keep it in mind for a later iteration

tests/testthat/test-flow.R
schochastics:
``` r
library(igraph)
g_zachary <- make_graph("Zachary")
msts <- min_st_separators(g_zachary)
sapply(msts, is_min_separator, graph = g_zachary)
#>  [1]  TRUE  TRUE  TRUE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
#> [13]  TRUE  TRUE FALSE FALSE FALSE FALSE  TRUE  TRUE  TRUE FALSE FALSE FALSE
#> [25] FALSE FALSE FALSE FALSE  TRUE  TRUE  TRUE  TRUE FALSE FALSE FALSE FALSE
#> [37] FALSE FALSE FALSE
```
@szhorvat not too familiar with the topic. Is the output problematic? shouldnt all be TRUE?

szhorvat: I have not verified all of this but it should be correct. 

This is indeed tricky and a bit confusing.

All of these are separators in the sense that their removal makes some vertex u unreachable from some vertex v.

"Minimal" in general means that we can't shrink the set while maintaining its property. We do indeed need to remove all these vertices to separate parts of the graph. But here's a quirk:

`msts[[4]]` is `1,3,14,20,31`. You drop any of 3, 14, 20 and 31 and this would still be a serparator, so it's not a minimal separator.

However, there are two vertices, for example `2` and `33`, which can be separated from each other only when removing _all_ of `1,3,14,20,31`. Dropping any one of them will no longer separate `2` from `33`. Thus it actually a _minimal_ separator with respect to some s-t pair (in this case s=2 t=33 is an example).

See the docs here for an example: https://igraph.org/c/html/latest/igraph-Separators.html#igraph_all_minimal_st_separators

schochastics: Ok thanks for the clarification! Missed the part that they are minimal for *some* vertex pair. I got it now!

szhorvat: See also https://github.com/igraph/igraph/issues/2512

PR #1696, opened 2025-02-19
chore: run {styler} (with {igraph.style})

maelle:
Some recent commits undid the style of literals (cc @schochastics), I set out to fix that and it became easier to just run styler on the whole package.


schochastics:
Hmm I gotta pay better attention on that in the future. Thanks for fixing

maelle:
No worries!

maelle:
It was a good opportunity to style things a bit

PR #1699, opened 2025-02-19
feat: hrg functions check their argument

### LINKED ISSUE #1074: HRG functions do not check arguments
### What happens?

Functions that take an `igraphHRG` object do not check that their input is indeed an `igraphHRG`. This can lead to crashes or worse.

### To Reproduce

```
> hrg_tree(make_ring(5))
Error in .subscript.2ary(x, i, , drop = drop) : subscript out of bounds
```

```
> hrg_tree(123)
IGRAPH 2d6cb1e D--- 3 2 -- 
+ attr: probability (v/n)
+ edges from 2d6cb1e:
[1] 3->1 3->1
```

(wrong output, and out-of-bounds memory access)

```
sample_hrg(123)
```

(typically crashes)

```
> sample_hrg(make_ring(5))
Error in .subscript.2ary(x, i, , drop = drop) : subscript out of bounds
```

maelle:
Fix #1074

@szhorvat for info I could only find two functions with HRG as input (well there are methods but methods don't need to check the class if I follow correctly)

### DIFF
--- a/R/hrg.R
+++ b/R/hrg.R
@@ -289,7 +289,7 @@ hrg <- hrg_create_impl
 hrg_tree <- function(hrg) {
 
   if (!inherits(hrg, "igraphHRG")) {
-    cli::cli_abort("{.arg hrg} must be a {.cls igraphHRG} object, not {.obj_type_friendly {hrg}}.")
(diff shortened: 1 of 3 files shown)


schochastics (approved):
Looks good to me!

PR #1715, opened 2025-02-24
refactor: one less tmp use case

maelle:
Related to #1328


schochastics:
@maelle it seems this change makes my PRs fail for R 4.0? https://github.com/igraph/rigraph/actions/runs/13567024838/job/37925237675?pr=1720 Do you think this is temporary or do we need to switch to `function(x)`?

schochastics:
testing with #1722

maelle:
oops so sorry

schochastics:
no worries, should be fixed now. I also missed it

maelle:
Thanks!

PR #1716, opened 2025-02-24
feat: added proper support for adding attributes via data.frames

### LINKED ISSUE #1373: do these tests need to be revisited or are they good so
https://github.com/igraph/rigraph/blob/d2095f33038449ee4a93eef13152c018aa408d52/tests/testthat/test-attributes.R#L256

and the two similar ones right after it.

### LINKED ISSUE #1669: Unclear error when storing data.frame as edge attributes

> Not sure if these are bugs, or just user experiences that result in unclear behaviors. Reporting primarily for visibility.

When a user accidentally uses a `data.frame` for edge attributes, many operations continue to work fine but errors will arise when subsetting when indices can't be recycled. Arises from this line, which iterates over (an assumed `list`) of attributes, subsetting with the indices. 

https://github.com/igraph/rigraph/blob/2a84bc2db90e6c5055e44925f00e14b7670883cf/R/iterators.R#L964

When the attributes are instead a `data.frame`, assignment with a value that can't be recycled to the data frame row count will throw an error.

When trying to create edge attributes with a `data.frame`, it would be very helpful if it was run through `as.list` automatically to avoid these pitfalls or if an error was raised to warn that this is unsupported. 

### To reproduce

```r
library(igraph)  # v2.1.4

g <- make_tree(10, 3)
n_edges <- length(E(g))

# whoops! igraph does not support data.frames
igraph::edge.attributes(g) <- head(mtcars, n_edges)

# but they seem to work fine in many cases, easy to miss wrong behavior
E(g)[cyl == 4]
#> + 3/9 edges from 2eeb173:
#> [1] 1-> 4 3-> 9 3->10

E(g)[1:3]
#> + 3/9 edges from 2eeb173:
#> [1] 1->2 1->3 1->4

E(g)[c(1, 1)]
#> + 2/9 edges from 2eeb173:
#> Error in `[[<-.data.frame`(`*tmp*`, i, value = c(21, 21)) :
#>  replacement has 2 rows, data has 9
```

### System information

```
R version 4.4.1 (2024-06-14)
Platform: aarch64-unknown-linux-gnu
Running under: Ubuntu 22.04.4 LTS (via rocker/r-ver image on podman)

Matrix products: default
BLAS:   /usr/lib/aarch64-linux-gnu/openblas-pthread/libblas.so.3 
LAPACK: /usr/lib/aarch64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so;  LAPACK version 3.10.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

time zone: Etc/UTC
tzcode source: system (glibc)

attached base packages:

other attached packages:
[1] igraph_2.1.4

[1] compiler_4.4.1  magrittr_2.0.3  cli_3.6.3       lifecycle_1.0.4
[5] pkgconfig_2.0.3 rlang_1.1.5    
```

schochastics:
Fix #1669 
Fix #1373 

The PR primarily adds the proper support for adding data.frames as attributes without potential error messages. Additionally, it removes some unnecessary attribute tests and generally refactors attribute tests


maelle:
awesome :rocket:

PR #1729, opened 2025-02-28
feat: rework is_bipartite to do an actual bipartivity check

### LINKED ISSUE #709: is_bipartite() is confusing / problematic
Before you read further, please guess what you think the `is_bipartite()` function does, if you are not already familiar with it.

----

No, it's not the same as `igraph_is_bipartite()` in C, `.is_bipartite()` in Python or `IGBipartiteQ[]` in Mathematica. Those functions decide if the graph is bipartite (i.e. 2-colourable) in the graph theoretical sense. 

`is_bipartite()` in R simply checks whether a graph has a `type` vertex attribute. This is documented, but in a way that is somewhat confusing, and easy to miss. It's at the bottom of [the doc page](https://r.igraph.org/reference/make_bipartite_graph.html):

> is_bipartite() checks whether the graph is bipartite or not. It just checks whether the graph has a vertex attribute called type.

In plain words, the term "bipartite" in R/igraph is used differently than just about anywhere else. We should consider changing this terminology in igraph 2.0, so it lines up better with common usage.

2-colourability is checked by `bipartite_mapping()`.

schochastics:
Fix #709 
A suggestion how a reworked is_bipartite could look like that does not (or should not) break existing code. 

``` r
library(igraph)

#bipartite no type
is_bipartite(make_ring(10))
#> Warning: `graph` is bipartite but has no 'type' vertex attribute. You can add it using:
#> `V(graph)$type <- bipartite_mapping(graph)$type`
#> [1] TRUE

# not bipartite but type
g <- make_full_graph(4)
V(g)$type <- c(TRUE,TRUE,FALSE,FALSE)
is_bipartite(g)
#> Warning: `graph` has a type attribute but is not actually bipartite.
#> [1] FALSE

#bipartite and type
g <- make_ring(4)
V(g)$type <- c(TRUE,FALSE,TRUE,FALSE)
is_bipartite(g)
#> [1] TRUE

#bipartite wrong type
g <- make_ring(4)
V(g)$type <- c(TRUE,TRUE,FALSE,FALSE)
is_bipartite(g)
#> Warning: `graph` is bipartite but has a wrong 'type' vertex attribute. You can correct
#> it using: `V(graph)$type <- bipartite_mapping(graph)$type`
#> [1] TRUE
```

<sup>Created on 2025-02-28 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

@szhorvat, thoughts?

### DIFF
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -1006,6 +1006,8 @@ is_weighted <- function(graph) {
 #' @param graph The input graph
 #' @export
 is_bipartite <- function(graph) {
+  ensure_igraph(graph)
+
   has_type_attr <- "type" %in% vertex_attr_names(graph)
 
   bipartite_result <- bipartite_mapping(graph)
--- a/R/attributes.R
+++ b/R/attributes.R
@@ -1006,10 +1006,30 @@ is_weighted <- function(graph) {
 #' @param graph The input graph
 #' @export
 is_bipartite <- function(graph) {
-  ensure_igraph(graph)
+  has_type_attr <- "type" %in% vertex_attr_names(graph)
 
-  "type" %in% vertex_attr_names(graph)
+  bipartite_result <- bipartite_mapping(graph)
+
+  if (has_type_attr) {
+    if (!bipartite_result$res) {
+      cli::cli_warn("{.arg graph} has a type attribute but is not actually bipartite.")
+      FALSE
+    } else {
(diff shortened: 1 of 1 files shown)


schochastics:
Thanks for your thoughtful reply! This makes a lot of sense and I think the topic does need more thinking about. I will close this PR for now and reopen eventually when there is a better idea on how to implement this.

PR #1731, opened 2025-03-03
fix: Recycling of logical vectors when indexing into edge/vertex selectors now throws an error

### LINKED ISSUE #848: Do not recycle logical vectors when indexing into edge/vertex selectors
With @krlmlr's agreement, the following behaviour is undesirable:

```r
> g <- make_ring(5) # graph with 5 vertices
> V(g)[c(T,F)] # recycles, completing the vector to T,F,T,F,T
+ 3/5 vertices, from 4cd3034:
[1] 1 3 5
```

Reasoning:

Suppose that we have a graph `g` and a modified version `g2`. Then we try to select some vertices of `g` based on properties of `g2`, like so: `V(g)[ V(g2)$foo > 2 ]`.  If `g` and `g2` have a different number of vertices, this should ideally result in an error. Right now it does not, due to recycling.

schochastics:
Fix #848 

Did not remove recycling for length 1 vectors because that still makes sense in my opinion

``` r
pkgload::load_all("~/git/R_packages/rigraph/")
#> ℹ Loading igraph
g <- make_ring(5)
V(g)[c(TRUE, FALSE)]
#> Error in `FUN()`:
#> ! Error: Logical index length does not match the number of vertices.
#>   Recycling is not allowed.
E(g)[c(TRUE, FALSE)]
#> Error in `FUN()`:
#> ! Error: Logical index length does not match the number of edges.
#>   Recycling is not allowed.

V(g)[TRUE]
#> + 5/5 vertices, from bafeef3:
#> [1] 1 2 3 4 5
E(g)[FALSE]
#> + 0/5 edges from bafeef3:
```

<sup>Created on 2025-03-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/iterators.R
+++ b/R/iterators.R
@@ -676,6 +676,10 @@ simple_vs_index <- function(x, i, na_ok = FALSE) {
       if (is.null(ii)) {
         return(NULL)
       }
+      if (is.logical(ii) && length(ii) != length(x)) {
+        cli::cli_abort("Error: Logical index length does not match the number of vertices. Recycling is not allowed.")
+      }
+
       ii <- simple_vs_index(x, ii, na_ok)
       attr(ii, "env") <- attr(x, "env")
       attr(ii, "graph") <- attr(x, "graph")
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks!

PR #1732, opened 2025-03-03
fix: changed base location for graph_from_graphdb and added tests

### LINKED ISSUE #1712: How useful is `graph_from_graphdb()`?
Just noticed this function in `R/foreign.R`, that does not have any test. In the CRAN mirror I only see it in igraph repo https://github.com/search?q=graph_from_graphdb+user%3Acran&type=code&ref=advsearch

Is it useful or should it be deprecated?

schochastics:
Fix #1712
Downloaded the database from here: https://zenodo.org/records/11204020
and hosted it here: https://github.com/schochastics/graphsdb/

General question: should we host it somewhere else?

### DIFF
--- a/R/foreign.R
+++ b/R/foreign.R
@@ -597,10 +597,16 @@ write.graph.dot <- function(graph, file, ...) {
 #' @export
 #' @keywords graphs
 graph_from_graphdb <- function(
-    url = NULL,
-    prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
-    base = "https://github.com/igraph/graphsdb/raw/refs/heads/main",
-    compressed = TRUE, directed = TRUE) {
+  url = NULL,
+  prefix = "iso",
+  type = "r001",
+  nodes = NULL,
+  pair = "A",
+  which = 0,
+  base = "https://github.com/igraph/graphsdb/raw/refs/heads/main",
+  compressed = TRUE,
+  directed = TRUE
+) {
   if (is.null(nodes) && is.null(url)) {
     cli::cli_abort("Either {.arg nodes}' or `{.arg url}' must be non-null.")
   }
--- a/R/foreign.R
+++ b/R/foreign.R
@@ -596,12 +596,13 @@ write.graph.dot <- function(graph, file, ...) {
 #' @family foreign
 #' @export
 #' @keywords graphs
-graph_from_graphdb <- function(url = NULL,
-                               prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
-                               base = "http://cneurocvs.rmki.kfki.hu/graphdb/gzip",
-                               compressed = TRUE, directed = TRUE) {
+graph_from_graphdb <- function(
+    url = NULL,
+    prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
+    base = "https://github.com/schochastics/graphsdb/raw/refs/heads/main",
+    compressed = TRUE, directed = TRUE) {
   if (is.null(nodes) && is.null(url)) {
-    stop("The `nodes' or the `url' argument must be non-null")
+    cli::cli_abort("Either {.arg nodes}' or `{.arg url}' must be non-null.")
   }
 
   if (is.null(url)) {
@@ -622,10 +623,10 @@ graph_from_graphdb <- function(url = NULL,
     typegroup <- typegroups[which(types == type)]
 
     if (!prefix %in% prefixes) {
-      stop("Invalid prefix!")
+      cli::cli_abort("{prefix} is not a valid prefix. Must be one of {prefixes}.")
     }
     if (!type %in% types) {
-      stop("Invalid graph type!")
+      cli::cli_abort("{type} is not a valid graph type. Must be one of {types}.")
     }
     suff <- if (compressed) ".gz" else ""
     filename <- paste(
@@ -641,7 +642,7 @@ graph_from_graphdb <- function(url = NULL,
 
   f <- try(gzcon(file(filename, open = "rb")))
   if (inherits(f, "try-error")) {
-    stop(paste("Cannot open URL:", filename))
+    cli::cli_abort("Cannot open URL:{.arg filename}")
   }
 
   buffer <- read.graph.toraw(f)
--- a/R/foreign.R
+++ b/R/foreign.R
@@ -629,12 +629,15 @@ graph_from_graphdb <- function(
     typegroup <- typegroups[which(types == type)]
 
     if (!prefix %in% prefixes) {
-      cli::cli_abort("{prefix} is not a valid prefix. Must be one of {prefixes}.")
+      cli::cli_abort(c(
+        "{.value {prefix}} is not a valid prefix.",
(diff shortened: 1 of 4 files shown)


R/foreign.R
```diff
@@ -596,12 +596,13 @@ write.graph.dot <- function(graph, file, ...) {
 #' @family foreign
 #' @export
 #' @keywords graphs
-graph_from_graphdb <- function(url = NULL,
-                               prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
-                               base = "http://cneurocvs.rmki.kfki.hu/graphdb/gzip",
-                               compressed = TRUE, directed = TRUE) {
+graph_from_graphdb <- function(
```
maelle, line 599: is this reformatting due to Air? :smile_cat:

R/foreign.R
```diff
@@ -622,10 +623,10 @@ graph_from_graphdb <- function(url = NULL,
     typegroup <- typegroups[which(types == type)]
 
     if (!prefix %in% prefixes) {
-      stop("Invalid prefix!")
+      cli::cli_abort("{prefix} is not a valid prefix. Must be one of {prefixes}.")
     }
     if (!type %in% types) {
-      stop("Invalid graph type!")
+      cli::cli_abort("{type} is not a valid graph type. Must be one of {types}.")
```
maelle, line 629: maybe split this and the message above into two lines?
```suggestion
      cli::cli_abort(c(
      "{type} is not a valid graph type.",
      i = "Must be one of {types}."
    ))
```

R/foreign.R
```diff
@@ -596,12 +596,13 @@ write.graph.dot <- function(graph, file, ...) {
...
 #' @export
 #' @keywords graphs
-graph_from_graphdb <- function(url = NULL,
-                               prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
-                               base = "http://cneurocvs.rmki.kfki.hu/graphdb/gzip",
-                               compressed = TRUE, directed = TRUE) {
+graph_from_graphdb <- function(
+    url = NULL,
+    prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
+    base = "https://github.com/schochastics/graphsdb/raw/refs/heads/main",
```
maelle, line 602: could we instead use a permalink from the Zenodo record?

maelle (commented):
good idea!

I see a typo in the file "algothitms", it could be fixed at the same time.

I wonder whether the database should be cached in an user dir. Should the argument of where to save it be exposed?

R/foreign.R
schochastics: The zenodo link is unfortunately the whole db. You cannot just download a single file which is need for the function

maelle: Oh ok, sorry, makes sense, thank you.

maelle: Would we be allowed to store a copy in https://github.com/igraph/igraphdata?

szhorvat:
> and hosted it here: https://github.com/schochastics/graphsdb/

I suggest hosting it under the igraph organization for the purposes of allowing the function to work.

In our own copy, we should mention the license and list the original authors, _in addition_ to linking to Zenodo. I'm really glads to see that this is on Zenodo given that the MIVIA website is gone now.

@maelle 

> Would we be allowed to store a copy in https://github.com/igraph/igraphdata?

This dataset is way too large for the igraphdata package. I suggest using a separate repo, exactly what @schochastics did, but under the igraph organization. I believe you _should_ be able to create a new repo, but if not let me know and I'll see if I can fix that.

----

Can this function operate with a local copy of the database? Would this be easy to implement, and if yes, can we include an explanation of how to do it? `file:///` URL in `base`?

----

Other than the comments above, this LGTM.

R/foreign.R
schochastics: Yes :D. Should I revert it?

R/foreign.R
schochastics: can confirm, the function works with a local copy of the database

schochastics:
I will try to host it under the igraph org and put all the necessary information into the README. The files have a CC license so that should all be fine. I also checked that it works with a local copy. You simply need to provide the path to the database. I will also update the docs so that this is clear

R/foreign.R
maelle: No! I was just curious.

R/foreign.R
```diff
@@ -596,12 +596,13 @@ write.graph.dot <- function(graph, file, ...) {
...
 #' @keywords graphs
-graph_from_graphdb <- function(url = NULL,
-                               prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
-                               base = "http://cneurocvs.rmki.kfki.hu/graphdb/gzip",
-                               compressed = TRUE, directed = TRUE) {
+graph_from_graphdb <- function(
+    url = NULL,
+    prefix = "iso", type = "r001", nodes = NULL, pair = "A", which = 0,
+    base = "https://github.com/igraph/graphsdb/raw/refs/heads/main",
+    compressed = TRUE, directed = TRUE) {
```
krlmlr, line 603:
```suggestion
graph_from_graphdb <- function(
  url = NULL,
  prefix = "iso",
  type = "r001",
  nodes = NULL,
  pair = "A",
  which = 0,
  base = "https://github.com/igraph/graphsdb/raw/refs/heads/main",
  compressed = TRUE,
  directed = TRUE
) {
```

R/foreign.R
```diff
@@ -622,10 +629,13 @@ graph_from_graphdb <- function(url = NULL,
     typegroup <- typegroups[which(types == type)]
 
     if (!prefix %in% prefixes) {
-      stop("Invalid prefix!")
+      cli::cli_abort("{prefix} is not a valid prefix. Must be one of {prefixes}.")
```
krlmlr, line 632:
```suggestion
      cli::cli_abort(c(
        "{.value {prefix}} is not a valid prefix.",
        i = "Must be one of {.value {prefixes}}."
      ))
```

R/foreign.R
```diff
@@ -622,10 +629,13 @@ graph_from_graphdb <- function(url = NULL,
...
     }
     if (!type %in% types) {
-      stop("Invalid graph type!")
+      cli::cli_abort(c(
+        "{type} is not a valid graph type.",
+        i = "Must be one of {types}."
+      ))
```
krlmlr, line 638:
```suggestion
      cli::cli_abort(c(
        "{.value {type}} is not a valid graph type.",
        i = "Must be one of {.value {types}}."
      ))
```

R/foreign.R
```diff
@@ -641,7 +651,7 @@ graph_from_graphdb <- function(url = NULL,
 
   f <- try(gzcon(file(filename, open = "rb")))
   if (inherits(f, "try-error")) {
-    stop(paste("Cannot open URL:", filename))
+    cli::cli_abort("Cannot open URL:{.arg filename}")
```
krlmlr, line 654:
```suggestion
    cli::cli_abort("Cannot open URL provided in {.arg filename}: {.url {filename}}")
```

krlmlr (commented):
Do these actually work and look better? Please merge when good and green.

PR #1735, opened 2025-03-06
chore: Remove `Rf_allocSExp()` which is no longer in R's C API

krlmlr:


szhorvat:
Can you change the title to "not in R's C API" or similar?

PR #1745, opened 2025-03-10
chore: format with air

### LINKED ISSUE #1734: protect literals from Air

maelle:
Fix #1734

I need to find the `~`as well. This only works with Air from the PR mentioned in the issue.


schochastics:
closing in favor of a new one. let me know if this should remain open

PR #1747, opened 2025-03-12
chore: require copyright assignment in PR template

szhorvat:
See https://github.com/igraph/igraph/pull/2720

This is meant to apply to external contributors. Team members should feel free to delete the text when opening PRs.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,5 @@
+<!-- Please disclose any use of LLMs (ChatGPT, Copilot, Gemini, etc.) during the preparation of this PR. -->
+
+<!-- The following text must be left intact and the box checked before the PR can be merged. -->
+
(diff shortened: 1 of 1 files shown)


krlmlr:
Why is the disclosure of LLM usage important?

szhorvat:
Good catch, that's actually included accidentally by copying the C core PR template verbatim. That said, I do think it's a good idea, and the C core has had this for a while now.

LLM are not only unreliable, but the mistakes they make are often very different from what a human would make. This makes it harder to catch errors in LLM-generated work. Being aware that AI tools were used is useful when reviewing a PR, whether it's code, documentation or something else that was AI-generated. This is doubly true in the C library, where the core algorithms (often having subtle yet important details) are implemented.

We added this line almost a year ago, and the experience I had with AI since then convinces me that this is necessary. I ask the same from my students. 

I don't insist on adding this line. Just let me know what you prefer.

krlmlr:
Thanks! Let's try it out.

PR #1794, opened 2025-04-03
docs: Document return value of `make_clusters()`

krlmlr:
This needs manual review and extension.


R/community.R
```diff
@@ -550,8 +550,11 @@ print.communities <- function(x, ...) {
 #'   is `TRUE` and the membership vector is available, then it the
 #'   modularity values is calculated automatically.
 #' @return A `communities` object.
-#'
-#'
+#'   \describe{
+#'     \item{membership}{A numeric vector giving the community id for each
+#'       vertex.}
+#'     \item{modularity}{The modularity score of the partition.}
+#'   }
```
schochastics, line 557:
```suggestion
#'     \item{algorithm}{If known, the algorithm used to obtain the communities.}
#'.    \item{vcount}{Number of vertices in the graph.}
#'   }
```

R/topology.R
```diff
@@ -774,7 +774,11 @@ graph.isoclass <- isoclass_impl
 #' @param directed Whether to create a directed graph (the default).
 #' @return An igraph object, the graph of the given size, directedness
 #'   and isomorphism class.
-#'
+#' @return A list with components:
+#'   \describe{
+#'     \item{topology}{The topology of the graph.}
+#'     \item{details}{Additional details about the topology.}
+#'   }
```
schochastics, line 781:
```suggestion
```

THis return value does not make any sense. There is only a graph returned as written in the return doc above

schochastics (changes requested):
I corrected some errors that need to be checked and committed

schochastics (approved):

Issue #1793, opened 2025-04-03
Alert downstream maintainers of breaking changes


krlmlr:
<html>
<head>

</head>
<body>

<p class="p1">The reverse dependency failures across the R packages you listed share several <span class="s1"><b>recurring patterns</b></span> that strongly suggest <span class="s1"><b>a common set of causes</b></span>, primarily due to <span class="s1"><b>recent changes in the igraph package</b></span>, particularly in <span class="s1"><b>version 2.1.5</b></span>. Here’s a summary of the <span class="s1"><b>main patterns</b></span> and <span class="s1"><b>root causes</b></span>:</p>
<p class="p2"><span class="s2"><hr></span></p>
<p class="p2"><span class="s2"><h3><b>✅<span class="Apple-converted-space"> </span></b></h3><h3><b>Common Technical Patterns</b></h3></span></p>
<p class="p3"><br></p>
<p class="p2"><span class="s2"><h4><b>1.<span class="Apple-converted-space"> </span></b></h4><h4><b>Error in paste0() with type 'closure'</b></h4></span></p>
<p class="p2"><span class="s2"><ul><li>
<p class="p1"><span class="s1"><b>Packages affected</b></span><span class="s2">: </span>alakazam<span class="s2">, </span>arulesViz<span class="s2">, </span>multivariance<span class="s2">, others.</span></p>
</li><li>
<p class="p1"><b>Symptom</b><span class="s1">:</span></p>
</li></ul></span></p>

<pre><code>Error in paste0(before, x, after) : cannot coerce type &#39;closure&#39; to vector of type &#39;character&#39;</code></pre>

<p class="p1"><span class="s1"><ul><li>
<p class="p1"><br></p>
</li><li>
<p class="p1"><span class="s1"><b>Likely cause</b></span>: A <span class="s1"><b>function name is being used instead of a variable</b></span> due to namespace conflict or unintended variable masking. This is often triggered by how <span class="s2">igraph</span> or <span class="s2">cli</span> formats output with default attributes.</p>
</li></ul></span></p>
<p class="p1"><span class="s1"><hr></span></p>
<p class="p1"><span class="s1"><h4><b>2.<span class="Apple-converted-space"> </span></b></h4><h4><b>get_edge_ids() and get.edge.ids() 2x2 matrix error</b></h4></span></p>
<p class="p1"><span class="s1"><ul><li>
[...]

PR #1798, opened 2025-04-04
fix: `get_edge_ids()`  error for 2x2 matrices

schochastics:
The function now throws a warning and indicates how the matrix is interpreted for backward compatibility


krlmlr:
I see that only three packages are affected: corpustools, mwcsr and SEMID. One of the results also hints at a misuse from igraph itself. I still think we should break cleanly and perhaps improve the error message.

krlmlr:
Another alternative: disallow matrices altogether, ask users to convert matrices with `c()` or to pass data frames if they have data in columns.

schochastics:
Fine by me, lets find a way in this PR

krlmlr:
Let's await revdepchecks in #1793.

R/interface.R
```diff
@@ -478,10 +478,12 @@ el_to_vec <- function(x, call = rlang::caller_env()) {
     nrow <- dimx[[1]]
     ncol <- dimx[[2]]
     if (nrow == 2 && ncol == 2) {
-      lifecycle::deprecate_stop(
+      lifecycle::deprecate_warn(
         "2.1.5",
-        "get_edge_ids(vp = 'is not allowed to be a 2 times 2 matrix')"
+        "get_edge_ids(vp = 'is not allowed to be a 2 times 2 matrix')",
+        details = c("For backward compatibility, edges are interpreted columnwise.", "convert to a data.frame with two columns and one row per edge.")
```
maelle, line 484: The second message (one row per edge) is clearer. Could the two be combined? My suggestion below is very bad though.

```suggestion
        details = c("For backward compatibility, it is assumed that each row is an edge, from the first column to the second column.", "convert to a data.frame with two columns and one row per edge.")
```

schochastics:
PRs send to the 3 packages. Should we close this PR then?

krlmlr:
Closing, let downstream break.

PR #1803, opened 2025-04-04
fix: remove string matrix support from biadjacency matrix functions

### LINKED ISSUE #1540: attr in as_biadjacency_matrix()
@szhorvat is this expected:

``` r
library("igraph")
#> 
#> 
#>     decompose, spectrum
#> 
#>     union

withr::local_seed(42)

I <- matrix(sample(0:1, 9, replace = TRUE, prob = c(3, 1)), ncol = 3)

g <- graph_from_biadjacency_matrix(I)
E(g)$something <- runif(ecount(g))
as_biadjacency_matrix(g, sparse = TRUE, attr = "something")
#> 3 x 3 sparse Matrix of class "dgTMatrix"
#>           4         5 6
#> 1 0.7050648 0.4577418 .
#> 2 0.7191123 .         .
#> 3 .         .         .

g2 <- graph_from_biadjacency_matrix(I)
E(g2)$something <- letters[1:ecount(g)]
as_biadjacency_matrix(g2, sparse = TRUE, attr = "something")
#> Error in getClass(Class, where = topenv(parent.frame())): "gTMatrix" is not a defined class
```

<sup>Created on 2024-09-30 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

If so, we should add some details to the documentation about what `attr`can be depending on sparse/dense, and some argument checks inside the function.

### LINKED ISSUE #1542: Remove string matrix support from biadjacency matrix functions
For 2.0, we removed string matrix support from graph to adjacency matrix conversion functions (TODO: find discussion). They are not supported with either sparse or dense matrices.

```r
> g <- make_graph(c(1,2, 2,3), n=4)
> E(g)$foo <- c('a', 'b')
> A <- as_adjacency_matrix(g, attr='foo')
Error in get.adjacency.sparse(graph, type = type, attr = attr, names = names) : 
  Matrices must be either numeric or logical, and the edge attribute is not
```

We should also remove it for biadjacency matrix functions, for consistency.

----

Note that the current support with dense matrices is arguably broken:

```r
> as_biadjacency_matrix(g, types=c(T,F,T,F), attr='foo')
  1   3  
2 "a" "b"
4 "0" "0"
```

The string `"0"` makes no sense as the null element of a string matrix.

----

Sparse support does not work at all:

```
as_biadjacency_matrix(g, types=c(T,F,T,F), attr='foo', sparse=T)
Error in getClass(Class, where = topenv(parent.frame())) : 
  “gTMatrix” is not a defined class
```

----

Finally, the C core does not support string matrices (dense or sparse), and is very unlikely to ever get support. Ideally, we will rely on C core functions for all (bi)adacency matrix functionality, so we should stick to numeric matrices.

Note that string matrices are conceptually very different from numeric ones as it is unclear how to handle multi-edges and undirected self-loops, which require composing attribute values together for graph -> matrix conversions, and splitting for matrix -> graph conversions.

----

CC @maelle @krlmlr @schochastics

schochastics:
Fix #1542 
Fix #1540

@krlmlr, @maelle, @szhorvat: I took the extreme approach of hard deprecation. In my opinion, this is ok, because this should simply not be allowed and has produced an error for adjacency matrices for a while

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -857,7 +857,14 @@ get.incidence.dense <- function(graph, types, names, attr, call = rlang::caller_
     el[idx, ] <- el[idx, 2:1]
     # el[ ,1] only holds values 1..n1 and el[ ,2] values 1..n2
     # and we can populate the matrix
-    res[el] <- edge_attr(graph, attr)
+    value <- edge_attr(graph, attr)
+    if (!is.numeric(value) && !is.logical(value)) {
+      cli::cli_abort(
+        "Matrices must be either numeric or logical, and the edge attribute is not",
+        call = call
+      )
+    }
+    res[el] <- value
 
     if (names && "name" %in% vertex_attr_names(graph)) {
       rownames(res) <- V(graph)$name[which(!types)]
@@ -902,6 +909,12 @@ get.incidence.sparse <- function(graph, types, names, attr, call = rlang::caller
       cli::cli_abort("No such edge attribute", call = call)
     }
     value <- edge_attr(graph, name = attr)
+    if (!is.numeric(value) && !is.logical(value)) {
+      cli::cli_abort(
+        "Matrices must be either numeric or logical, and the edge attribute is not",
+        call = call
+      )
+    }
   } else {
     value <- rep(1, nrow(el))
   }


szhorvat:
I support this.

R/conversion.R
```diff
@@ -857,7 +857,14 @@ get.incidence.dense <- function(graph, types, names, attr, call = rlang::caller_
     el[idx, ] <- el[idx, 2:1]
     # el[ ,1] only holds values 1..n1 and el[ ,2] values 1..n2
     # and we can populate the matrix
-    res[el] <- edge_attr(graph, attr)
+    value <- edge_attr(graph, attr)
+    if (!is.numeric(value) && !is.logical(value)) {
+      cli::cli_abort(
+        "Matrices must be either numeric or logical, and the edge attribute is not",
```
maelle, line 863: add the type of the attribute? with friendly type etc

maelle (commented):
:rocket:

krlmlr:
Thanks! Let's tackle the comments in a follow-up.

PR #1816, opened 2025-04-09
chore: added code formatting settings for Positron/Air users

### LINKED ISSUE #1734: protect literals from Air

schochastics:
Fix #1734

the `.vscode` folder contains custom styling rule for positron (might need to see if this needs to change for developers who use vscode, but not positron)

`air.toml` configures [air](https://github.com/posit-dev/air) to skip messing with `graph_from_literal`.


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "[r]": {
+    "editor.defaultFormatter": "positron.positron-r",
+    "editor.formatOnType": true
```
DavisVaughan, line 4: These are actually not right. I'd encourage you to use `usethis::use_air()` in the development version of usethis instead

DavisVaughan: i.e. you want these instead

https://github.com/tidyverse/dplyr/blob/main/.vscode/settings.json
https://github.com/tidyverse/dplyr/blob/main/.vscode/extensions.json

schochastics: You are right and I realize now that I have to rethink this. I turned Air of for igraph on purpose, because it did reformat old styling that was not supposed to be changed on a regular basis. So putting the air.toml in here but turning air off in the settings doesn't make much sense

PR #1821, opened 2025-04-11
docs: add author links

maelle:
@schochastics or should we use the cynkra/about links :thinking:


schochastics:
I think the "private" link is ok

PR #1823, opened 2025-04-12
docs: link to replacements of deprecated functions

szhorvat:
This is a simple find-and-replace job to actually link to replacements of deprecated functions instead of simply mentioning them.

Is there a reason it wasn't done this way? Perhaps it is not standard in the R world?

Do we want ``[`some_fun()`]`` (inline code) or simply ``[some_fun()]`` (not in code style)? I see both styles used at the moment.

I am hoping this encourages people to upgrade to the non-deprecated alternative more quickly. I still see tutorials published with the old names ...

The search pattern was:

```
was renamed to (`.+?`)
```

### DIFF
--- a/R/adjacency.R
+++ b/R/adjacency.R
@@ -3,7 +3,7 @@
 #' @description
 #' `r lifecycle::badge("deprecated")`
 #'
-#' `graph.adjacency()` was renamed to [`graph_from_adjacency_matrix()`] to create a more
+#' `graph.adjacency()` was renamed to [graph_from_adjacency_matrix()] to create a more
 #' consistent API.
 #' @inheritParams graph_from_adjacency_matrix
 #' @keywords internal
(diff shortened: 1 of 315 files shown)


krlmlr:
Would you like to share the code that did the search+replace? We might run into conflicts, I'd also like to tweak a detail.

szhorvat:
There's no code. I did it with a text editor, and regex find/replace, so that I could review at least a subset of replacements before proceeding.

krlmlr:
I now see you asked, `[some_fun()]` is preferred and could be changed everywhere as part of this PR. I'll remember merging this when all other ready PRs are done to keep conflicts contained.

szhorvat:
> `[some_fun()]` is preferred

I can fix this without too much trouble.

szhorvat:
@krlmlr, this is done now.

Replacement across all files in `/R`:

``\[`(.+?\(\))`\]`` $\longrightarrow$ `[$1]`

I didn't notice anything going wrong.

But I did notice an unusual way to refer to a class (use of parentheses), which is also fixed now.

krlmlr:
Thanks!

PR #1829, opened 2025-04-16
fix: added factor support for `graph_from_data_frame()`

### LINKED ISSUE #34: R: graph.data.frame converts factors to character
_From @gaborcsardi on July 26, 2014 3:42_

Add an option to keep factors as factors. See 
http://stackoverflow.com/questions/24965840/igraph-graph-data-frame-silently-converts-factors-to-character-vectors

_Copied from original issue: igraph/igraph#665_

schochastics:
Fix #34


krlmlr:
Thanks!

PR #1832, opened 2025-04-17
chore: refer to the dev version of roxygen2 as build dependency

szhorvat:
As suggested by @krlmlr in chat.


krlmlr:
Thanks! Looking at the sanitizer now so that checks are greener again.

PR #1833, opened 2025-04-18
test: don't `library(Matrix)` in tests

MichaelChirico:

`library()` has some pretty "strong" side-effects, better to avoid it. I don't think it's needed anyway -- `skip_if_not_installed()` will do `requireNamespace()` which will have the more "minor" side-effect of loading {Matrix} namespace.


krlmlr:
Thanks!

PR #1869, opened 2025-06-10
chore: add air code formatting

schochastics:
fix #1734

PR #1872, opened 2025-06-10
feat!: change arguments default and order for `graph_from_lcf()`

### LINKED ISSUE #1858: `graph_from_lcf()` should have default for the `n` parameter
I'm not sure how to implement this without a breaking change, but it would make a lot of sense for the `n` parameter of `graph_from_lcf` to default to `len(shifts) * repeats`. This is consistent with the standard notation and in almost no case would one want anything else.

maelle:
Fix #1858


maelle:
Yes I was in a really destructive mode. :sweat_smile:

R/make.R
```diff
@@ -2552,13 +2606,40 @@ full_citation_graph <- function(...)
...
 #'
 #' # This is the Franklin graph:
-#' g1 <- graph_from_lcf(12, c(5, -5), 6)
+#' g1 <- graph_from_lcf(shifts = c(5L, -5L), n = 12L, repeats = 6L)
 #' g2 <- make_graph("Franklin")
 #' isomorphic(g1, g2)
 #' @export
 #' @cdocs igraph_lcf_vector
-graph_from_lcf <- lcf_vector_impl
+graph_from_lcf <- function(
```
maelle, line 2614: the argument order changed. how bad is that? we had no test for the function :zany_face: 

Once we agree on the changes I'll remove formatting changes since those will be brought in by #1869 anyway.

schochastics:
I think I am fine with the change. Note #1869 is merged now I hope this doesnt cause too much trouble here

schochastics (approved):

Issue #219, opened 2017-08-10
induced.subgraph and subgraph.edges don't adjust layout if layout is matrix


ahmontgo:
I've been plotting multiple subgraphs while attempting to retain layout values, and came across an interesting bug. If g$layout is a matrix, vertex deletion doesn't modify it, which produces odd results. For example:

g <- make_ring(10)
g$layout <- layout_in_circle(g)
plot(induced.subgraph(g,c(1:9)))

The "ghost" node 10 is still plotted, picking up the label of the first vertex to boot!

Workaround is to manually remove the node from layout:
g2 <- induced.subgraph(g,1:9)
g2$layout <- g2$layout[1:9,]
plot(g2)

Similar but slightly more complex manual removal needs to occur when subgraph.edges is used:
g3 <- subgraph.edges(g,1:8)
g3$layout <- g3$layout[V(g)[inc(1:8)],]
plot(g3)

Expected behavior would be for plot() to ignore any "extra" vertices in layout. But depending on internal representation, there may be no way of doing so. In that case, subgraph.edges and induced.subgraph should probably modify any g$layout if g$layout is a matrix.

I'd offer a patch, but my C is very shaky.

Thanks!

ntamas:
Not planned for the time being for sake of backwards compatibility. If you want to adjust the layout when vertices are deleted, you can use vertex attributes named `x` and `y` instead of a graph-level `layout` attribute and this will treat vertex deletion correctly. These attributes are also recognized by `layout_nicely()`.

szhorvat:
I understand that there is no capacity for dealing with this now, but it seems to me that the behaviour is bad enough to be called a bug, which should ideally be tracked.

What the correct behaviour would be is debatable. This issue would provide a space to discuss that.

ntamas:
I'd probably just remove the special handling of the `layout` graph attribute from `layout_nicely()` (which is the only place where this attribute is treated as a layout matrix). Having a vertex-specific quantity stored as a _graph_ attribute is asking for trouble.

szhorvat:
Yes, I agree, and that is of course breaking, so it's probably for 2.0 only. My point is that either there should be a new issue for that or this one should be kept open so we do not forget about this problem that occurs with some usage patterns.

schochastics:
fixed in #1880

PR #1880, opened 2025-06-24
fix: Fail if `"layout"` attribute doesn't match the number of vertices

schochastics:
Trying to tackle https://github.com/igraph/rigraph/issues/219.
The overarching issue is that if the layout is saved as a graph attribute (`g$layout`), independent on how it is done, there is no checks implemented if the layout matrix can actually 
serve as a layout (that is, if it has the correct number of rows). 
This leads to weird behaviour of the plot function. 

This PR could be the starting point for a refactoring. Currently, it only adds a warning if the layout does not match the graph. @szhorvat could this be a good start?

``` r
devtools::load_all("~/git/R_packages/rigraph/")
#> ℹ Loading igraph
g <- make_ring(10)
g$layout <- layout_in_circle(g)
plot(induced_subgraph(g,c(1:9)))
#> Warning: The layout has 10 rows, but the graph has 9 vertices. Unintended results may
#> occur.
```

``` r

g <- make_full_graph(10)
g$layout <- layout_in_circle(g)[1:5,]
plot(g)
#> Warning: The layout has 5 rows, but the graph has 10 vertices. Unintended results may
#> occur.
```

``` r

g$layout <- rbind(layout_in_circle(g), cbind(runif(5),runif(5)))
plot(g)
#> Warning: The layout has 15 rows, but the graph has 10 vertices. Unintended results may
#> occur.
```

<sup>Created on 2025-06-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


szhorvat:
I think a complete fix is only possible in the C core. We need support for attributes with vector values. I proposed this here for 1.0:

https://github.com/igraph/igraph/issues/2744

But realistically, @ntamas and I just don't have the capacity to get this done for the 1.0 release, which is due in September.

My suggestion is to not put too much effort into this on the R side until it can be fixed in the C core. Perhaps discourage using `$layout` and suggest `$x`, `$y` instead ...

@ntamas Are there any hooks in the attribute handler that would allow updating such attributes upon vertex deletion/addition? I don't think there are any. Note that `$layout` is a _graph attribute_, which is the problem here.

krlmlr:
My question in https://github.com/igraph/rigraph/pull/1880#issuecomment-3007834034 is still relevant.

schochastics:
I'll investigate but I think this would be a big change. 
1) Matrix attributes are not supported for vertices and 2) The plot code assumes layout either to be a matrix graph attribute or x and y vertex attributes

krlmlr:
1. The code shows it’s possible to store them? 2. We own the plotting code?

David Schoch ***@***.***> schrieb am Sa. 5. Juli 2025 um
11:17:

krlmlr:
Specifically, this works with #1894:

``` r

g <- make_full_graph(10)
V(g)$layout <- layout_in_circle(g)
V(g)$layout
#>            [,1]          [,2]
#>  [1,]  1.000000  0.000000e+00
#>  [2,]  0.809017  5.877853e-01
#>  [3,]  0.309017  9.510565e-01
#>  [4,] -0.309017  9.510565e-01
#>  [5,] -0.809017  5.877853e-01
#>  [6,] -1.000000  1.224647e-16
#>  [7,] -0.809017 -5.877853e-01
#>  [8,] -0.309017 -9.510565e-01
#>  [9,]  0.309017 -9.510565e-01
#> [10,]  0.809017 -5.877853e-01
```

<sup>Created on 2025-07-05 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
Then let's make it happen. I could adapt the plot code to also check for a vertex layout attribute

szhorvat:
There are some issue with this approach. Since vector attributes are not supported in C yet, it will not be possible to export these to the standard formats such as GML.

In the long term it would be good to do this, but I'm wondering if it's a better approach to wait until the C core supports it, or at least wait until upgrading to C/igraph 1.0.  I have an issue open for this here: https://github.com/igraph/igraph/issues/2744  Realistically, we probably won't have the capacity to make it happen for 1.0, but I'd like to be optimistic and take another look ...

szhorvat:
(To be clear, IMO this is absolutely the right way to do it. My concern is making this an R-specific feature, especially before we had the opportunity to take a good look at it in C and determine what interface is feasible.)

krlmlr:
We could also make V(g)$layout <- … result in setting the x and y vertex
attributes. Or perhaps create a layout setting function.

I need to review the attribute code to give a better assessment.

szhorvat:
How I envisioned this working was exactly `V(g)$layout <- matrix`, and internally, in C, such an attribute would be stored precisely as a matrix. Some of the concerns I have, that I think we need to think through, are:

 - Issue with columns-major vs row-major storage in C. @krlmlr, can you confirm that R uses column-major (just like igraph)?
 - How well this fits various file formats?
 - We had some discussions with @ntamas about proposing a new format, but now that the NASCOL forum is dead (CC @vtraag) this'll be harder to push through.

And of course the biggest issue: Implementing this feature with the very limited capacity we have ...

krlmlr:
Thanks, merging.

PR #1881, opened 2025-06-24
fix: automatically arrange loops in `plot()`

### LINKED ISSUE #407: Plotting: multiple self-loops on a vertex are plotted on top of each other
When a vertex has multiple self-loops, they are plotted on top of each other. The graphs `1-1` and `1-1, 1-1` are not visually distinguishable when plotted.

`g1` and `g2` look the same:

```
library(igraph)
g1 <- make_graph(c(1,1))
g2 <- make_graph(c(1,1, 1,1))
plot(g1)
plot(g2)
```

Related: https://github.com/igraph/python-igraph/issues/321 (see there for a suggestion about how this might be plotted nicely)

### LINKED ISSUE #556: Nice to have in layout_in_circle: always plot self edges outside the circle.
In layout_in_circle: always plot self edges outside the circle.

**Use cases for the feature**
Example.
```
Berger/circle schedule for round robin tournaments
 n <- 32
 n <- max(n - -!(n %%2L), 1)
 r <- c(rbind((seq(from = 0, to = -(n+1)/2+1) %% n) , seq(from=0, to=(n-1)/2)))  # diagonal in Zn
 g <- make_graph(r+1)
  
for (i in seq_len(n-1)) {
   r <- ((r+(n-1)/2) %% n)
   g <- add_edges(g, r+1)
 }
 plot(g, layout=layout_in_circle)
```
For n = 8, the self-edges within the circle are visible.
For larger n, the edges become unrecognizable.
**References**

schochastics:
fix https://github.com/igraph/rigraph/issues/407 
fix #556

``` r
devtools::load_all("~/git/R_packages/rigraph/")
#> ℹ Loading igraph
g1 <- make_graph(c(1, 1))
g2 <- make_graph(c(1, 1, 1, 1, 1, 1, 1, 1, 1, 1))
plot(g1)
```

``` r
plot(g2, margin = 0.1)
```

<sup>Created on 2025-06-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


szhorvat:
In short: I would merge this, as it is an improvement, but I disagree with closing #407, as it's not a complete fix.

schochastics:
Thanks. I will have a look if I can implement an angle version instead and otherwise I would merge as is

szhorvat:
> I will have a look if I can implement an angle version

The examples I showed are from Mathematica, but the angle version is implemented in Python as well, so that might be a useful reference.

I think doing this might be quite a bit of work. IMO it'd make sense to merge the improvement in this PR, as it's already much better than the current situation. Further improvements can always come later.

schochastics:
You are right that these seems complicated. But the least that I will try is to place the loops more smartly in free space. This does not look very nice 
``` r
library(igraph)
g <- make_graph(c(1,2,2,3,3,1,1,1,1,1,1,1,1,1),directed = FALSE)
V(g)$x <- c(1,1.5,2)
V(g)$y <- c(0,1,0)
plot(g,margin = 0.1)
```



``` r
#> ℹ Loading igraph
plot(g,margin = 0.1)
```



<sup>Created on 2025-06-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
Getting closer. loops are now placed in the largest gap available. Need to adjust the handling of `edge.loop.angle`. Idea: default to NA to use dynamic calculation. 
``` r
#> ℹ Loading igraph

g2 <- make_graph(c(1,2, 2,3, 3,1, 1,4, 4,5, 5,1, 1,1, 1,1, 1,1, 1,1, 1,1, 1,1, 1,1))
plot(g2)
```



``` r

g2 <- make_graph(c(1,2, 2,3, 3,1, 1,4, 4,5, 5,1, 3,4, 5,2, 1,1, 1,1, 1,1, 1,1, 1,1, 1,1, 1,1))
plot(g2)
```



<sup>Created on 2025-06-24 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
need to fix label placement

schochastics:
Currently depending only on the loop count. I still have to make it adapt to the available space too

szhorvat:
Do you think it's a good idea to keep the original version (loops within loops) as an option? I haven't seen this style before. It's not actually bad, it's just different. Once the loop is fit in the largest space (which you implemented), it could be an interesting option.

schochastics:
Now its implemented. Will check on the nitpick and rectangular nodes. 
You mean you havent seen the "loop in loop" before? I actually have never seen the "flower pedals" style before! Personally, I think we should decide on one style. I am fine with the "pedals". Keeping both makes the already convoluted plotting code even worse.

schochastics:
I dont understand where these errors come from 😞

krlmlr (commented):
Thanks, let's strive for consistency.

PR #1899, opened 2025-06-27
feat: add `plot(vertex.label.angle = , vertex.label.adj = )` arguments to rotate vertex labels

### LINKED ISSUE #106: Feature request: rotatable vertex labels
I would like to request a new feature: the ability to rotate (set the text angle of) the vertex labels that are printed. My motivation for this was creating a graph with the sugiyama layout. I have long names for my vertices, and with them arranged along horizontal lines, the labels were overlapping to the point of being unreadable. 

As a brute force proof-of-concept, I edited a local copy of `plot.igraph` and added a `srt` argument to `text` at https://github.com/igraph/rigraph/blob/dev/R/plot.R#L394. All the labels were printed at my (fixed, constant) specified angle. A proper implementation would allow the angle to be set as a vertex parameter (`vertex.label.angle`, perhaps? A consideration as to if it should be in degrees (as `srt` expects) or radians (to be consistent with `vertex.label.degree`) is needed). There are other `text` calls in the "add the labels" section of `plot.igraph` that would need to be changed, and I did not even look at whether this was possible with `tkplot` or `rglplot`

In addition to helping with long labels with the sugiyama layout, I imagine it could be helpful for grid layouts too.

schochastics:
fix #106

``` r
#> ℹ Loading igraph
g <- make_ring(5, directed = FALSE, circular = FALSE)
V(g)$label <- c("AAAAA", "BBBBB", "CCCCC","DDDDD", "EEEEE")
g$layout <- cbind(1:5, rep(1, 5))
plot(
  g,
  vertex.label.angle = c(90, 90, 270, 270, 90),
  vertex.label.adj = c(1.3,0.5)
)
```

<sup>Created on 2025-06-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


krlmlr:
Thanks!

PR #1906, opened 2025-06-30
fix: `NA` checking only in from/to columns of edge data.frame

schochastics:
previously we were also checking attributes

### DIFF
--- a/R/conversion.R
+++ b/R/conversion.R
@@ -1586,8 +1586,8 @@ graph_from_data_frame <- function(d, directed = TRUE, vertices = NULL) {
     cli::cli_abort("{.arg d} should contain at least two columns")
   }
 
-  ## Handle if some elements are 'NA'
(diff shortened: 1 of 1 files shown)


R/conversion.R
```diff
@@ -1586,8 +1586,8 @@ graph_from_data_frame <- function(d, directed = TRUE, vertices = NULL) {
     cli::cli_abort("{.arg d} should contain at least two columns")
   }
 
-  ## Handle if some elements are 'NA'
-  ensure_no_na(d, "edge data frame")
+  ## Handle if some elements are 'NA' (first two columns are interpreted as from/to)
+  ensure_no_na(d[, c(1,2)], "edge data frame")
```
krlmlr, line 1590:
```suggestion
  ensure_no_na(d[, 1:2], "edge data frame")
```

krlmlr (commented):
Thanks, great!

Issue #1956, opened 2025-07-07
x and y axes use different units in plots, resulting in distortion


szhorvat:
It seems to me that in R, plots of graphs look distorted as the x and y axes seem use different units. Example

```r
plot(make_lattice(c(2,3)))
```

<img width="332" alt="Image" src="https://github.com/user-attachments/assets/58548c77-0053-4b56-bb9b-1a96fb536811" />

Compare this in Mathematica:

<img width="441" alt="Image" src="https://github.com/user-attachments/assets/ebf02ed8-cc8a-47ad-a84b-43751413b3fa" />

Both sides of the grid should look like a square, not like a squished rectangle. I get what R shows me when forcing the bounding box to a square, at the cost of using different units along the x and y axes.

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/42508e0d-deb7-421a-aa6e-2d3c97f33c33" />

@schochastics @maelle Do you think this could be fixed? What's the standard approach in R? For graph plots, the units should really be the same, otherwise a rotated rectangle doesn't even look like a rectangle, but a parallelogram:

<img width="348" alt="Image" src="https://github.com/user-attachments/assets/e9173a2a-1f95-4cb2-bb3a-2416cf53343b" />

The original coordinates returned from the C core do form a rectangle.

szhorvat:
Yes, I think it's the same. I can't find where the code does this though. The C core doesn't do it. ChatGPT told me that `asp=1` should use the same units, but that doesn't fix this issue, and I couldn't tell if ChatGPT gave bad advice or there's some internal coordinate normalization. It seems that there is.

schochastics:
I am quite confident that I know where to fix this. I will address it this week

PR #1958, opened 2025-07-07
feat: expose `align_layout()` and add to `layout_nicely()` to align layout with axis automatically

### LINKED ISSUE #1907: Auto-align graph layouts with `layout_nicely()`
Update `layout_nicely()` to auto-align layouts using the new `igraph_layout_align()` C function.

See https://github.com/igraph/igraph/pull/2781

It'd be nice to also expose the align feature separately.

**Use cases for the feature**
Pretty plots.

**References**
—

### LINKED ISSUE #1957: `layout_nicely()` should identify forests/trees

`layout_nicely()` should detect small forests/trees using `is_forest()` and automatically use `layout_as_tree()`. I recommend doing this for graphs with at most 30 vertices only.

**Use cases for the feature**

More readable, prettier layout.

**References**

—

szhorvat:
This is for #1907

This needs more work, of course, but I very much hope that it could be part of the next release. Can you take it over @schochastics ? Even this one-liner change makes a big difference IMO.

It'd be even better if #1956 could be addressed at the same time.

CC @maelle 

----

**Examples**

```r
plot(make_ring(5))
```

image

```r
plot(make_lattice(c(2,4)))
```

image

```r
plot(make_graph(~ 1-2-3-1-4))
```

image

See some more example in https://github.com/igraph/igraph/pull/2781

----

**Further improvements I'd suggest**

 - Expose the alignment feature to users as `align_layout()` (or choose a better name, but I recommend not starting with `layout_` to avoid confusion with functions that create layouts)
 - Increase the iteration count for `layout_with_fr()` for _small graphs only_. It makes a notable improvement.
 - #1957 

Fix #1957 
Fix #1907


szhorvat:
Yes, this is expected.

It doesn't always do well with layouts that are rotationally symmetric, such as a cycle graph, a complete graph, or an $n\times n$ grid. An $n\times m, n\ne m$ grid would be fine.

What the system does is that it attempts to align, and if the order parameter is too small, it removes an edge and tries again. With a $C_4$ or a grid like this, what may happen is that the first step succeeds due to imperfections of the output from `layout_with_fr()`, and there is no need for a second try after an edge removal. If it succeeds, sometimes it aligns the corners, i.e. gives a 45 degree orientation, which is still better than something that's neither 0/90, nor 45.

A way to improve on this is to increase the iteration count in `layout_with_fr()`, even by as much as a factor of 10. This is more computationally costly, but I think it's worth doing for very small graphs, where good layout matters the most anyway.

schochastics:
@szhorvat please check

szhorvat:
Once we have this in R, it'd be easier to experiment and we can get feedback from users. Then we can experiment with the threshold that controls when to remove an edge and try again.

One more comment here is that I believe this algorithm to be practically fast (it's linear time), but alignment is somewhat pointless for very large graphs. So it's okay to be safe and skip doing it for very large ones (e.g. above 1000 vertices).

R/layout.R
```diff
@@ -954,10 +961,15 @@ layout_nicely <- function(graph, dim = 2, ...) {
...
+
+    if (is_forest(graph) && vcount(graph) <= 30) {
+      return(align_layout(graph, do.call(layout_as_tree, args)))
+    }
+
     args$dim <- dim
 
     if (vcount(graph) < 1000) {
-      do.call(layout_with_fr, args)
+      align_layout(graph, do.call(layout_with_fr, args))
```
szhorvat, line 972: Suggestion: Do not use alignment for the tree layout, which is already perfectly aligned.

Alignment makes sense for organic layouts such as `fr` and `kk`, but not for trees.

szhorvat:
Brainstorming a bit about the best interface:

How about adding an `align` parameter to all organic layouts, and setting it to `TRUE` by default?  Or even just auto-aligning all organic layouts above a certain vertex count, without an option to disable this behaviour?

This is perhaps nicer than restricting this feature to `layout_nicely()`?  It also resolves the issue with deciding what layouts to align (e.g. that trees don't need alignment).

schochastics:
My opinion: 
Leave it at `layout_nicely()` for now but expose `align_layout()` which this PR does.
Adding this tacitly to all layout functions is not a good idea because it removes the non-determinism which might confuse users. In the longrun, we probably should add an parameter (or an option?) that is either TRUE or FALSE (for backward compatibility the better choice) as default. 

But I do not fell that strongly about this

Issue #1121, opened 2024-01-16
write code to advance lifecycle deprecations


maelle:
I made  #1120 "by hand" but next time, we'll have many more messages to change.

schochastics:
@krlmlr this might be a good claude issue (cc @maelle)

schochastics:
here is a hybrid LLM solution (The regex were too advanced for me). Can you check if it does the right thing @maelle?
```r
library(stringr)

r_files <- list.files("R", pattern = "\\.R$", full.names = TRUE)

update_deprecations <- function(file, new_version) {
  levels <- c("deprecate_soft", "deprecate_warn", "deprecate_stop")
  level_map <- setNames(
    deprecation_levels[-1],
    deprecation_levels[-length(deprecation_levels)]
  ) |>
    rev()

  content <- paste(readLines(file), collapse = "\n")

  for (old_level in names(level_map)) {
    new_level <- level_map[[old_level]]

    content <- gsub(
      paste0("(?<![a-zA-Z0-9_])(?:lifecycle::)?", old_level),
      paste0("lifecycle::", new_level),
      content,
      perl = TRUE
    )

    content <- gsub(
      paste0("(lifecycle::", new_level, "\\s*\\(\\s*\")([^\"]+)(\")"),
      paste0("\\1", new_version, "\\3"),
      content,
      perl = TRUE
    )
  }

  writeLines(strsplit(content, "\n")[[1]], file)

  invisible(TRUE)
}

purrr::map(r_files,update_deprecations,new_version = "2.2.0")

```

PR #1960, opened 2025-07-07
example of deprecations advancig

maelle:
Example for #1121
@schochastics


R/adjacency.R
```diff
@@ -293,17 +293,17 @@ graph_from_adjacency_matrix <- function(
   mode <- igraph.match.arg(mode)
 
   if (!is.matrix(adjmatrix) && !inherits(adjmatrix, "Matrix")) {
-    lifecycle::deprecate_soft(
-      "1.6.0",
+    lifecycle::deprecate_warn(
```
maelle, line 296: @schochastics actually I think we might need to leave the version number untouched. The version number has to be the one at which the decision was made (that said, I've been confused by this before).

``` r
lifecycle::deprecate_soft("1.0.0", "foo()")
lifecycle::deprecate_warn("1.0.0", "foo()")
#> Warning: `foo()` was deprecated in <NA> 1.0.0.
#> ℹ The deprecated feature was likely used in the base package.
#>   Please report the issue to the authors.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
lifecycle::deprecate_stop("1.0.0", "foo()")
#> Error:
#> ! `foo()` was deprecated in <NA> 1.0.0 and is now defunct.
```

<sup>Created on 2025-07-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/basic.R
```diff
@@ -65,7 +65,7 @@ is_igraph <- function(graph) {
 #' @export
 #' @keywords internal
 get.edge <- function(graph, id) {
-  lifecycle::deprecate_stop("2.1.0", "get.edge()", "ends()")
+  lifecycle::deprecate_stop("2.2.0", "get.edge()", "ends()")
```
maelle, line 68: this is wrong.

maelle:
yes, and we could just get rid of the version changing! thank you!

PR #1962, opened 2025-07-07
fix: dont rescale coordinates to `[-1,1] x [-1,1]` by default

### LINKED ISSUE #1492: plot rescales and normalizes graph even if it has vertex x and y attributes

I expect the graph drawing to be based on the `x` and `y` attributes of the vertices. The documentation for `layout_nicely` says
> The current implementation works like this:
> 1. If the graph has a graph attribute called ‘layout’, then this is used. If this attribute is an R function, then it is called, with the graph and any other extra arguments.
> 2. Otherwise, if the graph has vertex attributes called ‘x’ and ‘y’, then these are used as coordinates. If the graph has an additional ‘z’ vertex attribute, that is also used.

When I force `xlim` and `ylim` to the range of those attributes, I get what I would expect by default, based on the documentation. There's a [post in the forum where I asked about this](https://igraph.discourse.group/t/how-do-i-stop-igraph-from-normalizing-my-layout-to-a-1-1-to-1-1-bounding-box/1865/2), and the `rescale = TRUE` suggestion was made.

I think the issue is with the [line of code where `xlim = c(-1, 1) ylim = c(-1, 1)` is imposed](https://github.com/igraph/rigraph/blob/1f15302cc28eb3acf9ae46450e7bbf3944915bce/R/plot.R#L88C25-L88C57).

### To reproduce
```r
library(igraph)
n <- 11
a <- seq(0, 2 * pi, length.out = n + 1)[-1]
x <- 5 * cos(a)
y <- 3 * sin(a)
L <- matrix(c(x, y), ncol = 2)

G <- make_full_graph(n) |> 
  set_vertex_attr("x", value = x) |>
  set_vertex_attr("y", value = y)

par(mfrow = c(1, 3))
plot(G, layout = layout_nicely(G), axes = TRUE)
plot(G, layout = L, rescale=FALSE, axes = TRUE)
plot(G, xlim = range(V(G)$x), ylim = range(V(G)$y), 
     rescale = FALSE, asp = 1, axes = TRUE)

Created on 2024-09-08 with [reprex v2.1.1](https://reprex.tidyverse.org/)
```
image
To be clear, the third panel is what I expect.

### System information

I don't think any of this is relevant, but for completeness:

```r
> version
platform       aarch64-apple-darwin20      
arch           aarch64                     
os             darwin20                    
system         aarch64, darwin20           
status                                     
major          4                           
minor          3.3                         
year           2024                        
month          02                          
day            29                          
svn rev        86002                       
language       R                           
version.string R version 4.3.3 (2024-02-29)
nickname       Angel Food Cake
```
All packages updated today, and `igraph` at version 2.0.3

### LINKED ISSUE #1956: x and y axes use different units in plots, resulting in distortion

schochastics:
Fix #1956
Fix #1492 

``` r
#> ℹ Loading igraph
n <- 11
a <- seq(0, 2 * pi, length.out = n + 1)[-1]
x <- 5 * cos(a)
y <- 3 * sin(a)
L <- matrix(c(x, y), ncol = 2)

G <- make_full_graph(n) |>
  set_vertex_attr("x", value = x) |>
  set_vertex_attr("y", value = y)

par(mfrow = c(1, 3))
plot(G, layout = layout_nicely(G), axes = TRUE)
plot(G, layout = L, rescale = FALSE, axes = TRUE)
plot(
  G,
  xlim = range(V(G)$x), ylim = range(V(G)$y),
  rescale = FALSE, asp = 1, axes = TRUE
)
```

<sup>Created on 2025-07-07 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


krlmlr:
Thanks!

PR #1975, opened 2025-07-10
docs: update allcontributors info

maelle:
How could we remember to run `allcontributors::add_contributors()`  regularly? If we used `usethis::use_release_issue()` to prepare releases we could rely on the bullets https://usethis.r-lib.org/reference/use_release_issue.html#customization

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ GNU GPL version 2 or later
 ## Contributors
 
 
+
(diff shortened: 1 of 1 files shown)


schochastics (approved):

schochastics:
Yes either we use the release issue or just try to remember (from now on) to run it 😄

Issue #731, opened 2023-03-20
Review all error messages


maelle:
- Create helper functions if needed.
- Switch to `cli::cli_abort()`.
- Make sure all messages end with a period.

Following up on  #702

krlmlr:
We have bigger problems, let's shelve for now.

maelle:
https://github.com/igraph/rigraph/issues/1693#issuecomment-2668458703

maelle:
https://simonpcouch.github.io/chores/reference/cli_helper.html

maelle:
After that we can put `stop()` in a list of unwanted functions fed to lintr.

maelle:
After those PRs, maybe we can close this issue and fix error messages as we see them.

maelle:
https://github.com/igraph/rigraph/commit/c4a2c3339d1d31d3b992a44ea28d311617eda0d1

PR #1996, opened 2025-07-22
chore: improve `igraph.match.arg()`

maelle:
Part of #731

### DIFF
--- a/R/utils-assert-args.R
+++ b/R/utils-assert-args.R
@@ -17,7 +17,12 @@ ensure_igraph <- function(graph, optional = FALSE) {
 }
 
 
-igraph.match.arg <- function(arg, values, error_call = rlang::caller_env()) {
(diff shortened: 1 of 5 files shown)


R/utils-assert-args.R
```diff
@@ -17,7 +17,12 @@ ensure_igraph <- function(graph, optional = FALSE) {
 }
 
 
-igraph.match.arg <- function(arg, values, error_call = rlang::caller_env()) {
+igraph.match.arg <- function(
+  arg,
+  values,
+  error_call = rlang::caller_env()
+) {
+  error_arg <- rlang::caller_arg(arg)
```
maelle, line 25: it's important to place this at the beginning otherwise it does not work.

schochastics (approved):
Thanks! Looks good to me :)

PR #1997, opened 2025-07-22
chore: simplify error by relying on default behavior

maelle:
Part of #731

I find the error message good as it is.

### DIFF
--- a/R/foreign.R
+++ b/R/foreign.R
@@ -521,10 +521,7 @@ write_graph <- function(
 # Plain edge list format, not sorted
 ################################################################
 
-read.graph.edgelist <- function(file, n = 0, directed = TRUE, ...) {
-  if (length(list(...)) > 0) {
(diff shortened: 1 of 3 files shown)


R/foreign.R
```diff
@@ -521,10 +521,7 @@ write_graph <- function(
 # Plain edge list format, not sorted
 ################################################################
 
-read.graph.edgelist <- function(file, n = 0, directed = TRUE, ...) {
-  if (length(list(...)) > 0) {
-    stop("Unknown arguments to read_graph (edgelist format)")
```
maelle, line 526: the name of the function is ok. However, it might make sense to rename all those functions with underscores?

schochastics (approved):
Nice, thanks! This is probably subjective opinion but I like the "." for functions that are internal only because it sets them apart from user facing functions. Maybe thats a design question we could eventually discuss. For now, I would probably leave it as is

PR #1998, opened 2025-07-22
chore: rm useless error message

maelle:
Part of https://github.com/igraph/rigraph/issues/731

To me it seems superfluous, and not usual, to check the class of an object in an S3 method.


schochastics (approved):
Agreed. Never seen this before

PR #2006, opened 2025-07-22
chore: improve errors in iterators.R

### LINKED ISSUE #1891: Setting an attribute to `NULL` fails
``` r

g <- make_ring(10)
V(g)$name <- paste0("v", 1:10)
V(g)$name <- NULL
#> Error in `V<-`(`*tmp*`, value = structure(1:10, names = c("v1", "v2", : invalid indexing
```

<sup>Created on 2025-06-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

maelle:
Fix #1891 
Part of #731


R/iterators.R
```diff
@@ -1144,14 +1144,11 @@ simple_es_index <- function(x, i, na_ok = FALSE) {
 #' @name igraph-vs-attributes
 #' @export
 `[[<-.igraph.vs` <- function(x, i, value) {
-  if (
-    !"name" %in% names(attributes(value)) ||
-      !"value" %in% names(attributes(value))
```
maelle, line 1149: I don't understand how value could be missing. I've added tests for when name is missing, but value missing would mean nothing on the right side of the arrow?! Am I missing something obvious?

In any case we need to improve the mysterious "Invalid indexing" error.

krlmlr:
Let's split this PR, the "Graph is unknown" thing is special and might become unnecessary with the new attribute handling, I hope.

maelle:
Sorry, but split into what? One PR with all error messages minus the one I commented on?

maelle:
https://github.com/igraph/rigraph/pull/2006#issuecomment-3227505694

krlmlr:
Let's find or generate extensive tests for the intended behavior of this code. The RHS for these assignments is an object with `name` and `value` attributes, I'd like to understand actual usage examples.

krlmlr:
I wonder if we should just make `<- NULL` work idiomatically -- drop attribute. We could assess the revdep impact.

maelle:
I had added the hackathon label to https://github.com/igraph/rigraph/issues/1891 :smile_cat:

krlmlr:
Can we solve #1891 as part of this PR?

R/iterators.R
```diff
@@ -1227,11 +1197,16 @@ simple_es_index <- function(x, i, na_ok = FALSE) {
 #' @name igraph-vs-attributes
 #' @export
 `[[<-.igraph.vs` <- function(x, i, value) {
+  if (!rlang::has_name(attributes(value), "attr_name")) {
```
maelle, line 1230: I asked Claude to rename `name` and `value` to `attr_name` and `attr_value` in this context because I think the double "value" added to my confusion when I first looked at this code.

schochastics (approved):

krlmlr:
Let's see what the continuous benchmarks are doing.

R/iterators.R
```diff
@@ -1227,11 +1227,19 @@ simple_es_index <- function(x, i, na_ok = FALSE) {
...
+    cli::cli_abort("Can't find {.val name} for vertex attribute.")
+  }
   if (
-    !"name" %in% names(attributes(value)) ||
-      !"value" %in% names(attributes(value))
+    !rlang::has_name(attributes(value), "attr_value") &&
+      !is_complete_iterator(value)
```
krlmlr, line 1235: Do we really need `is_complete_iterator()`, or can we use a faster and stricter check?

maelle: The definition of `is_complete_iterator()` is

```r
is_complete_iterator <- function(x) {
  identical(attr(x, "is_all"), TRUE)
}
```

maelle:
Changing to `isTRUE()` because it will look cleaner to me but I don't expect this will change the speed. Are we still worried about those benchmark results?

maelle:
The new benchmark results have no noise :grin:

tests/testthat/_snaps/iterators.md
```diff
@@ -106,5 +106,129 @@
...
+      E(g)[1:3]$weight <- NULL
+    Condition
+      Error in `[<-`:
+      ! Can't find "value" for edge attribute "weight".
+      i Removing an attribute is only supported for the whole edge sequence, e.g. `E(g)$weight <- NULL`, not a subset. Use `delete_edge_attr()`.
+
+# assigning `NULL` for a non-existent attribute errors like `delete_vertex_attr()`/`delete_edge_attr()`
+
+    Code
+      V(g)$color <- NULL
```
krlmlr, line 132: Allowing this here (but not for `delete_vertex_...()` would be more idiomatic.

tests/testthat/_snaps/iterators.md
```diff
@@ -106,5 +106,129 @@
...
+      Error in `delete_edge_attr()`:
+      ! No edge attribute `weight` found.
+
+# direct misuse of `V<-`/`E<-`/`[<-`/`[[<-` errors well
+
+    Code
+      V(g) <- "blue"
+    Condition
+      Error in `V<-`:
+      ! Can't find "name" for vertex attribute.
```
krlmlr, line 151: Improve error message here and below. Can we check if RHS is an iterator object?

tests/testthat/_snaps/iterators.md
```diff
@@ -106,5 +106,129 @@
...
+
+---
+
+    Code
+      es$weight <- 0
+    Condition
+      Error in `$<-`:
+      ! Can't find graph.
+
+# `[<-.igraph.vs` reports an internal error when the graph is unknown
```
krlmlr, line 225: Why is this an internal error?

krlmlr (commented):
Thanks! Merging now, addressing review in a follow-up PR.

PR #2008, opened 2025-07-22
docs: remove demo

### LINKED ISSUE #2004: Retire the demos?
I think it's a pretty outdated form of docs, and it would be one less thing to maintain. Thoughts @schochastics?

maelle:
Fix #2004

:fire:


schochastics (approved):

maelle:
Let's also wait for a third opinion.

krlmlr:
Is there an easy way to get a static preview of the demos, perhaps as a Quarto document? We could then decide if we want to reuse this content.

To avoid blockers, we could remove the `igraphdemo()` function first, in a separate PR.

Issue #2097, opened 2025-08-27
Notes from the changelog


maelle:
From #2058 reading https://github.com/igraph/igraph/blob/develop/CHANGELOG.md

- The R package does not use `igraph_setup()`, should it do so somehow?

maelle:
>  - `igraph_is_simple()` gained an extra `igraph_bool_t` argument that decides whether edge directions should be considered. Directed graphs with a mutual edge pair are treated as non-simple if this argument is set to `IGRAPH_UNDIRECTED` (which treats the graph as if it was undirected).

This is not documented @szhorvat? At least I see no difference between https://igraph.org/c/html/develop/igraph-Structural.html#igraph_is_simple and https://igraph.org/c/html/0.10.16/igraph-Structural.html#igraph_is_simple

maelle:
>  - The type of the `loops` argument of `igraph_adjlist_init_complementer()`, `igraph_centralization_degree()`, `igraph_centralization_degree_tmax()`, `igraph_degree()`, `igraph_maxdegree()`, `igraph_sort_vertex_ids_by_degree()` and `igraph_strength()` was changed to `igraph_loops_t` from `igraph_bool_t`, allowing finer-grained control about how loop edges are treated.

In the R package we have for instance https://github.com/igraph/rigraph/blob/90be8d944bd4d22cdf2cb9934fd6995e2f599a3d/R/centralization.R#L411-L415 (this function should actually called the one that is in aaa-auto.R, maybe we could fix that at the same time)

maelle:
>  - `igraph_distances()`, `igraph_distances_cutoff()`, `igraph_get_shortest_path()`, `igraph_get_shortest_paths()` and `igraph_get_all_shortest_paths()` gained a `weights` argument. The functions now automatically select the appropriate implementation (unweighted, Dijkstra, Bellman-Ford or Johnson) algorithm based on whether weights are present and whether there are negative weights or not. You can still call the individual methods by their more specific names.

Not documented? https://igraph.org/c/html/develop/igraph-Structural.html#igraph_distances

maelle:
>  - `igraph_lcf()` was renamed to `igraph_lcf_small()` and `igraph_lcf_vector()` was renamed to `igraph_lcf()`. Now `igraph_lcf()` takes shifts as a vector input, while `igraph_lcf_small()` accepts a shorthand notation where shifts are given as a variable number of function arguments.

[...]

PR #2098, opened 2025-09-01
chore: autogenerate `igraph_edge()` and `igraph_edges()`

maelle:
Part of the work on the "next" branch, where we can hopefully then integrate this commit, re-generate the function then handle the new argument that `igraph_edges()`  gained.


src/cpp11.dd
```diff
@@ -1,4 +1,256 @@
 # Generated by deps.mk, do not edit by hand and do not add dependencies to system headers
 cpp11.o: \
   cpp11.cpp \
+ /home/maelle/R/x86_64-pc-linux-gnu-library/4.5/cpp11/include/cpp11/altrep.hpp \
```
maelle, line 4: those changes are probably not good :sweat_smile:

szhorvat:
Again, this is low-level stuff and not a good first choice (and perhaps not a very good choice at all) for auto-generation. I strongly suggest you start with easier things, i.e. high-level network analysis functions.

I'd have to take a more careful look at whether it is a good idea to auto-generate this, so I won't make a recommendation yet beyond saying that this should not be the first thing to look at.

szhorvat:
@maelle Why don't you start with the MST functions? It was a fairly major change but it's exactly the kind of thing that is best handled with auto-generation and will allow stripping out a lot of manually written code form the R interface.

maelle:
But is autogenerating them problematic in any way?

maelle:
I don't know what MST functions are. I'm making my way through https://github.com/igraph/rigraph/issues/2097 from top to bottom. I have to start somewhere. :sweat_smile:

PR #2099, opened 2025-09-01
chore: autogenerate `igraph_neighbors()`

maelle:
also part of the work on "next"


maelle:
But in R we have this homemade function: https://github.com/igraph/rigraph/blob/2177ea868620a1bd599905da933b532eef9c328f/src/rinterface_extra.c#L5883 Is that problematic?

R/interface.R
```diff
@@ -357,8 +357,8 @@ neighbors <- function(graph, v, mode = c("out", "in", "all", "total")) {
   if (length(v) == 0) {
     stop("No vertex was specified")
   }
-  on.exit(.Call(R_igraph_finalizer))
-  res <- .Call(R_igraph_neighbors, graph, v - 1, as.numeric(mode)) + 1L
+
+  res <- neighbors_impl(graph, vid = v - 1, mode = mode) + 1L
```
szhorvat, line 361: @maelle When you move to auto-generation, you don't usually need to correct for zero-based vs one-based indexing. Check the auto-generated sources to see what you can replace here: it will often be quite a bit more than just the `.Call`.

In this case you can pass `v` instead of `v-1`.

Note that the autogen R sources already do this:

```r
  if (igraph_opt("return.vs.es")) {
    res <- create_vs(, res)
  }
```

Therefore this should be removed from _below_.

I expect the tests to pass after this.

tools/stimulus/functions-R.yaml
```diff
@@ -27,7 +27,7 @@ igraph_ecount:
     IGNORE: RR, RC
 
 igraph_neighbors:
-    IGNORE: RR, RC
+    DEPS: vid ON graph
```
szhorvat, line 30: When you need to add a DEPS line to make things work, consider submitting an upstream PR to do it directly in the C core sources. (I'm going to do this one.)

szhorvat: I hope I didn't mess anything up, here are some more DEPS added:

https://github.com/igraph/igraph/commit/bcb8fbcfa7dfaeed02a5f4472eae36513c716f29

These won't show up here until the auto-vendoring is resumed.

tools/stimulus/functions-R.yaml
```diff
@@ -1076,3 +1076,9 @@ igraph_vertex_path_from_edge_path:
 #######################################
 
 igraph_version:
+
+#######################################
+# Unclassified
+#######################################
+
+igraph_vs_adj:
```
szhorvat, line 1084: This must be removed, this function should not use auto-generation.

maelle:
Reg the vendoring does this mean I need to wait for the autocommit to main then to rebase this branch?

szhorvat:
> Reg the vendoring does this mean I need to wait for the autocommit to main then to rebase this branch?

For DEPS, yes. But that's currently stuck, Kirill can unblock it: https://github.com/igraph/rigraph/pull/2108

maelle:
@szhorvat the vendoring has worked, so I updated this PR. Does it look correct to you? Thank you!

maelle:
I'll open a new fresh PR and add tests.

maelle:
#2151

PR #2058, opened 2025-08-17
fix: Fix tests on next branch

krlmlr:
This shows the result of updating snapshot tests for the `next` branch. Also clearing the "embedding" and "games" tests.

- [ ] snapshot tests
- [ ] embedding shows too many failures, need to test independently
- [ ] games sometimes hangs on my system
- [x] Error (test-iterators.R:9:5): iterators work
- [x] Error (test-iterators.R:190:3): identical_graphs considers attributes
- [x] Error (test-make.R:455:3): graph_from_lcf() works
- [x] Error (test-paths.R:5:3): radius() works
- [x] Error (test-paths.R:12:3): radius() works -- weights
- [x] Error (test-paths.R:22:3): radius() works -- lifecycle
- [x] Error (test-paths.R:29:3): eccentricity() works
- [x] Error (test-paths.R:36:3): eccentricity() works -- weights
- [x] Error (test-paths.R:46:3): eccentricity() works -- lifecycle
- [x] Error (test-print.R:45:5): print.igraph() works
- [x] Error (test-rewire.R:2:3): rewire(each_edge(mode='in')) keeps the in-degree distribution
- [x] Error (test-rewire.R:10:3): rewire(each_edge(mode='out')) keeps the out-degree 
- [x] Error (test-rewire.R:18:3): rewire() with zero probability does not do anything
- [x] Error (test-similarity.R:3:3): similarity works
- [x] Error (test-topology.R:3:3): automorphisms works
- [x] Error (test-topology.R:12:3): automorphisms works with colored graphs
- [x] Error (test-topology.R:97:3): canonical_permutation works
- [x] Error (test-trees.R:10:3): is_tree works for non-trees
- [x] Error (test-trees.R:186:3): sample_spanning_tree works for connected graphs
- [x] Error (test-trees.R:210:3): sample_spanning_tree works for disconnected graphs
- [x] Error (test-trees.R:223:3): subgraph.edges deprecation
- [x] Error (test-structural-properties.R:68:3): max_degree() works
- [x] Error (test-structural-properties.R:379:3): all_shortest_paths() works
- [x] Error (test-structural-properties.R:416:3): shortest_paths() works
- [x] Error (test-structural-properties.R:806:3): mean_distance works correctly for 
- [x] Error (test-structural-properties.R:834:3): mean_distance can provide details
- [x] Error (test-structural-properties.R:922:3): knn works
- [x] Error (test-centrality.R:53:3): `authority_score()` works
[...]

PR #2110, opened 2025-09-02
chore: autogenerate `igraph_diameter()`

maelle:
@szhorvat I don't understand why the autogenerated version doesn't have the `weights` argument that the non autogenerated version had. Was the R version different from the C version?

I know there are changes related to this in the develop branch of the C library but I was hoping to start the autogeneration for main.

### DIFF
--- a/R/aaa-auto.R
+++ b/R/aaa-auto.R
@@ -828,6 +828,19 @@ are_adjacent_impl <- function(graph, v1, v2) {
   res
 }
 
+diameter_impl <- function(graph, directed=TRUE, unconnected=TRUE) {
+  # Argument checks
+  ensure_igraph(graph)
+  directed <- as.logical(directed)
+  unconnected <- as.logical(unconnected)
+
+  on.exit( .Call(R_igraph_finalizer) )
+  # Function call
+  res <- .Call(R_igraph_diameter, graph, directed, unconnected)
+
+  res
+}
+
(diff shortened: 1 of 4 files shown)


szhorvat:
When you do this work, look carefully at what is in `rinterface_extra.c`. First, don't assume that similar function names all do the same thing and don't remove anything until you confirmed what it does. This refers to the `igraph_incident()` PR more than this one—I see you tried to remove two functions there and I did not (yet) check what the difference is, what each of them does, and where they are used. This level of attention is necessary to move forward safely.

Second, check if the function in `rinterface_extra.c` wraps a _single_ C core function or not. Also pay attention to differences between `master` and `develop`. There are many cases where we have multiple functions in C, but a _single_ corresponding function in R. This is exactly the case here:

 - In R, `diameter()` uses _either_ `igraph_diameter()` or `igraph_diameter_dijkstra()`, depending on the presence of weights. So not everything can be auto-generated here: you must write some code that selects which function to call, and you can do this either in C or in R. Currently it seems to be done in C.
 - On the `develop` branch this is changed: now there is a single `igraph_diameter()` for both the weighted and unweighted case. This makes full auto-generation feasible, but only with `develop`.

maelle:
[`7cbf490` (#2058)](https://github.com/igraph/rigraph/pull/2058/commits/7cbf490e9bff4bc3c6fa7c217a1c042f8fa880fd)

PR #2124, opened 2025-09-08
chore: fix `full_bipartite_impl()`

maelle:
:wave: @szhorvat! @schochastics and I would need your help to fix `full_bipartite_impl()`, which might need to happen in the configuration files for Stimulus that are stored in the C library.

The `full_bipartite_impl()` function has those lines:


maelle:
Note that I think it is useless because other functions with those lines take `graph` as an input whereas this function creates a new graph from scratch. So where would names come from?

maelle:
The same problem exists for `biadjacency_impl()` (reference to a not existing `graph` object in the body):

https://github.com/igraph/rigraph/blob/0ef90cc8a60746e4de153a7be462c2e6a352a101/R/aaa-auto.R#L2579-L2581

PR #2154, opened 2025-09-22
feat: new arguments `vids_from` and `vids_to` for `similarity()`

maelle:
This makes the test-similarity.R test pass again.

Do you agree on the way I added the new parameters? We can't put them at the same place as in the C core, since that would break existing code that relies on argument position rather than argument name.

Relevant lines in the C changelog:

>  - `igraph_similarity_jaccard()` and `igraph_similarity_dice()` now take two sets of vertices to create vertex pairs of, instead of one.

PR in the C core: https://github.com/igraph/igraph/pull/2346

I haven't found the documentation for this in the C core, which could mean I didn't search properly.


maelle:
use https://igraph.org/c/html/1.0.0/igraph-Structural.html#igraph_similarity_jaccard to update

PR #2178, opened 2025-09-30
feat: new argument for `incident()`

maelle:


maelle:
the error I was getting (now gone) made me notice we were not using `incident_impl()` but maybe there's a reason.

PR #2189, opened 2025-10-03
fix: handle new names from C in `average_path_length_impl()`

maelle:
@szhorvat two automatic names changed here, I wonder whether I should rather handle this in some YAML (what I tried failed)

- the argument `unconnected` is now `unconn`
- the element of the output list `unconnected` is now `unconn_pair`

What I tried to edit was `src/vendor/cigraph/interfaces/functions.yaml`:

```yaml
igraph_average_path_length:
    PARAMS: GRAPH graph, OPTIONAL EDGE_WEIGHTS weights, PRIMARY OUT REAL res,
        OPTIONAL OUT REAL unconnected, BOOLEAN directed=True, BOOLEAN unconnected=True
    DEPS: weights ON graph
```

Doing this created two arguments called unconnected. :sweat_smile:

### DIFF
--- a/R/structural-properties.R
+++ b/R/structural-properties.R
@@ -842,8 +842,28 @@ farthest_vertices <- function(
 #' @export
 #' @rdname distances
 #' @cdocs igraph_average_path_length
-mean_distance <- average_path_length_impl
+mean_distance <- function(
+  graph,
+  weights = NULL,
+  directed = TRUE,
+  unconnected = TRUE,
+  details = FALSE
+) {
+  res <- average_path_length_impl(
+    graph,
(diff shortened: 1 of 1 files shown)


maelle:
This PR works in that it removes the related test failures, but I want to edit things the correct way. :innocent:

szhorvat:
> What I tried to edit was `src/vendor/cigraph/interfaces/functions.yaml`:

Please don't edit that. If you need overrides, do it in `tools/stimulus/functions-R.yaml`

I don't quite understand your question. You added two `unconnected` parameters in YAML, so of course it generates two.

PR #2203, opened 2025-10-09
lpa_variant

maelle:
@szhorvat in the auto-generated function, there's no enumeration of possible values for `lpa.variant`.

```r
community_label_propagation_impl <- function(graph, mode=c("all", "out", "in", "total"), weights=NULL, initial=NULL, fixed=NULL, lpa.variant) {
``` 

This leads to an error because `igraph.match.arg()` needs those values.

I am puzzled because there is a type for `LPA_VARIANT`

```yaml
LPA_VARIANT:
    DEFAULT:
        Default: c("dominance", "retention", "fast")
    INCONV: |-
        %I% <- switch(igraph.match.arg(%I%),
          "dominance"=0L, "retention"=1L, "fast"=2L)
```

in `tools/stimulus/types-RR.yaml`

And `src/vendor/cigraph/interfaces/functions.yaml` has

```yaml
igraph_community_label_propagation:
    PARAMS: |-
        GRAPH graph, OUT VECTOR_INT membership, NEIMODE mode=ALL,
        OPTIONAL EDGE_WEIGHTS weights, OPTIONAL INDEX_VECTOR initial,
        OPTIONAL VECTOR_BOOL fixed, IN LPA_VARIANT lpa_variant
    DEPS: weights ON graph
 ``` 
 
 What am I missing?

### DIFF
--- a/tools/stimulus/types-RR.yaml
+++ b/tools/stimulus/types-RR.yaml
@@ -435,7 +435,9 @@ RWSTUCK:
 
 LPA_VARIANT:
     DEFAULT:
-        Default: c("dominance", "retention", "fast")
+        DOMINANCE: c("dominance", "retention", "fast")
+        RETENTION: c("retention", "dominance", "fast")
+        FAST: c("fast", "dominance", "retention")
     INCONV: |-
         %I% <- switch(igraph.match.arg(%I%),
           "dominance"=0L, "retention"=1L, "fast"=2L)


krlmlr:
I'm backporting this change.

krlmlr:
Thanks! Actually, no need to backport because we wrap the function anyway and provide our defaults independently.

We could add a test that the defaults in `aaa-auto.R` are consistent with those in our wrappers.

PR #2209, opened 2025-10-14
chore: `mode` argument of centralization functions

maelle:
>  - `igraph_eigenvector_centrality()`, `igraph_centralization_eigenvector_centrality()` and `igraph_centralization_eigenvector_centrality_tmax()` now use a `mode` parameter with possible values `IGRAPH_OUT`, `IGRAPH_IN` or `IGRAPH_ALL` to control how edge directions are considered. Previously they used a boolean `directed` parameter.

Like #2200


maelle:
Thank you @szhorvat!

PR #2210, opened 2025-10-14
autogenerate mst

maelle:
@szhorvat in the YAML it's

```
igraph_minimum_spanning_tree:
    PARAMS: GRAPH graph, OUT EDGE_INDICES res, OPTIONAL EDGE_WEIGHTS weights, MSTALGORITHM method=AUTOMATIC
    DEPS: res ON graph, weights ON graph
```

What should we change AUTOMATIC for to get something useful? Should it happen here in the R package or in the C core?

### DIFF
--- a/tests/testthat/_snaps/minimum.spanning.tree.md
+++ b/tests/testthat/_snaps/minimum.spanning.tree.md
@@ -4,5 +4,6 @@
       mst(g, algorithm = "undefined")
     Condition
       Error in `mst()`:
-      ! Invalid `algorithm`.
+      ! `algorithm` must be one of "automatic", "unweighted", "prim", or "kruskal", not "undefined".
+      i Did you mean "unweighted"?
 


R/aaa-auto.R
```diff
@@ -5411,6 +5411,28 @@ is_complete_impl <- function(graph) {
...
+    weights <- as.numeric(weights)
+  } else {
+    weights <- NULL
+  }
+  method <- switch(igraph.match.arg(method), "automatic"=0L, "unweighted"=1L, "prim"=2L, "kruskal"=3L)
+
+  on.exit( .Call(R_igraph_finalizer) )
+  # Function call
+  res <- .Call(R_igraph_minimum_spanning_tree, graph, weights, method)
+  if (igraph_opt("return.vs.es")) {
```
maelle: @szhorvat how to change this to return a graph, not edges?

tools/stimulus/types-RR.yaml
```diff
@@ -546,3 +546,7 @@ VERTEX_WEIGHTS:
         } else {
           %I% <- NULL
         }
+
+MSTALGORITHM:
+    DEFAULT:
+        AUTOMATIC: c("automatic", "prim", "kruskal", "unweighted")
```
szhorvat: This is technically sufficient, but I strongly recommend setting up all values (`PRIM`, `KRUSKAL`, etc.), otherwise someone will get confused in the future when _just for this one type_ (`MSTALGORITHM`) only _some_ values work, while for other types all values work.

tools/stimulus/types-RR.yaml
```diff
@@ -546,3 +546,8 @@ VERTEX_WEIGHTS:
         } else {
           %I% <- NULL
         }
+
+MSTALGORITHM:
+    DEFAULT:
+        AUTOMATIC: c("automatic", "unweighted", "prim", "kruskal")
```
szhorvat: This is technically sufficient, but I strongly recommend setting up all values (PRIM, KRUSKAL, etc.), otherwise someone will get confused in the future when just for this one type (MSTALGORITHM) only some values work, while for other types all values work.

Issue #2202, opened 2025-10-09
Warning


maelle:
From #2058 when running `test-centrality` we get

```r
> G <- sample_gnm(10, sample(1:20, 1))
+     hs <- hits_scores(G)
Warning message:
In hub_and_authority_scores_impl(graph = graph, weights = weights,  :
  At vendor/cigraph/src/centrality/hub_authority.c:294 : Hub and authority scores requested for undirected graph. These are the same as eigenvector centralities.
```

How should we handle this warning? Does this mean `hits_score()` should run `is.directed()` on its input before calling another function?

@schochastics @szhorvat

krlmlr:
@maelle: Absolutely. Instead of calling `Rf_warning()`, we would call into an R function we provide that forwards to `cli::cli_warn()` .

@szhorvat: Do you remember why `R_igraph_warning_handler()` only _records_ the warning instead of emitting it? Do you see any reason why that recording shouldn't happen at the R level? In this case, we'd also change `R_igraph_warning_handler()` to record in R.

szhorvat:
The only reason not to emit the warning immediately is if `Rf_warning()` can potentially longjump instead of returning, preventing igraph from freeing allocated resources. Can R be configured to immediately abort on a warning? If not, the better way is to emit the warning immediately.

The errors are collected so that the abort could be delayed until all igraph resources have been freed.

maelle:
That reformatting would be nice for errors too, actually.

szhorvat:
It is of course quite unpleasant that there is no information about where the error is coming from at a higher level other than a C source file and line number. I'm not sure if you guys can make this better on the R side, and identify the R function (not C function) within which the warning/error was emitted. Perhaps this is what could be achieved by delaying the printing?

In the Mathematica interface I gave up on trying to make this nicer.

The C source file and line number have still proven to be very useful for debugging purposes when users report issues or ask for help, so it might be good to keep these. Note that neither of these should be expected to be stable even through bugfix releases (adding a new line to the source file shifts line numbers). But note that this information is passed to the error/warning handler in a structures way. It is this handler function that decides what to do with it and how to assemble it into a message string suitable for presentation.

PR #2214, opened 2025-10-17
test: improve centralitystress test

maelle:
Related to #2202 

Doing this change on main, in the hope that we will then rebase the #2058 PR, because I don't want to have too many moving parts in #2058.

### DIFF
--- a/tests/testthat/test-centrality.R
+++ b/tests/testthat/test-centrality.R
@@ -146,27 +146,28 @@ test_that("authority_score survives stress test", {
 
   withr::local_seed(42)
 
-  is.principal <- function(M, lambda) {
-    all.equal(eigen(M)$values[1], lambda)
+  expect_principal <- function(M, lambda) {
+    expect_equal(eigen(M)$values[1], lambda)
   }
 
-  is.ev <- function(M, v, lambda) {
-    all.equal(as.vector(M %*% v), lambda * v)
+  expect_ev <- function(M, v, lambda) {
+    expect_equal(as.vector(M %*% v), lambda * v)
   }
 
-  is.good <- function(M, v, lambda) {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-centrality.R
```diff
@@ -147,30 +147,26 @@ test_that("authority_score survives stress test", {
...
-    expect_equal(eigen(M)$values[1], lambda)
+    all.equal(eigen(M)$values[1], lambda)
   }
 
   is.ev <- function(M, v, lambda) {
-    expect_equal(as.vector(M %*% v), lambda * v)
+    all.equal(as.vector(M %*% v), lambda * v)
   }
 
   is.good <- function(M, v, lambda) {
```
krlmlr: The original implementation is better if renamed to `expect_principal()` and `expect_ev()` . We could pass the loop index `i` as a label or something, for clearer messages.

tests/testthat/test-centrality.R
```diff
@@ -147,30 +147,26 @@ test_that("authority_score survives stress test", {
...
-    M <- as_adjacency_matrix(G, sparse = FALSE)
-    is.good(t(M) %*% M, as$authority, as$value)
-  }
-
-  for (i in 1:100) {
-    G <- sample_gnm(10, sample(1:20, 1))
+    G <- sample_gnm(10, sample(15:25, 1), directed = FALSE)
+    while (!is_connected(G)) {
+      G <- sample_gnm(10, sample(15:25, 1), directed = FALSE)
+    }
```
krlmlr: A `repeat {}` with a `break` avoids duplication?

krlmlr (commented):
Thanks for looking into it!

PR #2421, opened 2025-10-26
Changes between main and next for review

krlmlr:
These are the essential difference between the `main` and `next` branches. This is not intended to be merged, just as a review tool.

Do any of these differences look suspicious? Let's discuss here and resolve.

Generated with:

```sh
git diff next -- ':(exclude)man' ':(exclude)src/vendor' ':(exclude)patch' ':(exclude)src/Makevars.*' ':(exclude)src/rinterface.*' ':(exclude)src/cpp11.cpp' ':(exclude)NEWS.md' ':(exclude)R/aaa-auto.R' ':(exclude)cran-comments.md' ':(exclude)tools/stimulus/' ':(exclude)src/sources-igraph.mk' | patch -p1 -R
```

@szhorvat @schochastics: We do benefit from taking a closer look here, but we will also catch errors later. I'll review too.


R/centrality.R
```diff
@@ -1390,10 +1404,33 @@ eigen_centrality <- function(
     }
   }
 
+  mode <- igraph.match.arg(mode)
+
+  if (lifecycle::is_present(directed)) {
+    if (directed) {
+      lifecycle::deprecate_soft(
+        "2.2.0",
```
krlmlr: Here and everywhere:

```suggestion
        "3.0.0",
```

R/centralization.R
```diff
@@ -397,8 +402,8 @@ centr_degree <- function(
 #' @examples
 #' # A BA graph is quite centralized
 #' g <- sample_pa(1000, m = 4)
-#' centr_degree(g, normalized = FALSE)$centralization %>%
-#'   `/`(centr_degree_tmax(g, loops = FALSE))
+#' centr_degree(g, normalized = FALSE)$centralization /
+#'   centr_degree_tmax(g, loops = "twice")
```
krlmlr: I remember I wasn't sure about this. For `centr_degree()`, the default is `"twice"` IIRC. Is then the same `loops` value needed in the denominator?

R/rewire.R
```diff
@@ -41,7 +41,7 @@
 rewire <- function(graph, with) {
   if (!is(with, "igraph_rewiring_method")) {
     cli::cli_abort(
-      "{.arg with} must be an igraph rewiring method, 
+      "{.arg with} must be an igraph rewiring method,
```
krlmlr: Should be in `main`.

R/structural-properties.R
```diff
@@ -922,8 +950,36 @@ degree <- function(
 #' @rdname degree
 #' @export
 #' @cdocs igraph_maxdegree
-max_degree <- maxdegree_impl
```
krlmlr: Could ask Copilot, could manually backport to `main` .

Two more functions that were forgotten: `mean_distance()` and `sample_spanning_tree()` .

krlmlr (commented):
Looks solid to me, by and large. Appreciate a second pair of eyes.

R/centralization.R
szhorvat: Yes, this is correct.

PR #2514, opened 2026-01-06
Reformat error+warning messages from C

maelle:
Just a proof of concept, courtesy of Claude.

I don't like that with this approach, in the error message we can't easily find the function name. Would we really need to walk the stack?!


maelle:
@krlmlr this was superseded by https://github.com/igraph/rigraph/commit/c4a2c3339d1d31d3b992a44ea28d311617eda0d1

PR #2515, opened 2026-01-06
feat: improve formatting of error messages from C

### LINKED ISSUE #2202: Warning

maelle:
Fix #2202

One thing we might consider adding, is a period at the end of the message if it's not already there.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -51,7 +51,11 @@ expect_snapshot_igraph_error <- function(x, ...) {
     transform = function(y) {
       # Scrub file name and line number from error/warning messages
       # Handles "Source: filename:linenumber" and "At path/to/file:line :" patterns
-      y <- gsub("Source: [^:]+:(\\d+|xx|<linenumber>)", "Source: <file>:<line>", y)
+      y <- gsub(
+        "Source: [^:]+:(\\d+|xx|<linenumber>)",
+        "Source: <file>:<line>",
+        y
+      )
       y <- gsub("At [^:]+:(\\d+|xx) :", "At <file>:<line> :", y)
       y
     },


src/rinterface_extra.c
```diff
@@ -2315,6 +2315,17 @@ static inline const char* maybe_add_punctuation(const char* msg, const char* pun
...
+/* Strip vendor/cigraph/src/ prefix from file path */
+static inline const char* simplify_file_path(const char* file) {
+  const char* prefix = "vendor/cigraph/src/";
+  size_t prefix_len = strlen(prefix);
+
+  if (strncmp(file, prefix, prefix_len) == 0) {
+    return file + prefix_len;
+  }
+  return file;
+}
```
szhorvat:
```suggestion
/* Strip vendor/cigraph/src/ prefix from file path. This prefix depends on the
 * build procedure, namely on the directory that the compiler is invoked from. */
static inline const char* simplify_file_path(const char *file) {
  const char prefix[] = "vendor/cigraph/src/";
  const size_t prefix_len = sizeof(prefix) / sizeof(prefix[0]);

  if (strncmp(file, prefix, prefix_len) == 0) {
    return file + prefix_len;
  }
  return file;
}
```

You can get the length statically. It's good to have a comment about where this prefix comes from because it will change if changes are made to the build procedure. The reason it even appears (compared to the standard use of the C core) is the tweaked build procedure.

szhorvat:
> One thing we might consider adding, is a period at the end of the message if it's not already there.

Isn't this already there with `maybe_add_puctuation`?

LGTM, but I didn't test.

maelle:
Thank you @szhorvat @schochastics!

src/rinterface_extra.c
```diff
@@ -2315,6 +2315,18 @@ static inline const char* maybe_add_punctuation(const char* msg, const char* pun
   return is_punctuated(msg) ? "" : punctuation;
 }
 
+/* Strip vendor/cigraph/src/ prefix from file path. This prefix depends on the
+ * build procedure, namely on the directory that the compiler is invoked from. */
+static inline const char* simplify_file_path(const char *file) {
+  const char prefix[] = "vendor/cigraph/src/";
+  const size_t prefix_len = sizeof(prefix) / sizeof(prefix[0]);
```
szhorvat:
```suggestion
  const size_t prefix_len = sizeof(prefix) / sizeof(prefix[0]) - 1;
```

PR #2598, opened 2026-03-27
fix: stricter bipartite type checking

### LINKED ISSUE #710: 'type' vertex attributes that are not logical should be handled better
First read #709 to familiarize yourself with `is_bipartite()`.

`is_bipartite()` simply checks whether there is a `type` vertex attribute:

```
function (graph) 
{
    if (!is_igraph(graph)) {
        stop("Not a graph object")
    }
    "type" %in% vertex_attr_names(graph)
}
```

The intention is that this attribute should be logical (i.e. of boolean type), however, that is not checked. It is concceivable that people will set this attribute to a different type, not being aware of its special meaning in igraph. It happened [here](https://stackoverflow.com/q/75749328/695132).

@krlmlr Would is make sense to add an `is.logical()` test as well, or are non-logical types still usable in boolean contexts in R (like they are in Python)?

----

Now let's look at other functions that make use of the `type` vertex attribute, and see how they behave with a non-logical one.

```
> g <- make_(ring(5), with_vertex_(type=letters[1:5]))
> V(g)$type
[1] "a" "b" "c" "d" "e"
```

```
> is_bipartite(g)
[1] TRUE
```

As said above, the usefulness of this result is dubious.

```
> bipartite_projection(g)
Error in handle_vertex_type_arg(types, graph) : 
  `NA' is not allowed in vertex types
In addition: Warning message:
In handle_vertex_type_arg(types, graph) : vertex types converted to logical
```

This error message is not entirely unclear, but it might be difficult for beginners to figure out what went wrong.

```
> layout_as_bipartite(g)
Error in handle_vertex_type_arg(types, graph) : 
  `NA' is not allowed in vertex types
In addition: Warning message:
In handle_vertex_type_arg(types, graph) : vertex types converted to logical
```

Same as above.

```
> as_incidence_matrix(g)
Error in handle_vertex_type_arg(types, graph) : 
  `NA' is not allowed in vertex types
In addition: Warning message:
In handle_vertex_type_arg(types, graph) : vertex types converted to logical
```
Same again.

```
> max_bipartite_match(g)
Error in handle_vertex_type_arg(types, graph) : 
  `NA' is not allowed in vertex types
In addition: Warning message:
In handle_vertex_type_arg(types, graph) : vertex types converted to logical
```

These are all the functions that I found.

schochastics:
fix #710 (cc @szhorvat)


R/other.R
```diff
@@ -159,11 +159,21 @@ handle_vertex_type_arg <- function(types, graph, required = T) {
...
     if (!is.logical(types)) {
-      cli::cli_warn("vertex types converted to logical.")
-    }
-    types <- as.logical(types)
-    if (any(is.na(types))) {
-      cli::cli_abort("`NA' is not allowed in vertex types")
+      converted <- suppressWarnings(as.logical(types))
+      if (anyNA(converted)) {
+        cli::cli_abort(
+          "The {.arg type} vertex attribute is not logical and could not be \\
```
krlmlr: The backslash does weird things here.

krlmlr: Add a snapshot test?

Issue #1485, opened 2024-09-05
increase all deprecation levels before the release?


maelle:

maelle:
Related #1121

szhorvat:
> 🎄

Before Christmas? 😁

maelle:
Yes but I'm trying to argue in favor of St Nicholas Day https://en.wikipedia.org/wiki/Saint_Nicholas_Day :joy: (especially as there's a CRAN break over the holidays)

maelle:
Oh no St Nicholas Day came and went.

schochastics:
Good idea! After 2.2.0 we just break everything 😆 
(Changed milestone)

maelle:
And I've now added the label! 🤝

maelle:
We are just following @krlmlr's good advice. 😆

maelle:
- [ ] Increase the deprecation levels.
- [ ] Make sure this is included in the changelog.
- [ ] Write a blog post.

maelle:
Script Claude and I developed:

```r
library(treesitter)
library(treesitter.r)

# --- Reset to baseline (undo any previous partial run) ---
# .claude/settings.local.json is untracked, so git reset --hard leaves it alone.
# This script is tracked, so we save and restore it across the reset.
local({
  baseline <- "fb68fc60b2e673a2994885ed454dbbaca709096c"
  script_path <- here::here("tools", "bump-lifecycle.R")
  saved <- readLines(script_path, warn = FALSE)
  cli::cli_alert_info("Resetting to {substr(baseline, 1L, 7L)} ...")
  system2("git", c("-C", here::here(), "reset", "--hard", baseline))
  writeLines(saved, script_path)
  cli::cli_alert_success("Restored tools/bump-lifecycle.R")
})

lang <- treesitter.r::language()

# deprecate_soft / deprecate_warn / deprecate_stop are all 14 chars — same length,
# so substr<-() replacement never shifts line/column positions.
next_level <- c(
  deprecate_soft = "deprecate_warn",
  deprecate_warn = "deprecate_stop"
)

# Uncomment to inspect node types for a sample expression:
# text <- 'lifecycle::deprecate_soft("1.0.0", "foo()")'
# node_show_s_expression(tree_root_node(parser_parse(parser(lang), text)))

# --- tree navigation helpers ---

fn_to_call <- function(fn_node) {
  # "deprecate_X" identifier -> namespace_operator -> call
  node_parent(node_parent(fn_node))
}

[...]

PR #2634, opened 2026-05-12
chore: Advance lifecycle stage for many functions

maelle:
Part of #1485


maelle:
Best would be to do an interactive rebase and edit the commits that introduced the test failures so that each commit would be green.

maelle:
and some bug with the iterators nei/inc etc.

maelle:
Rather than an interactive rebase, once the changes are good, based on them create 1 commit per script + test file with a `document()` run at each commit.

krlmlr:
Running revdepchecks now.

krlmlr:
Thanks, wonderful!

Issue #2652, opened 2026-05-26
jarl


maelle:
Posting this here, I'll then go rule by rule.

```
  125 [ ] true_false_symbol
   77 [*] numeric_leading_zero
   28 [*] outer_negation
   21 [*] any_is_na
   19 [*] seq
   16 [ ] implicit_assignment
   16 [*] lengths
    9 [ ] internal_function
    8 [*] class_equals
    6 [*] any_duplicated
    5 [*] string_boundary
    4 [ ] duplicated_arguments
    3 [*] matrix_apply
    2 [ ] unreachable_code
    1 [ ] seq2
    1 [ ] vector_logic
```

schochastics:
is everything jarled?

PR #2664, opened 2026-05-26
chore: implicit assignments

maelle:
Part of #2652

### DIFF
--- a/R/hrg.R
+++ b/R/hrg.R
@@ -569,9 +569,12 @@ as.dendrogram.igraphHRG <- function(object, hang = 0.01, ...) {
 
   for (k in 1:nMerge) {
     x <- merges[k, 1:2]
-    if (any(neg <- x >= 0)) {
+    neg <- (x >= 0)
+
+    if (any(neg)) {
       h0 <- if (hang < 0) 0 else max(0, oHgt[k] - hang * hMax)
     }
+
     if (all(neg)) {
       # two leaves
       zk <- as.list(x + 1)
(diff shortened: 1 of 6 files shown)


R/hrg.R
```diff
@@ -569,9 +569,12 @@ as.dendrogram.igraphHRG <- function(object, hang = 0.01, ...) {
 
   for (k in 1:nMerge) {
     x <- merges[k, 1:2]
-    if (any(neg <- x >= 0)) {
+    neg <- (x >= 0)
```
maelle, line 572: !!!!!

R/par.R
```diff
@@ -224,13 +224,18 @@ igraph_i_options <- function(..., .in = parent.frame()) {
   temp <- list(...)
   if (length(temp) == 1 && is.null(names(temp))) {
     arg <- temp[[1]]
-    switch(
-      mode(arg),
-      list = temp <- arg,
-      character = return(.igraph.pars[arg]),
+
```
maelle, line 227: maybe we can add an exception comment instead but I have to say I had trouble understanding the `switch()` call.

schochastics: IMHO switch only makes sense if it is more than 2 option OR if we expect that the list of options might grow in the future. 

Independently, I also find this statement confusing

R/hrg.R
maelle: @krlmlr @schochastics small hackathon topic?

schochastics (approved):

R/hrg.R
krlmlr: Naming is hard, but Claude knows how. Is there more substance to discuss?

krlmlr (approved):

R/hrg.R
maelle: well not having this named neg would be better for sure

PR #2666, opened 2026-05-26
chore: outer negation

maelle:
Part of #2652

### DIFF
--- a/tests/testthat/test-attributes.R
+++ b/tests/testthat/test-attributes.R
@@ -331,7 +331,7 @@ test_that("without_loops", {
   )
 
   expect_identical_graphs(g1, g2)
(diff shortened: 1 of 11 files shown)


tests/testthat/test-attributes.R
```diff
@@ -331,7 +331,7 @@ test_that("without_loops", {
   )
 
   expect_identical_graphs(g1, g2)
-  expect_true(all(!which_loop(g2)))
+  expect_all_false(which_loop(g2))
```
maelle, line 334: `which_loops()` returns a vector a logical so I think this is what we meant?

schochastics (approved):
I was always bad at bollean algebra but this seems correct to me 😆

Issue #30, opened 2015-01-14
R: is.temporal()


gaborcsardi:
_From @gaborcsardi on July 22, 2014 18:33_

_Copied from original issue: igraph/igraph#656_

PR #2677, opened 2026-05-28
feat: retire `attr` in favor of `weights` for adjacency matrix functions

### LINKED ISSUE #906: `as_adjacency_matrix()` should use `weights` instead of `attr` parameter

### LINKED ISSUE #1137: Implement sparse in as_adjacency_matrix

schochastics:
## Summary

Resolves milestone [#30 — retire `attr` for `weight`](https://github.com/igraph/rigraph/milestone/30) in one PR. Bundles the API change with the dependent centrality function migrations so no temporary translation shims are needed.

Closes: #906, #903, #910 — partially closes #1137 (the sparse path already worked; this PR audits and adds parity tests).

### 1. `as_adjacency_matrix()` / `as_adj()` ([#906](https://github.com/igraph/rigraph/issues/906))

New `weights` argument follows the standard rigraph weight convention:

| `weights =`          | Behavior |
|----------------------|----------|
| `NULL` (default)     | Use the `weight` edge attribute if present, else unweighted |
| `NA`                 | Explicitly unweighted (ignores any `weight` attribute) |
| `<numeric vector>`   | Use directly as edge weights (length must equal `ecount()`) |
| `<character scalar>` | Edge attribute name (must be numeric/logical) |

The legacy `attr` argument is now `lifecycle::deprecate_warn()`-ed against `igraph 3.0.0`. When `attr` is supplied it is forwarded into `weights`, so all existing user code keeps working with a soft warning.

### 2. `as_biadjacency_matrix()`

Same migration mirrored for the bipartite sibling. The C binding `R_igraph_get_biadjacency` doesn't have a weights slot, so the four-mode resolution happens entirely in R via the helper functions.

### 3. `alpha_centrality()` ([#910](https://github.com/igraph/rigraph/issues/910))

The previous implementation laundered a numeric `weights` vector through a `set_edge_attr(graph, "weight", value = ...)` mutation just to use the old `attr =` API. That workaround is gone — both the dense and sparse helpers now pass `weights` straight to `as_adjacency_matrix()`. This also structurally fixes the bug class behind [#915](https://github.com/igraph/rigraph/issues/915) (custom weight attribute names other than `"weight"`).

### 4. `power_centrality()` ([#903](https://github.com/igraph/rigraph/issues/903))

Gained a `weights` argument. Previously it silently dropped edge weights, returning identical results for a weighted graph and its unweighted version. Weights now thread through both `bonpow.dense()` and `bonpow.sparse()`. Drive-by: the sparse path used `degree(graph, mode = "out")` where the dense path used `rowSums(d)`; these agree only for unweighted graphs. The sparse path now uses `Matrix::rowSums(d)` for consistency.

### 5. Internal call site

`R/indexing.R:269` (the `[` operator on graphs) now passes `weights = if (is.null(attr)) NA else attr` instead of `attr = attr` — preserving its existing "unweighted by default" semantics rather than tripping the new auto-pickup.

### 6. Tests

- Migrated existing legacy `attr =` test calls to `weights =` (test-conversion.R, test-incidence.R, test-structural-properties.R, plus the matching snapshot).
- Added focused coverage for all four `weights` modes, length/type errors, dense/sparse parity for arbitrary numeric vectors, and the new deprecation warnings (with snapshots).
- Added `power_centrality()` reprex from #903 (weighted graph equivalent to unweighted graph with parallel edges).
- Added regression test for #915 (custom weight attribute name).

## Soft-breaking change

`as_adjacency_matrix(g)` on a graph that already has a `weight` edge attribute now returns a **weighted** matrix by default. Old behavior: pass `weights = NA`. This brings the function into line with the rest of rigraph (`shortest_paths()`, `cluster_leiden()`, etc.).

## Out of scope (followups)

- Deprecated wrappers `get.adjacency()`, `get.incidence()`, `as_incidence_matrix()` are intentionally **not** touched. They forward `attr = attr` to the migrated functions, so calling them now emits the outer "function is deprecated" warning plus the new `attr =` warning. That doubling is acceptable; users are supposed to be migrating off these anyway.
- A future PR will flip `lifecycle::deprecate_warn()` → `lifecycle::deprecate_stop()` for `attr =` after one release cycle on CRAN.

## Test plan

- [x] `devtools::test()` — 0 failures across the full suite (`NOT_CRAN=true`).
- [x] Targeted REPL spot-check:
  ```r
  g <- make_ring(5); E(g)$weight <- 1:5
  as_adjacency_matrix(g)                    # weighted (new default)
  as_adjacency_matrix(g, weights = NA)      # unweighted
  as_adjacency_matrix(g, weights = 10:14)   # explicit vector
  as_adjacency_matrix(g, weights = "weight")# attribute name
  as_adjacency_matrix(g, attr = "weight")   # deprecated; warns; weighted
  ```


schochastics:
To me this looks good. @krlmlr do we run revdeps on each potentially breaking PR? If so, this definitely needs it 😬

R/conversion.R
```diff
@@ -219,10 +219,81 @@ get.adjedgelist <- function(
...
+# Returns: numeric() (empty) for unweighted, or a numeric vector of length
+# ecount(graph) otherwise.
+resolve_edge_weights <- function(graph, weights, call = rlang::caller_env()) {
+  if (is.null(weights)) {
+    if ("weight" %in% edge_attr_names(graph)) {
+      value <- edge_attr(graph, "weight")
+      if (!is.numeric(value) && !is.logical(value)) {
+        cli::cli_abort(
+          c(
+            "Matrices must be either numeric or logical, and the edge attribute is not",
```
krlmlr: I don't understand what's wrong here from the wording. Perhaps `the {.val weight} edge attribute` ?

R/conversion.R
```diff
@@ -219,10 +219,81 @@ get.adjedgelist <- function(
...
+        call = call
+      )
+    }
+    if (!weights %in% edge_attr_names(graph)) {
+      cli::cli_abort("No such edge attribute", call = call)
+    }
+    value <- edge_attr(graph, weights)
+    if (!is.numeric(value) && !is.logical(value)) {
+      cli::cli_abort(
+        "Matrices must be either numeric or logical, and the edge attribute is not",
```
krlmlr: Same as above. Can we unify the check?

krlmlr (commented):
I'm not sure I like it. Could we keep two arguments, `attr` and `weight`, and resolve them in unison?

schochastics:
> I'm not sure I like it. Could we keep two arguments, `attr` and `weight`, and resolve them in unison?

You mean no deprecation warning for attr and silently forward attr to weight?

R/conversion.R
```diff
@@ -1091,26 +1143,38 @@ get.incidence.sparse <- function(
 as_biadjacency_matrix <- function(
   graph,
   types = NULL,
-  attr = NULL,
+  weights = NULL,
+  attr = deprecated(),
   names = TRUE,
   sparse = FALSE
```
krlmlr: How about adding an ellipsis after `types`, while we're at it? Adding arguments mid-signature calls for trouble.

R/conversion.R
```diff
@@ -1091,26 +1143,38 @@ get.incidence.sparse <- function(
...
 ) {
   # Argument checks
   ensure_igraph(graph)
 
+  if (lifecycle::is_present(attr)) {
+    lifecycle::deprecate_warn(
```
krlmlr: Why do we start with `deprecate_warn()` right away?

R/conversion.R
```diff
@@ -1091,26 +1143,38 @@ get.incidence.sparse <- function(
...
+      "3.0.0",
+      "as_biadjacency_matrix(attr = )",
+      "as_biadjacency_matrix(weights = )"
+    )
+    weights <- attr
```
krlmlr: I guess this is what confuses me.

R/conversion.R
```diff
@@ -1091,26 +1143,38 @@ get.incidence.sparse <- function(
...
+  }
+
   names <- as.logical(names)
   sparse <- as.logical(sparse)
 
+  weights <- resolve_edge_weights(graph, weights)
```
krlmlr: Perhaps, with setting `attr` to `NULL` if not present?

```suggestion
  weights <- resolve_edge_weights(graph, weights, attr)
```

R/conversion.R
```diff
@@ -1061,12 +1103,22 @@ get.incidence.sparse <- function(
...
+#' @param weights One of the following:
+#'   \itemize{
+#'     \item `NULL` (default): use the `weight` edge attribute if the graph has
+#'       one, otherwise return a traditional (unweighted) bipartite adjacency
+#'       matrix.
+#'     \item `NA`: explicitly unweighted, ignoring any `weight` edge attribute.
+#'     \item A numeric or logical vector of length [ecount()]: use these values
+#'       directly as edge weights.
+#'     \item A character scalar: the name of an edge attribute whose values are
+#'       used as weights. The attribute must be numeric or logical.
```
krlmlr: Is this really a convention we want to adopt/keep throughout the package? Keeping the two arguments would have been my intuitive choice, but I'm not sure about existing conventions.

krlmlr (commented):
Took another look. Let's discuss.

R/conversion.R
schochastics: yes thats to harsh

R/conversion.R
schochastics: not sure exactly what you mean but the `weights = NULL/NA/numeric/char` is used at various places already (`cluster_leiden`, `shortest_paths`,...)

R/conversion.R
schochastics: I add the ellipses but I think this can also lead some (soft) breaking changes

R/centrality.R
```diff
@@ -1884,6 +1886,11 @@ bonpow.sparse <- function(
 #' @param sparse Logical scalar, whether to use sparse matrices for the
 #'   calculation. The \sQuote{Matrix} package is required for sparse matrix
 #'   support
+#' @param weights Edge weights to use, following the standard rigraph weight
```
krlmlr: Is this a case for `@inheritParams` ?

R/conversion.R
```diff
@@ -365,30 +432,41 @@ get.adjacency.sparse <- function(
...
-  attr = NULL,
+  ...,
+  weights = NULL,
+  attr = deprecated(),
   edges = deprecated(),
   names = TRUE,
   sparse = igraph_opt("sparsematrices")
 ) {
   ensure_igraph(graph)
+  rlang::check_dots_empty()
```
krlmlr: It is possible to implement this in a non-breaking way. I did this for dm 1.0.0, the idea is to detect non-empty dots and create a function that matches the args according to the old signature. We can do this depending on revdepcheck results (maybe faster now), or adopt this practice as a rule (allows us to easily add dots without breaking downstream other than soft-deprecation warnings).

krlmlr (commented):
Thanks, great! I skimmed the patch, let's accumulate breaking PRs (by merging into main) and run revdepchecks in bulk.

krlmlr (approved):

R/conversion.R
schochastics: I like your suggestion because I had some headaches over this. Let me try and come up with something

schochastics: hmm @krlmlr Claude tells me that what we are doing here is what dm 1.0.0 does

krlmlr: We had a substantial interface change for `dm_filter()` , I think this is the most relevant one. We don't have to do it exactly like with that function, but a design along these lines might be useful.

schochastics: I used dm_filter as inspiration. There is now a function that handles this similarly here

R/conversion.R
```diff
@@ -365,30 +484,57 @@ get.adjacency.sparse <- function(
...
+  weights = NULL,
+  attr = deprecated(),
   edges = deprecated(),
   names = TRUE,
   sparse = igraph_opt("sparsematrices")
 ) {
   ensure_igraph(graph)
 
+  user_env <- rlang::caller_env()
+  recovered_attr <- recover_positional_attr(
```
krlmlr: How does this code handle `as_adjacency_matrix(graph, type, attr, names = letters)` ?

R/conversion.R
```diff
@@ -1091,26 +1207,50 @@ get.incidence.sparse <- function(
...
+  weights = NULL,
+  attr = deprecated(),
   names = TRUE,
   sparse = FALSE
 ) {
   # Argument checks
   ensure_igraph(graph)
 
+  user_env <- rlang::caller_env()
+  recovered_attr <- recover_positional_attr(
```
krlmlr: Here it's even more obvious: can/should we still support `as_biadjacency_matrix(graph, types, attr, names, sparse)` (all arguments passed unnamed) with a lifecycle warning?

krlmlr (commented):
Thanks. I had hoped we could recover other unnamed arguments as well. I wonder if we can create a very generic helper function that accepts the old signature and returns a list of values from a call with the new signature that we can then deconstruct again. That function would also issue lifecycle warnings. With such a function, we could add dots everywhere in one swoop: once the helper exists, Claude can embed it everywhere.

```r
as_biadjacency_matrix <- function(graph, types, ..., weights, attrs, names, sparse) {
  args <- handle_args(
    graph, types, ..., weights, attrs, names, sparse,
    .old_signature = list(graph, types, attrs, names, sparse)
  )
  graph <- args$graph
  types <- args$types
  ...
}
```

We could also create a dedicated helper for each such function (deterministic code generation) to simplify debugging and enable custom changes in the helper code.

```r
as_biadjacency_matrix <- function(graph, types, ..., weights, attrs, names, sparse) {
  args <- handle_args_as_biadjacency_matrix(
    graph, types, ..., weights, attrs, names, sparse
  )
  graph <- args$graph
  types <- args$types
  ...
}
```

If we do this, we can do a separate PR before this. Well in scope for 3.0.0 if we agree on that; an alternative is a hard break and dealing with the revdepchecks.

R/centrality.R
```diff
@@ -1936,13 +1939,30 @@ power_centrality <- function(
...
   rescale = FALSE,
   tol = 1e-7,
-  sparse = TRUE
+  sparse = TRUE,
+  weights = NULL
 ) {
   nodes <- as_igraph_vs(graph, nodes)
   if (sparse) {
-    res <- bonpow.sparse(graph, nodes, loops, exponent, rescale, tol)
+    res <- bonpow.sparse(
```
maelle, line 1947: since `bonpow.sparse()` is not exported, could we give it a better name? Or do we want to keep dot names for unexported functions?

R/centrality.R
```diff
@@ -1966,25 +1986,7 @@ alpha.centrality.dense <- function(
   exo <- rep(exo, length.out = vcount(graph))
```
maelle, line 1986: What does exo mean? In my French-speaking head it's an abbreviation of exercise so I'm confused.

R/centrality.R
```diff
@@ -1966,25 +1986,7 @@ alpha.centrality.dense <- function(
...
-    ## weights != NULL and weights != rep(NA, x)
-    graph <- set_edge_attr(graph, "weight", value = as.numeric(weights))
-    attr <- "weight"
-  } else {
-    ## weights != NULL, but weights == rep(NA, x)
-    attr <- NULL
-  }
-
-  d <- t(as_adjacency_matrix(graph, attr = attr, sparse = FALSE))
+  d <- t(as_adjacency_matrix(graph, weights = weights, sparse = FALSE))
```
maelle, line 1989: better name than d?

R/centrality.R
```diff
@@ -2013,25 +2015,7 @@ alpha.centrality.sparse <- function(
...
-    ## weights != NULL and weights != rep(NA, x)
-    graph <- set_edge_attr(graph, "weight", value = as.numeric(weights))
-    attr <- "weight"
-  } else {
-    ## weights != NULL, but weights == rep(NA, x)
-    attr <- NULL
-  }
-
-  M <- Matrix::t(as_adjacency_matrix(graph, attr = attr, sparse = TRUE))
+  M <- Matrix::t(as_adjacency_matrix(graph, weights = weights, sparse = TRUE))
```
maelle, line 2018: or are these single letters sort of convention?

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
 #
 ###################################################################
 
+# Resolve the user-facing `weights` argument into a numeric vector
```
maelle, line 222: is this now valid in all of igraph?

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
...
+#   weights = <char>  -> length-1 edge attribute name
+#   weights = <num>   -> numeric/logical vector of length ecount(graph)
+#
+# Returns: numeric() (empty) for unweighted, or a numeric vector of length
+# ecount(graph) otherwise.
+resolve_edge_weights <- function(
+  graph,
+  weights,
+  attr = deprecated(),
+  fn = "function",
```
maelle, line 237: why a default value?

R/conversion.R
maelle: @schochastics @krlmlr  side-note, this would be a very nice short blog post. Adding the empty dot checks in a non breaking way.

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
...
+      user_env = user_env
+    )
+    weights <- attr
+  }
+
+  if (is.null(weights)) {
+    if ("weight" %in% edge_attr_names(graph)) {
+      return(edge_attr_as_weights(graph, "weight", call))
+    }
+    return(numeric())
```
maelle, line 255: for some reason I'd find a else better?

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
...
+  }
+
+  if (is.character(weights)) {
+    if (length(weights) != 1) {
+      cli::cli_abort(
+        "{.arg weights} as character must be a single edge attribute name.",
+        call = call
+      )
+    }
+    if (!weights %in% edge_attr_names(graph)) {
```
maelle, line 269: I can't wait for `%notin%` to be old enough for us to use it.

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
...
+      cli::cli_abort("No such edge attribute", call = call)
+    }
+    return(edge_attr_as_weights(graph, weights, call))
+  }
+
+  if (!is.numeric(weights) && !is.logical(weights)) {
```
maelle, line 275: couldn't we check the types earlier?

R/conversion.R
```diff
@@ -219,10 +219,145 @@ get.adjedgelist <- function(
...
+# Recover a legacy positional `attr` argument passed via `...`.
+#
+# Before 3.0.0, `attr` was the 3rd positional slot of as_adjacency_matrix()
+# and as_biadjacency_matrix(). The new signatures insert `...` in that slot
+# and move `weights`/`attr` after the dots, so calls like
+#   as_adjacency_matrix(g, "both", "weight")
+# would otherwise hard-fail via rlang::check_dots_empty().
+#
+# Dispatcher is dm-style: discriminate on whether dots are named or not
+# (cynkra/dm R/filter-dm.R, dm_filter_api1()). Named extras stay an error
```
maelle: better to use a GitHub permalink?

R/conversion.R
```diff
@@ -334,18 +449,22 @@ get.adjacency.sparse <- function(
...
-#'   of an arbitrarily chosen edge (for the multiple edges) is included. This
-#'   argument is ignored if `edges` is `TRUE`.
-#'
-#'   Note that this works only for certain attribute types. If the `sparse`
-#'   argumen is `TRUE`, then the attribute must be either logical or
-#'   numeric. If the `sparse` argument is `FALSE`, then character is
-#'   also allowed. The reason for the difference is that the `Matrix`
-#'   package does not support character sparse matrices yet.
+#' @param ... These dots are for future extensions and must be empty.
+#' @param weights One of the following:
```
maelle, line 403: so this will be inherited all over the place?

R/conversion.R
```diff
@@ -334,18 +449,22 @@ get.adjacency.sparse <- function(
...
+#'     \item A numeric or logical vector of length [ecount()]: use these values
+#'       directly as edge weights.
+#'     \item A character scalar: the name of an edge attribute whose values are
+#'       used as weights. The attribute must be numeric or logical.
+#'   }
+#'   If multiple edges share endpoints, the value of an arbitrarily chosen edge
+#'   is included in the matrix.
+#' @param attr `r lifecycle::badge("deprecated")` Use `weights` instead. If
+#'   supplied, the value is forwarded to `weights` as a character edge
+#'   attribute name.
```
maelle, line 417: but this will be removed in future versions

maelle (commented):
a few (most off PR topic?) comments, thank you!!

R/centrality.R
schochastics: I think this could be a separate PR to rename all internal (used) "dot" functions

R/centrality.R
schochastics: Yes, these mirror the mathematical notation. I have no hard opinion on naming but in these cases I actually favor the single letter notations because it is easier to trace the (mathematical) meaning from the corresponding formulas

R/conversion.R
schochastics: I need to verify that

schochastics:
now using the generic migration tool

R/conversion.R
schochastics: This is the else of a type dispatch, not a deferred validation. It's only reachable once we've ruled out `NULL`, all-`NA`, and `character` which are all valid inputs handled by the branches above. Moving it earlier would reject those legitimate cases, since character is a valid type and the NA branch relies on `is.na()` matching across types. So it's correctly positioned as the final catch-all.

R/conversion.R
```diff
@@ -334,18 +399,22 @@ get.adjacency.sparse <- function(
...
-#'   in the adjacency matrix. If the graph has multiple edges, the edge attribute
-#'   of an arbitrarily chosen edge (for the multiple edges) is included. This
-#'   argument is ignored if `edges` is `TRUE`.
-#'
-#'   Note that this works only for certain attribute types. If the `sparse`
-#'   argumen is `TRUE`, then the attribute must be either logical or
-#'   numeric. If the `sparse` argument is `FALSE`, then character is
-#'   also allowed. The reason for the difference is that the `Matrix`
-#'   package does not support character sparse matrices yet.
+#' @param ... These dots are for future extensions and must be empty.
```
krlmlr:
```suggestion
#' @inheritParams rlang::args_dots_empty
```

Or in a future issue.

R/conversion.R
```diff
@@ -365,30 +434,74 @@ get.adjacency.sparse <- function(
...
 #' @family conversion
 #' @export
 as_adjacency_matrix <- function(
   graph,
   type = c("both", "upper", "lower"),
-  attr = NULL,
+  ...,
+  weights = NULL,
+  attr = deprecated(),
   edges = deprecated(),
```
krlmlr: Move to end?

R/conversion.R
```diff
@@ -1091,26 +1174,67 @@ get.incidence.sparse <- function(
 as_biadjacency_matrix <- function(
   graph,
   types = NULL,
-  attr = NULL,
+  ...,
+  weights = NULL,
+  attr = deprecated(),
```
krlmlr: Same?

tests/testthat/_snaps/conversion.md
```diff
@@ -23,34 +23,100 @@
...
+
+# as_adjacency_matrix() recovers legacy positional `attr`
+
+    Code
+      A <- as_adjacency_matrix(g, "both", "weight", sparse = FALSE)
+    Condition
+      Warning:
+      Calling `as_adjacency_matrix()` with positional or abbreviated arguments was deprecated in igraph 3.0.0.
+      i Detected call: as_adjacency_matrix(graph, type, attr)
+      i Use instead: as_adjacency_matrix(graph, type, weights = )
```
krlmlr: We are largely testing the arg fallback here and in the tests around this. Do we need these tests?

krlmlr (commented):
Nice, thanks! Merge conflicts now...

schochastics:
conflict was minor and I think I addressed all comments

krlmlr:
Thanks!

PR #2684, opened 2026-06-01
feat: add tooling for parameter migration

schochastics:
@krlmlr a suggestion derived from the discussions in #2677. This would be a way to generate arg migration functions.


tools/README.md
```diff
@@ -145,6 +145,72 @@ git subrepo status tools/py-stimulus
...
+
+### Known limitations
+
+The generator handles **pure renames and insertions** only:
+
+- **Removed args** — deleting an old positional slot (rather than renaming it)
+  makes a legacy positional call silently rebind to the wrong new slot. Not
+  detected.
+- **Reordered args** — the new signature must preserve the relative order of
+  surviving old args. Reordering breaks the position math.
```
krlmlr: I don't see a way to handle reordering. If reordering is needed, put args past the ellipsis.

tools/README.md
```diff
@@ -145,6 +145,72 @@ git subrepo status tools/py-stimulus
...
+- **Partial matching** — pre-`...` R allowed `f(g, att = "x")` to partial-match
+  `attrs`. Post-insertion, partial names fall into `...` and are treated as
+  unknown keywords (the strict `check_dots_empty()` path), never recovered.
```
krlmlr: This would be cool to have, though, with the same lifecycle warning.

tools/README.md
```diff
@@ -145,6 +145,72 @@ git subrepo status tools/py-stimulus
...
+**even under** `lifecycle_verbosity = "warning"`. Threading the *caller's* env
+through `.user_env` is what keeps the deprecation visible to users.
+
+### Known limitations
+
+The generator handles **pure renames and insertions** only:
+
+- **Removed args** — deleting an old positional slot (rather than renaming it)
+  makes a legacy positional call silently rebind to the wrong new slot. Not
+  detected.
```
krlmlr: Same solution: move past the ellipsis.

tools/README.md
```diff
@@ -145,6 +145,72 @@ git subrepo status tools/py-stimulus
...
+  )
+  graph <- args$graph; types <- args$types; weights <- args$weights
+  # ... rest of the body
+}
+```
+
+The `.user_env` argument is essential. `lifecycle::deprecate_soft()` only signals
+when `is_direct(user_env)` is `TRUE`; a package-internal `user_env` is swallowed
+**even under** `lifecycle_verbosity = "warning"`. Threading the *caller's* env
+through `.user_env` is what keeps the deprecation visible to users.
```
krlmlr: I don't buy this argument. Why can't the `.user_env` default be used? What does `is_direct()` do?

tools/migrations.R
```diff
@@ -0,0 +1,88 @@
...
+  # Exercises the generator end-to-end without migrating a real function.
+  # Old signature: f(a, b, c, d); new: f(a, b, ..., c_renamed, d) with c renamed
+  # to c_renamed. Consumed by tests/testthat/test-handle-args.R.
+  list(
+    fn = "migration_fixture",
+    old = c("a", "b", "c", "d"),
+    new = c("a", "b", "...", "c_renamed", "d"),
+    renames = c(c = "c_renamed"),
+    defaults = list(c_renamed = "NULL", d = "NULL"),
+    when = "3.0.0"
```
krlmlr: Perhaps a bit more idiomatic with less typing:

```suggestion
  migration_fixture = list(
    old = function(a, b, c = c_renamed, d) {},
    new = function(a, b, ..., c_renamed = NULL, d = NULL),
    when = "3.0.0"
```

Using `pairlist()` is more awkward.

.github/workflows/check-migrations.yaml
```diff
@@ -0,0 +1,42 @@
...
+          # The generator is base-R only; no package dependencies are needed.
+          use-public-rspm: true
+
+      - name: Regenerate migration helpers
+        run: Rscript tools/generate-migrations.R
+
+      - name: Fail if generated files drift from the registry
+        run: |
+          if ! git diff --exit-code -- 'R/handle-args-*.R'; then
+            echo "::error::R/handle-args-*.R is out of sync with tools/migrations.R."
```
krlmlr: This can be solved by running the generation as a side effect of `devtools::document()` or `pkgload::load_all()` (perhaps in a testthat helper). For speed, only if the timestamp of the source is newer than of all the generated files. I can dig up examples if needed, are we even using this pattern here?

Then, updating the handlers happens automatically.

tools/generate-migrations.R
```diff
@@ -0,0 +1,298 @@
...
+    "  primary <- if (length(renamed)) renamed[[1L]] else 1L",
+    "  lifecycle::deprecate_soft(",
+    paste0("    when = \"", entry$when, "\","),
+    paste0("    what = sprintf(\"", fn, "(%s)\", recover_old[[primary]]),"),
+    paste0("    with = if (recover_old[[primary]] != recover_new[[primary]]) sprintf(\"", fn, "(%s)\", recover_new[[primary]]),"),
+    "    details = c(",
+    paste0("      \"Positional use of the pre-", entry$when, " signature of `", fn, "()` is deprecated.\","),
+    "      i = paste0(",
+    "        \"Name the argument(s) instead: \",",
+    "        toString(sprintf(\"`%s = `\", recover_new[rebound])), \".\"",
```
krlmlr: Weird combo, prefer `paste0()` here and elsewhere?

tools/generate-migrations.R
```diff
@@ -0,0 +1,298 @@
...
+    "  # renamed slot if there is one, else the first recovered slot.",
+    "  rebound <- seq_along(dots)",
+    "  renamed <- rebound[recover_old[rebound] != recover_new[rebound]]",
+    "  primary <- if (length(renamed)) renamed[[1L]] else 1L",
+    "  lifecycle::deprecate_soft(",
+    paste0("    when = \"", entry$when, "\","),
+    paste0("    what = sprintf(\"", fn, "(%s)\", recover_old[[primary]]),"),
+    paste0("    with = if (recover_old[[primary]] != recover_new[[primary]]) sprintf(\"", fn, "(%s)\", recover_new[[primary]]),"),
+    "    details = c(",
+    paste0("      \"Positional use of the pre-", entry$when, " signature of `", fn, "()` is deprecated.\","),
```
krlmlr: Can we show the detected and the requested signature next to each other? I've also stumbled over the "pre-3.0.0" bit.

krlmlr (commented):
Nice, thanks for driving this!

.github/workflows/R-CMD-check.yaml
```diff
@@ -132,6 +132,20 @@ jobs:
           _R_SHLIB_STRIP_=true R CMD INSTALL .
         shell: bash
 
+      # Regenerate the argument-migration helpers from tools/migrations.R and
+      # fail if they drift. Runs from the source checkout (tools/ is excluded
+      # from the built package) and before the auto-commit steps below, so a
+      # contributor who forgot to regenerate gets a hard failure.
+      - name: Check migration helpers are in sync with the registry
+        run: |
+          Rscript tools/generate-migrations.R
```
krlmlr: We'd have to do it in a custom before- or after-action. Even better, as part of `load_all()`. Why has it landed here?

tools/migrations.R
```diff
@@ -0,0 +1,77 @@
...
+  #                   sparse = FALSE) {},
+  #   when = "3.0.0"
+  # ),
+
+  # --- test fixture --------------------------------------------------------
+  # Exercises the generator end-to-end without migrating a real function.
+  # Old signature f(a, b, c, d); new f(a, b, ..., c_renamed, d) with c renamed
+  # to c_renamed. Consumed by tests/testthat/test-handle-args.R.
+  migration_fixture = list(
+    old = function(a, b, c = c_renamed, d) {},
```
krlmlr: Thinking again, this assumes that the default values don't change as part of this exercise. Is this a restriction/limitation? If not relevant, we should still document it. If yes, we can still reintroduce the `renames` arg. Sorry for the mess.

tools/README.md
```diff
@@ -145,6 +145,81 @@ git subrepo status tools/py-stimulus
...
+The `.user_env` argument matters because the helper interposes an extra call
+frame. `lifecycle::deprecate_soft()` only signals when `is_direct(user_env)` is
+`TRUE` — that is, when `topenv(user_env)` is the global env (or under testthat).
+Its auto-detected `user_env` would resolve to the helper's caller, the **igraph
+namespace**, so the warning is swallowed (this gate runs *before* the
+`lifecycle_verbosity = "warning"` override, so forcing verbosity does not bring
+it back). Threading the public function's `rlang::caller_env()` through
+`.user_env` makes the *user's* frame the reference: real user calls warn, while
+genuinely internal igraph callers stay correctly silent — the point of a *soft*
+deprecation.
```
krlmlr: I still don't buy it. What if the default of the handlers is `rlang::caller_env(2)` so that it works in the vast majority of cases (handler is called directly at entry)?

tools/README.md
```diff
@@ -145,6 +145,81 @@ git subrepo status tools/py-stimulus
...
+```r
+as_biadjacency_matrix <- function(graph, types, ..., weights = NULL,
+                                  attr = lifecycle::deprecated(),
+                                  names = TRUE, sparse = FALSE) {
+  args <- handle_args_as_biadjacency_matrix(
+    graph, types, ..., weights = weights, attr = attr,
+    names = names, sparse = sparse,
+    .user_env = rlang::caller_env()
+  )
+  graph <- args$graph; types <- args$types; weights <- args$weights
```
krlmlr: Prefer spread out per line.

Crazy idea (with the handler updating the caller's environment):

```suggestion
  handle_args_as_biadjacency_matrix(
    graph, types, ..., weights = weights, attr = attr,
    names = names, sparse = sparse,
    .user_env = rlang::caller_env()
  )
```

krlmlr (commented):
Beautiful!

krlmlr:
And if we think that a little further: what if our code generator patches the code in place, delineated by `# BEGIN ARG_HANDLE` and `# END ARG_HANDLE` comments? No handler function, no magic, no caller env crap.

.github/workflows/R-CMD-check.yaml
schochastics: Good question. I iterate again

schochastics:
> And if we think that a little further: what if our code generator patches the code in place, delineated by `# BEGIN ARG_HANDLE` and `# END ARG_HANDLE` comments? No handler function, no magic, no caller env crap.

Interesting idea, I will iterate a bit more

schochastics:
Iterated one more time

.github/workflows/custom/after-install/action.yml
```diff
@@ -8,3 +8,18 @@ runs:
...
+    # if they have drifted. Runs from the source checkout (R is available after
+    # install; tools/ is excluded from the built package) so a contributor who
+    # forgot to regenerate gets a hard failure. Local dev refreshes them
+    # automatically via tests/testthat/helper-migrations.R.
+    - name: Regenerate argument-migration blocks and check for drift
+      run: |
+        Rscript tools/generate-migrations.R
+        if ! git diff --exit-code -- R; then
+          echo "::error::Generated ARG_HANDLE blocks are out of sync with tools/migrations.R."
+          echo "Run 'Rscript tools/generate-migrations.R' and commit the result."
```
krlmlr, line 22: The existing pipeline handles committing/user notifications (for external contributors), if the check is dropped, the right thing happens most of the time.

No action needed now, but we might just drop that later.

R/migration-fixture.R
```diff
@@ -0,0 +1,107 @@
...
+# It exists only to exercise the generator end-to-end; see
+# tests/testthat/test-handle-args.R.
+
+#' @noRd
+migration_fixture <- function(a, b, ..., c_renamed = NULL, d = NULL) {
+  # BEGIN GENERATED ARG_HANDLE: migration_fixture
+  # Generated from tools/migrations.R by tools/generate-migrations.R -- do not edit.
+  # Recovers a legacy call to the pre-migration signature; regenerate with
+  # `Rscript tools/generate-migrations.R`.
+  .arg_handle <- (function(dots, current) {
```
krlmlr: Why are we back to a closure? I fear this will be hard to debug.

If we need something that `match.call()` or the rlang equivalent can't give us, can we use a narrowly scoped function that accepts arguments and returns a list, or similar?

R/migration-fixture.R
```diff
@@ -0,0 +1,107 @@
...
+# f(a, b, c, d) -- now f(a, b, ..., c_renamed, d) with c renamed to c_renamed.
+# It exists only to exercise the generator end-to-end; see
+# tests/testthat/test-handle-args.R.
+
+#' @noRd
+migration_fixture <- function(a, b, ..., c_renamed = NULL, d = NULL) {
+  # BEGIN GENERATED ARG_HANDLE: migration_fixture
+  # Generated from tools/migrations.R by tools/generate-migrations.R -- do not edit.
+  # Recovers a legacy call to the pre-migration signature; regenerate with
+  # `Rscript tools/generate-migrations.R`.
```
krlmlr: TMI?

```suggestion
  # BEGIN GENERATED ARG_HANDLE: migration_fixture, do not edit, see tools/generate-migrations.R
```

krlmlr (commented):
Sweet! We're getting there. I really like the new inline checkers, really looking forward to rolling it out.

I wonder if we could also get a series of snapshot tests that exercise the migration fixture with lifecycle warnings enabled.

schochastics:
Switched the inline closure for a plain named helper `migrate_recover_args()` in `R/migrate-args.R`. The generated block just passes config and assigns results, with `deprecate_soft()` still inline. Also collapsed the generated comment onto the marker. Left the after-install drift check for now

R/migration-fixture.R
```diff
@@ -0,0 +1,33 @@
...
+# tools/generate-migrations.R). `migration_fixture()` carries a generated
+# ARG_HANDLE block that recovers a legacy call to its pre-3.0.0 signature
+# f(a, b, c, d) -- now f(a, b, ..., c_renamed, d) with c renamed to c_renamed.
+# It exists only to exercise the generator end-to-end; see
+# tests/testthat/test-handle-args.R.
+
+#' @noRd
+migration_fixture <- function(a, b, ..., c_renamed = NULL, d = NULL) {
+  # BEGIN GENERATED ARG_HANDLE: migration_fixture, do not edit, see tools/generate-migrations.R
+  .arg_handle <- migrate_recover_args(
```
krlmlr: One last change request: let's guard this with `if (...length() > 0)` at the call site, for performance. Function calls are expensive in R. We might then avoid the `if (!is.null())` check below.

krlmlr (commented):
Wonderful!

I'd still like to see a `test-migration-fixture.R` with many snapshot tests. May need to extend the arg names to also cover the "partial matching" case.

Alternatively (or in addition), we can start applying this to a few functions.

schochastics:
If this is ok we could merge to main and apply in #2677?

tools/migrations.R
```diff
@@ -0,0 +1,93 @@
...
+  # ),
+
+  # --- test fixture --------------------------------------------------------
+  # Exercises the generator end-to-end without migrating a real function. The
+  # arg names are chosen to cover every recovery path: a rename (`weight ->
+  # weights`), unique abbreviations (`ty`, `dir`) and an ambiguous one (`weig`
+  # matches both `weight` and `weights`). Consumed by
+  # tests/testthat/test-migration-fixture.R.
+  migration_fixture = list(
+    old = function(graph, n, weight = weights, type, directed) {},
```
krlmlr, line 82: Playing around the following axis would be great in a follow-up PR, including calls with `g = ...` or `gr = ...` instead of `graph = ...`:


```suggestion
    old = function(graph, n, weight = weights, kind = type, directed) {},
```

krlmlr (approved):
Thanks, great! Would you like to do a follow-up PR for refining this?

PR #2685, opened 2026-06-03
feat: add touchstone benchmarks

### LINKED ISSUE #1338: How do we check changes don't affect performance

schochastics:
fix #1338

So far, the benchmarks only cover the R heavy functions. We can include c level benchmarks but I am in favor of keeping it on the R side


krlmlr:
Thanks!

PR #2686, opened 2026-06-03
fix: harden air support and broaden migration fixture to the full arg-matching matrix

schochastics:
Follow-up to #2684. Add a second rename to the fixture (`kind -> type`, in addition to `weight -> weights`) and snapshot the cross-product of call styles: positional, renamed-away old names, unique and ambiguous abbreviations, mixed positional+named, and head-arg abbreviations (`gr =`, `g =`).

Confirms the two matching regimes compose: head args before `...` are resolved silently by base R's partial matching (not deprecated), while renamed/abbreviated args after `...` are recovered with a single soft-deprecation. No behaviour change -- fixture and tests only.

@krlmlr did you expect any behaviour changes or just these tests?

### DIFF
--- a/tests/testthat/test-migration-fixture.R
+++ b/tests/testthat/test-migration-fixture.R
@@ -67,20 +67,17 @@ test_that("positional and named recovery can be mixed in one call", {
   expect_equal(res$type, "out")
 })
 
-test_that("head args are matched by base R, including abbreviations", {
-  rlang::local_options(lifecycle_verbosity = "warning")
-  # Abbreviating a head arg (before `...`) is plain R partial matching: it is
-  # resolved silently and is *not* deprecated.
-  expect_no_warning(res <- migration_fixture(gr = "G", n = 5))
-  expect_equal(res$graph, "G")
-
-  # A head abbreviation alongside a recovered tail arg: head stays silent, the
-  # tail arg is recovered with a single deprecation.
-  lifecycle::expect_deprecated(
-    res <- migration_fixture(g = "G", 5, kind = "in")
+test_that("head args go through base R partial matching, not our recovery", {
+  # Abbreviating a head arg (before `...`) is plain R partial matching, not our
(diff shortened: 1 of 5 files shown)


tests/testthat/test-migration-fixture.R
```diff
@@ -60,6 +67,22 @@ test_that("positional and named recovery can be mixed in one call", {
   expect_equal(res$type, "out")
 })
 
+test_that("head args are matched by base R, including abbreviations", {
+  rlang::local_options(lifecycle_verbosity = "warning")
+  # Abbreviating a head arg (before `...`) is plain R partial matching: it is
+  # resolved silently and is *not* deprecated.
+  expect_no_warning(res <- migration_fixture(gr = "G", n = 5))
```
krlmlr: Prefer snapshot test with the corresponding base R warning enable.

tests/testthat/test-migration-fixture.R
```diff
@@ -80,10 +103,13 @@ test_that("recovery deprecation messages", {
   expect_snapshot(x <- migration_fixture("g", 5, 1:3, "in", TRUE))
   expect_snapshot(x <- migration_fixture("g", 5, 1:3))
   expect_snapshot(x <- migration_fixture("g", 5, weight = 1:3))
-  expect_snapshot(x <- migration_fixture("g", 5, ty = "in"))
+  expect_snapshot(x <- migration_fixture("g", 5, kind = "in"))
+  expect_snapshot(x <- migration_fixture("g", 5, kin = "in"))
```
krlmlr: Should we just omit the assignment to x to actually print the arguments recognized?

krlmlr (commented):
Nice, thanks!

schochastics:
caught an issue when the generated code has lines longer than 80 chars (we dont want to run air after generation). the fix is now also part of this PR

krlmlr:
Call `air` on the affected files after updating?

PR #2688, opened 2026-06-04
fix: repair touchstone benchmark CI

schochastics:
The touchstone benchmark CI added in #2685 failed on its first run ([example](https://github.com/igraph/rigraph/actions/runs/26946685203/job/79501303006)). Three linked problems:

### 1. Receive workflow skipped every maintainer PR (root cause)
`prepare` gated on `author_association == OWNER/MEMBER/COLLABORATOR`. That value comes from the `pull_request` **webhook payload**, which reports `CONTRIBUTOR`/`NONE` (never `MEMBER`) for maintainers whose org membership is **private** — so the gate was false, both jobs skipped, and no `"pr"` artifact was produced.

→ Gate on the PR head being a **same-repo branch** instead. Creating a same-repo branch requires push access, so this is the real security boundary against untrusted fork code, and it's immune to membership visibility.

### 2. Comment workflow crashed instead of no-op'ing (the visible error)
It only checked `workflow_run.event == 'pull_request'`, so it ran after the skipped receive run and died downloading the missing artifact: `Cannot read properties of undefined (reading 'id')`.

→ Also require `workflow_run.conclusion == 'success'`.

### 3. Deprecated `::set-output` (latent)
Would have left `needs.prepare.outputs.config` empty once `prepare` actually ran, breaking the `fromJson` build matrix.

→ Switch to `$GITHUB_OUTPUT`, compacting `config.json` with `jq -c` (drops the multi-line escaping hack).

This PR touches `.github/workflows/touchstone-*.yaml`, which is in the receive trigger's path filter, so it exercises the fixed workflow itself.

### DIFF
--- a/.github/workflows/touchstone-comment.yaml
+++ b/.github/workflows/touchstone-comment.yaml
@@ -26,6 +26,10 @@ jobs:
       github.event.workflow_run.event == 'pull_request' &&
       github.event.workflow_run.conclusion == 'success'
     steps:
-      - uses: lorenzwalthert/touchstone/actions/comment@v1
+      # Pinned to a default-branch SHA: the released tags (v1, v1.0.1) still
+      # use actions/{upload,download}-artifact@v2, which GitHub now auto-fails.
+      # This commit uses artifact@v4/v5 and github-script@v7. Bump to a tag once
+      # upstream cuts a release that includes the fix.
+      - uses: lorenzwalthert/touchstone/actions/comment@7e6072a25d1f8d72649188bc711113039ef94b75
         with: 
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/touchstone-comment.yaml
+++ b/.github/workflows/touchstone-comment.yaml
@@ -18,8 +18,13 @@ permissions:
 jobs:
   upload:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


schochastics:
### Validation status

The **receive** workflow runs from the PR branch and is fully green here: it ran the benchmarks and uploaded the `pr`, `results`, and `visual-benchmarks` artifacts ([run](https://github.com/igraph/rigraph/actions/runs/26949110547)). This validates the three receive-side fixes (same-repo gate, `$GITHUB_OUTPUT`, SHA pin).

The **comment** workflow is `workflow_run`-triggered, which GitHub **always runs from the default branch (`main`)** — never from the PR branch. So the "Continuous Benchmarks (Comment)" check on this PR still executes the *old* `main` code (`comment@v1` + `github-script@v3`) and fails with `401 InvalidAuthenticationInfo` while downloading the artifact from Azure storage — the exact bug fixed by pinning to a commit that uses `actions/download-artifact@v5`.

**That red check is expected and cannot turn green until this PR is merged.** The pinned comment action is a verified drop-in (same `GITHUB_TOKEN` input; replaces the hand-rolled github-script download with `actions/download-artifact@v5`). After merge, the next PR touching `R/**` / `touchstone/**` will exercise the full receive → comment flow end-to-end.

Issue #2695, opened 2026-06-05
`return.vs.es`: fix performance, then deprecate and remove


schochastics:
This issue unifies three related threads:

- **#1652** — remove the `return.vs.es` option.
- **#1614** — some functions ignore `return.vs.es` (consistency); partially
  addressed by PR #2439.
- **#557** — `igraph.vs`/`igraph.es` are much slower than plain edge lists/indices.

They are the same problem from three angles: the option exists mainly for
**performance** (#557), it is applied **inconsistently** (#1614), and the
long-term goal is to **remove it** (#1652) with `as.integer()` as the
replacement. Removal is blocked until the performance gap is closed — e.g.
`max_cliques(sample_gnp(500, 0.15))` is ~8× slower with `return.vs.es = TRUE`
(268 ms vs 33 ms) because `create_vs()` builds names eagerly. The endorsed
direction is lazy (ALTREP) naming so `igraph.vs`/`igraph.es` is as cheap as a
numeric vector and names are materialized only on print/use.

### Steps to close

**Phase 0 — Close the performance gap (prerequisite)**
- [ ] Make vertex/edge sequence construction lazy (ALTREP-backed names), so
      `create_vs()`/`create_es()` and `E()` defer the `names`/`vnames` work
      until printed or accessed.
- [ ] Benchmark gate: `max_cliques(sample_gnp(500, 0.15))` with the default
      approaches the old `return.vs.es = FALSE` timing. **If the gap isn't
      closed, stop here** — do not proceed to removal.

**Phase 1 — Decouple internal usage from the public option**
- [ ] Give the `_impl` layer a private numeric-output path that does not read
      `return.vs.es`.
- [ ] Replace the `with_igraph_opt(list(return.vs.es = FALSE), ...)` wrappers in
      `all_simple_paths()` and `get_edge_ids()` with that mechanism (no
      user-visible change).

**Phase 2 — Soft-deprecate the option**
- [ ] Emit `lifecycle::deprecate_warn()` when a user sets `return.vs.es`,
      pointing to `as.integer()` / `as_ids()`.
- [ ] Update the option's roxygen doc entry with a deprecation badge + recipe.
- [ ] Add a NEWS entry. Keep the option functional for at least one release.

**Phase 3 — Remove**
- [ ] Drop the `if (igraph_opt("return.vs.es"))` guard from the `OUTCONV`
      templates in `tools/stimulus/types-RR.yaml` (keep the `create_vs`/
[...]

Issue #1652, opened 2025-01-09
Remove `return.vs.es` option


krlmlr:
Motivation: https://github.com/igraph/rigraph/issues/1614#issuecomment-2579684587

schochastics:
now tracked in #2695

PR #2696, opened 2026-06-05
perf: Lazy ALTREP names for vertex/edge sequences

schochastics:
Split 4 of 4, per https://github.com/igraph/rigraph/pull/2696#issuecomment-4989953620 — this is the "**if needed**, ALTREP on top" step. Rebased onto the stack, so the diff here is now *only* the ALTREP layer.

**Base:** #2888. Review order: #2886 -> #2887 -> #2888 -> this.

Part of #2695, relates to #1652.

### Honest answer to "if needed": mostly not

With the first three PRs merged, ALTREP no longer moves the flagship case at all. Measured head-to-head (#2888 vs this PR, same machine, warm, 20-30 iterations):

| | #2888 (eager names) | this PR (ALTREP) | |
|---|---|---|---|
| `max_cliques_named` | 2.667 ms | 2.674 ms | unchanged |
| `max_cliques_sizes_named` | 3.144 ms | 3.204 ms | unchanged |
| `all_simple_paths_named` | 1.939 ms | 1.922 ms | unchanged |
| `ego_order2_named` | 4.427 ms | 3.134 ms | **-29%** |
| `vs_subset_positional` | 0.111 ms | 0.095 ms | **-14%** |

The reason `max_cliques()` sees nothing: once the construction loop is in C (#2888), building the names eagerly is just a run of `SET_STRING_ELT` against already-interned `CHARSXP`s. No allocation of new strings, no R-level copy. Laziness has almost nothing left to save.

Where it *does* help is where the name vectors are large relative to the work: `ego(order = 2)` builds 2000 sequences whose neighborhoods are a large fraction of the graph, and positional subsetting of a 50k-vertex sequence composes indices in O(1) instead of copying 10k names.

So this is a real but narrow win, and it is the one layer of the four that carries ongoing maintenance cost: a custom ALTREP class with `Length`/`Dataptr`/`Dataptr_or_null`/`Extract_subset`/`Elt` methods, and a dependence on R's ALTREP API staying stable. **I'd be fine closing this** if the maintenance cost isn't worth 29% on `ego()` — #2886-#2888 already deliver 9.3× on the case #1652 was actually about.

### What changed

A sequence's `names` attribute becomes an instance of an `igraph_lazy_names` ALTREP string class instead of a materialized character vector. `data1` holds `list(source, idx)` — the graph's full name vector, shared by reference across every sequence of that graph, plus a 1-based index into it; `data2` caches the materialized `STRSXP` once anything reads it.

- `V()`/`E()` build their names through `lazy_index_names()`.
- `Rx_igraph_vs_list()` attaches the ALTREP directly instead of eagerly subsetting.
- `Extract_subset` returns a fresh lazy vector with composed indices, so subsetting a sequence never materializes; it falls back to the default for index types it does not handle.
- `DATAPTR_RO`/`DATAPTR_OR_NULL` are used rather than `DATAPTR`, which is non-API as of R 4.5.

### Correctness

Same checks as #2888, re-run here: integer payloads with identical values and names against the `return.vs.es = FALSE` path, no `names` attribute at all on unnamed graphs, `NA`/out-of-range IDs map to `NA_STRING`, double ID vectors coerced, caller inputs unmutated, and `get_vs_graph(seq)` still `NULL` after `rm(graph); gc()`.

Compiles without new warnings, clean under `gctorture(TRUE)`.

Full suite: **FAIL 0 | WARN 0 | SKIP 7 | PASS 9299**.

### Scope

Vertex sequences and the `names` of edge sequences. Edge `vnames` ("tail|head") are still built eagerly; a lazy `vnames` and a `create_es_list()` batch form remain follow-ups.

### On `return.vs.es`

The original framing of this PR still holds, and is now better supported. #1652 was blocked on an 8x gap between the default and `return.vs.es = FALSE` that stibu81 called "unusable". After #2886-#2888 that gap is ~1.15x. The option's main justification is gone, so it can be deprecated and left inert rather than urgently removed — a separate decision, untouched by this stack.


man/biconnected.components.Rd
```diff
@@ -17,7 +17,7 @@ it is directed.}
 consistent API.
 }
 \section{Related documentation in the C library}{
-\href{https://igraph.org/c/html/0.10.17/igraph-Structural.html#igraph_biconnected_components}{\code{biconnected_components()}}, \href{https://igraph.org/c/html/0.10.17/igraph-Basic.html#igraph_vcount}{\code{vcount()}}, \href{https://igraph.org/c/html/0.10.17/igraph-Basic.html#igraph_edges}{\code{edges()}}, \href{https://igraph.org/c/html/0.10.17/igraph-Basic.html#igraph_get_eids}{\code{get_eids()}}, \href{https://igraph.org/c/html/0.10.17/igraph-Basic.html#igraph_ecount}{\code{ecount()}}
```
maelle, line 20: should this be part of this diff?

R/iterators.R
```diff
@@ -276,8 +292,46 @@ unsafe_create_vs <- function(graph, idx, verts = NULL) {
...
-  res <- simple_vs_index(verts, idx, na_ok = TRUE)
-  add_vses_graph_ref(res, graph)
+  # `idx` are vertex IDs straight from C, and `verts` is the full `V(graph)`,
+  # so `verts[idx]` would just be `idx` again -- skip that copy and use the
+  # IDs directly as the payload. Names are taken lazily from `verts` (an
+  # ALTREP that composes in O(1) under subsetting), and the graph reference is
+  # shared from `verts`. All attributes are set in one `attributes<-` call to
+  # avoid the per-object shallow copies that dominate when many sequences are
+  # built (e.g. `max_cliques()`).
+  nms <- attr(verts, "names")
```
maelle: why not `vertex_names`? The code is already complicated enough, no need to add mysterious names 😉

R/iterators.R
```diff
@@ -276,8 +292,46 @@ unsafe_create_vs <- function(graph, idx, verts = NULL) {
...
+  res <- as.integer(idx)
```
maelle, line 327: I wanted to suggest `rlang::set_attrs()` but it is deprecated 😅

R/iterators.R
```diff
@@ -276,8 +292,46 @@ unsafe_create_vs <- function(graph, idx, verts = NULL) {
...
+#
+# This is the batch form of `unsafe_create_vs()` and replaces the
+# `lapply(idx_list, unsafe_create_vs, graph = graph, verts = V(graph))`
+# pattern. The per-graph work -- `V(graph)`, the shared weak reference, the
+# graph id and the lazy name source -- is hoisted out of the loop, so each
+# sequence only costs one `as.integer()` and one `attributes<-`. This is what
+# brings construction of many sequences (e.g. `max_cliques()`) down close to
+# the cost of returning bare numeric indices.
+create_vs_list <- function(graph, idx_list) {
+  # `verts <- V(graph)` is what mints the single shared weak reference and graph
```
maelle, line 347: combine with comment above?

touchstone/script.R
```diff
@@ -177,5 +177,122 @@ benchmark_run(
...
+# Functions that return (many) vertex/edge sequences pay for building the
+# `names`/`vnames` attribute and attaching a graph reference to every object.
+# These benchmarks exercise that construction path on *named* graphs, where
+# the cost is highest. `max_cliques()` is the canonical case: it returns tens
+# of thousands of vertex sequences, one per clique.
+# ---------------------------------------------------------------------------
+benchmark_run(
+  expr_before_benchmark = {
+    library(igraph)
+    set.seed(42)
```
maelle, line 191: why don't we use withr here? I'm genuinely curious.

schochastics: Claude not a fan, maybe? I nudge to it 😆

man/biconnected.components.Rd
schochastics: not sure why this is a diff here

schochastics: maybe it disappears when rerunning document

touchstone/script.R
schochastics: Ah I think because these scripts are supposed to be quite dependency free? All that is needed is touchstone and igraph

krlmlr:
Conflicting now.

krlmlr (commented):
Perhaps use an existing ALTREP class from sparsevctrs or stringfish? What's the performance tradeoff?

``` r
.Internal(inspect(letters))
#> @11ae9a930 16 STRSXP g1c7 [MARK,REF(1)] (len=26, tl=0)
#>   @12ba04b40 09 CHARSXP g1c1 [MARK,REF(452),gp=0x61] [ASCII] [cached] "a"
#>   @12cb694c8 09 CHARSXP g1c1 [MARK,REF(12),gp=0x61] [ASCII] [cached] "b"
#>   @12b013488 09 CHARSXP g1c1 [MARK,REF(426),gp=0x61] [ASCII] [cached] "c"
#>   @12baba488 09 CHARSXP g1c1 [MARK,REF(6),gp=0x61] [ASCII] [cached] "d"
#>   @12b059a10 09 CHARSXP g1c1 [MARK,REF(562),gp=0x61] [ASCII] [cached] "e"
#>   ...
.Internal(inspect(sparsevctrs::as_sparse_character(letters)))
#> @10d3a45a8 16 STRSXP g0c0 [] sparsevctrs_altrep_sparse_string (materialized=F, length=26)
.Internal(inspect(sparsevctrs::as_sparse_character(letters)[1:3]))
#> @10d3dc2a8 16 STRSXP g0c0 [] sparsevctrs_altrep_sparse_string (materialized=F, length=3)
```

<sup>Created on 2026-07-09 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

schochastics:
## Re: reusing an existing ALTREP class (sparsevctrs / stringfish)

Short version: neither fits, because this class isn't really a "string vector" — it's an **indexed gather-view over a shared parent vector**, which is a different ALTREP pattern than what either library implements.

### What this class actually is

`igraph_lazy_names` stores `data1 = list(source, idx)`, where `source` is the graph's **full** name vector (shared by reference across every sequence of that graph) and `idx` is a small 1-based integer index. The logical value is `source[idx]` — a gather (permutation-with-repeats) of the shared source. Two properties fall out of this:

1. **O(1) construction, O(len(idx)) memory per sequence** — building e.g. 30k cliques attaches 30k tiny integer `idx` vectors plus *one* shared `source`. No strings copied, no CHARSXP touched.
2. **O(1) composable subsetting** — `Extract_subset` just does `idx <- idx[i]` and returns a new view; no name copying.

### Why sparsevctrs doesn't fit

`sparse_character` compresses a vector with one dominant *default* value (non-defaults + positions + scalar default + length). Vertex/edge names are a gather from a shared source — generally all distinct, no dominant default — so the sparsity ratio is ~1.0 and you'd store every value *plus* every position: strictly worse than a plain character vector. It also has no shared-parent concept, so each sequence materializes its own strings (loses win #1), and `[` copies rather than composing an index (loses win #2).

### Why stringfish doesn't fit

`sf_vec` is an ALTREP STRSXP backed by its own `std::vector<sfstring>`, built to bypass R's global string hash table and keep string *ops* in ALTREP form. It owns its own copy of the strings, so constructing one from `source[idx]` materializes the gather at build time — back to per-sequence string work for all 30k sequences (loses win #1). No shared source, no index-composition on subset (loses win #2). Its actual value (skipping R's string cache) is orthogonal to what we're optimizing: we're avoiding building the vector *at all*, not building it more cheaply.

### Dependency cost (independent of fit)

igraph is `LinkingTo: cpp11` only, with a deliberately lean import list. `stringfish` would add `Rcpp` (linking + compiled runtime) into a cpp11 codebase; `sparsevctrs` adds a tidymodels-orbit dep that we'd import specifically to *not* use its core feature. For a base-of-the-ecosystem package, adding a dependency to replace ~180 lines of self-contained C we already control is a poor trade even before the fit problem.

### Summary

| Approach | Construct 30k seqs | Mem/seq | Subset `v[i]` | New dep |
|---|---|---|---|---|
| **`igraph_lazy_names` (this PR)** | O(1)/seq, no strings | tiny int `idx` + shared source | O(1) view | none |
| `sparsevctrs` | materialize all strings | ≥ full char vec (no sparsity) | copy | tidymodels dep |
| `stringfish` | materialize all strings | full string copy/seq | copy | Rcpp dep |

**Recommendation: keep the custom class.** The abstraction we need — a lazy indexed view into a shared vector — is genuinely bespoke; neither library exposes it, and both would regress exactly the construction-time and subsetting costs. The closest published analog is R's own `deferred_string` ALTREP (used for `as.character()` on int/real) and the general "row-index view" pattern — not a string-storage class.

krlmlr:
Let's split this in three separate PRs:

- `lapply()` -> `create_vs_list()`
- C implementation of `create_vs_list()` without ALTREP
- If needed, ALTREP on top

krlmlr:
Did we agree to split this PR?

schochastics:
Yes, I just did not have the time yet to do anything

schochastics:
Split done, per @krlmlr's https://github.com/igraph/rigraph/pull/2696#issuecomment-4989953620. The stack:

| PR | layer | base | `max_cliques` (named, 200 nodes) |
|---|---|---|---|
| — | `main` | — | 24.7 ms |
| #2886 | share the graph weak-reference, set attributes in one pass | `main` | 7.8 ms (3.2×) |
| #2887 | `lapply()` -> `create_vs_list()` | #2886 | 5.6 ms (4.4×) |
| #2888 | C `create_vs_list()`, **no ALTREP** | #2887 | 2.65 ms (9.3×) |
| #2696 | ALTREP on top | #2888 | 2.67 ms (unchanged) |

Two notes on the split as requested:

**There is a fourth PR, and it goes first.** Profiling the split showed the largest single win — the shared weak reference plus one-pass attribute setting — is independent of all three requested steps and accounts for most of the total. It is #2886, ~15 lines in `iterators.R`, no codegen and no C.

**The "if needed" on ALTREP resolves to "mostly not".** Once the construction loop is in C, eager names are just `SET_STRING_ELT` over interned `CHARSXP`s, so `max_cliques()` sees nothing from laziness. It still buys 29% on `ego(order = 2)` and 14% on positional subsetting of a large sequence — real, but narrow against the cost of maintaining a custom ALTREP class. Details and numbers in the rewritten description above; I am happy to close #2696 if that trade does not look worth it.

Each PR was verified independently: full suite `FAIL 0 | WARN 0 | SKIP 7 | PASS 9299`, and #2888/#2696 additionally checked against the `return.vs.es = FALSE` numeric path and under `gctorture(TRUE)`.

schochastics:
These four are now a native GitHub stack (**stack #2890**), so the stack map is on each PR and #2886–#2888 can be merged from the stack UI — retargeting and rebasing of the layers above cascade automatically, including under this repo's squash merges. The "Base:" lines in the descriptions are just there for reading order.

krlmlr:
PRs 1-3 in the stack need conflict resolution and perhaps more touchstone benchmarks, need to understand why this PR is slower for some examples.

Issue #550, opened 2022-07-16
delete_edges bug and counterintuitive add_edges behaviour (and workaround)


jessexknight:
**Describe the bug**
The documented behaviour of `delete_edges` is not correct, and `add_edges` does some counterintuitive things.

**To reproduce**
```R
library('igraph')
set.seed(1)
G = make_ring(10)
E(G)$color='blue'
e = get.edgelist(G) # 10x2 matrix
i = sample(ecount(G),3) # random edges
print(e[i,]) # random edges' vertices
layout = layout_nicely(G) # save layout
plot(G,layout=layout); print(ecount(G))
G = delete_edges(G,apply(e[i,],1,paste0,collapse='|')) # OK
# G = delete_edges(G,c(t(e[i,]))) # incorrect (1)
plot(G,layout=layout); print(ecount(G))
# G = add_edges(G,apply(e[i,],1,paste0,collapse='|'),color='red') # error (2)
G = add_edges(G,c(t(e[i,])),color='red') # OK
plot(G,layout=layout); print(ecount(G))
```
correctly yields:
```
     [,1] [,2]
[1,]    9   10
[2,]    4    5
[3,]    7    8
[1] 10
[1] 7
[1] 10
```
However, the commented uses of `delete_edges` (1) and `add_edges` (2) do not work:

1. `delete_edges` with the numeric argument removes twice as many edges as described in docs, yielding "10, 4, 7" for the last three outputs instead of "10, 7, 10".
2. `add_edges` with the character argument throws an error `Invalid vertex name(s)`. While this input format is not documented, I was expecting symmetry with `delete_edges`.
3. Semi-related: it's a bit counterintuitive that passing the matrix `get.edgelist` directly to `add_edges` results in the wrong edges, since R iterates over `matrix` data by row _then_ column. So, the matrix must be transposed (the `c()` above is actually not needed).

**Version information**
igraph v1.3.2 on linux mint 20.1

clpippel:
The documentation here is rather terse. It mentions the edgelist, but there are many formats that sometimes work and sometimes don't. Consider for example.
```
g1 <- graph_from_literal(A-B, B-C, C-D, D-E)
[1] A--B B--C C--D D--E
```
then
```
g2 <- delete_edges(g, get.edgelist(g1)) produces error message: Error in as.igraph.es(graph, edges) : Invalid edge names
[...]

PR #2700, opened 2026-06-17
feat: accept two-column matrices/data frames in edge functions (#1827)

### LINKED ISSUE #1827: adjust `add_edges()` and related functions to `get_edge_ids()`
`get_edge_ids()` allows nx2 matrices and data frames now for the edge attribute. Once #1798 is settled, these changes need to also be applied to:

- [ ] `add_edges()`

schochastics:
## Summary

Closes #1827. `get_edge_ids()` was changed in #1663 to accept the natural edge-list forms — an n×2 matrix or a two-column data frame — via the `el_to_vec()` helper. This PR applies the same treatment to the other functions that take edges as pairs of vertices, so the *same* edge list can be passed everywhere with no manual transposition (the core complaint in #550):

- `add_edges()` / `delete_edges()`
- `edge()` / `edges()` (via the `+` and `-` operators)
- **`make_graph()`** (and `make_directed_graph()` / `make_undirected_graph()`)
- `E(graph, P = )`

For all of these, an n×2 matrix or two-column data frame is now the canonical edge-list form. 2×n and 2×2 matrices are deprecated (defunct), consistent with `get_edge_ids()`.

## ⚠️ Behavior change for `make_graph()` (please read)

Unlike the other functions, **`make_graph()` already accepted a matrix** — but it flattened it **column-major**. The practical consequences:

- A plain n×2 edge list (e.g. `as_edgelist(g)`) silently produced the **wrong** graph.
- The **transposed** 2×n form (`t(edgelist)`) was the working idiom.

This PR makes the n×2 matrix the correct, canonical form and turns the 2×n form into a defunct error. **This is a breaking change**: code relying on `make_graph(t(el))` must now pass `make_graph(el)` (n×2) directly. This is intentional and is exactly what #550 asked for, but it is more disruptive than for the other functions, which previously errored or mis-handled matrices rather than silently accepting a transposed one.

## Implementation notes

- `el_to_vec()` gains `arg`/`fn` parameters so error/deprecation messages name the calling function (e.g. *"the `edges` argument of `add_edges()`"*). Defaults preserve the existing `get_edge_ids` wording, so existing snapshots are unchanged.
- Each entry point gates on `is.data.frame(x) || inherits(x, "matrix")` before normalizing, so vectors and `igraph.vs`/`igraph.es` sequences keep their existing behavior.
- `delete_edges()` converts a matrix/df of vertex pairs to edge IDs via `get_edge_ids(..., error = TRUE)` (its `edges` arg is an *edge selector*, not a vertex sequence), preserving numeric-edge-id and `"a|b"` semantics for vectors.
- **Internal callers** that relied on passing 2×n matrices are updated to pass explicit flat vectors (`c(...)`): `graph_from_edgelist()`, `graph_from_data_frame()`, the incidence/adjacency constructors, and `[<-` indexing. (`graph_from_data_frame()` was in fact already broken by the `add_edges` change and is fixed here.)

## Tests

Added matrix/data-frame acceptance, vector/`igraph.vs` back-compat, 2×n/2×2 rejection, and `as_edgelist()` round-trip (#550) cases across `test-interface.R`, `test-operators.R`, `test-make.R`, and `test-iterators.R`. Updated `test-conversion.R` to use the now-natural n×2 form.

Full suite: **1553 test groups, 0 failures, 0 errors**. No snapshot changes.


R/adjacency.R
```diff
@@ -542,7 +542,7 @@ graph.adjacency.sparse <- function(
     res <- make_empty_graph(n = vc, directed = (mode == "directed"))
     weight <- list(el[, 3])
     names(weight) <- weighted
-    res <- add_edges(res, edges = t(as.matrix(el[, 1:2])), attr = weight)
+    res <- add_edges(res, edges = c(t(as.matrix(el[, 1:2]))), attr = weight)
```
krlmlr, line 545: Would that work?


```suggestion
    res <- add_edges(res, edges = as.matrix(el[, 1:2]), attr = weight)
```

R/conversion.R
```diff
@@ -1748,7 +1748,7 @@ graph_from_data_frame <- function(d, directed = TRUE, vertices = NULL) {
   }
 
   # add the edges
-  g <- add_edges(g, edges, attr = attrs)
+  g <- add_edges(g, c(edges), attr = attrs)
```
krlmlr, line 1751: Add a comment, or:

```suggestion
  g <- add_edges(g, row_major_to_vertex_pairs(edges), attr = attrs)
```

with

`row_major_to_vertex_pairs <- c`

or

```r
row_major_to_vertex_pairs <- function(x) c(x)
```

R/conversion.R
```diff
@@ -1800,7 +1800,7 @@ graph_from_edgelist <- function(el, directed = TRUE) {
       V(res)$name <- names
     } else {
       ## normal edge list
-      res <- make_graph(t(el), directed = directed)
+      res <- make_graph(c(t(el)), directed = directed)
```
krlmlr, line 1803:
```suggestion
      res <- make_graph(el, directed = directed)
```

R/indexing.R
```diff
@@ -521,7 +521,7 @@ expand.grid.unordered <- function(i, j, loops = FALSE, directed = FALSE) {
       ## Addition or update of an attribute (or both)
       ids <- get_edge_ids(x, c(rbind(from, to)))
       if (any(ids == 0)) {
-        x <- add_edges(x, rbind(from[ids == 0], to[ids == 0]))
+        x <- add_edges(x, c(rbind(from[ids == 0], to[ids == 0])))
```
krlmlr, line 524:
```suggestion
        x <- add_edges(x, to_vertex_pairs(from[ids == 0], to[ids == 0]))
```

krlmlr (commented):
Thanks, nice! Now is the time to do it. Should we do revdepchecks before merging?

krlmlr:
Now we can revdepcheck if needed, with the revdep4 workflow.

krlmlr:
After seeing the results, we really have to _want_ that. It will also postpone submission by at least a month. No problem, just to be aware.

schochastics:
Yes thats a lot. Lets keep it open for now. I will look at the issues. This feature should not block the release though

krlmlr:
Should we do a loud warning first? What does that do to the revdeps?

krlmlr:
Let's also split out the internal parts (that only affect us and that we should do either way) from the external flip. Stacked PRs.

Issue #1056, opened 2023-12-30
Clarify error handling


krlmlr:
https://github.com/igraph/rigraph/pull/1051#issuecomment-1872573532

maelle:
@schochastics I wonder whether this belongs to the ISC milestone, is it related to clarifying what's an R vs C responsability?

szhorvat:
This is about ensuring that the glue code in the R interface plays nice with the error handling of the C core. In particular, longjumps MUST NOT happen except at the top level. Rf_error does a longjump by default.

Basically this boils down to writing down some rules about when we use IGRAPH_CHECK vs IGRAPH_R_CHECK, and ensuring that we're following these rules. In other words, decide what function is considered top level and what isn't, and make sure that top-level functions are never called from non-top-level ones. Top-level uses IGRAPH_R_CHECK, everything else uses IGRAPH_CHECK.

Related: #514

Currently, the error handling is a bit of a mess, and when certain errors happen, the R interface WILL blow up in some way, or have its internal state corrupted. The reason why we don't see this is that most such errors are unlikely (such as out-of-memory).

PR #2708, opened 2026-06-23
chore: Separate and annotate top-level vs helper functions in `rinterface_extra.c`

### LINKED ISSUE #2105: Separate top-level vs helper functions in rinterface_extra.c
`rinterface_extra.c` contains two kinds of functions: 

 - Top-level: Those that will eventually be called through `.Call` from R, and use `IGRAPH_R_CHECK()`
 - Helper: Those that are called by other functions, including top-level functions, and use `IGRAPH_CHECK()`

These are mixed together in the same file and are not annotated. I am not convinced that the use of `IGRAPH_R_CHECK` vs `IGRAPH_CHECK` is not incorrect somewhere. Even if this is not incorrect now, it is very easy to make a mistake since it's unclear which functions are top-level.

I suggest making a clean separation between these, either placing them in separate files, or separate sections of the same file. 

It would be even better to use different naming schemes. One might think that all `R_` functions are top-level, but it is not so.

schochastics:
Closes #2105. Biggest step toward #1056.

## Background

`src/rinterface_extra.c` mixes two functionally distinct kinds of functions with no annotation:

- **TOP-LEVEL** — called from R via `.Call`, registered in the `CallEntries[]` table in the generated `src/cpp11.cpp`, and must use `IGRAPH_R_CHECK()` (which longjmps on error and never returns an error code).
- **HELPER** — called only from other C code; must use `IGRAPH_CHECK()` and return their `igraph_error_t` so the caller can propagate it.

Because the categories were unmarked and the `Rx_`/`Ry_`/`Rz_`/`Rw_` name prefixes don't map to them, it was easy to use the wrong error macro — exactly the worry raised in #2105.

### 1. Annotate every function + add a header doc block
Each of the 258 function definitions now carries a greppable marker — **87** `/* TOP-LEVEL ... */` and **171** `/* HELPER ... */`. The TOP-LEVEL set is derived authoritatively from the `cpp11.cpp` `CallEntries` registration table, **not** from name prefixes. A new file-header block defines the two kinds and the `IGRAPH_R_CHECK` / `IGRAPH_CHECK` rule, and notes that the historical prefixes do not indicate the kind.

### 2. Remove 38 dead functions (~1000 lines) — please review carefully
This is the bulk of the diff (1037 deletions). These are **old hand-written top-level entry points that were long ago superseded by stimulus-generated `R_igraph_*` versions**:

`Rx_igraph_create`, `Rx_igraph_star`, `Rx_igraph_ring`, `Rx_igraph_kary_tree`, `Rx_igraph_full`, `Rx_igraph_random_sample`, `Rx_igraph_get_adjacency`, `Rx_igraph_get_edge`, `Rx_igraph_decompose`, `Rx_igraph_get_eids`, `Rx_igraph_graphlets`, `Rx_igraph_community_to_membership`, `Rx_igraph_community_leading_eigenvector`, `Rx_igraph_adjacency_spectral_embedding`, `Rx_igraph_laplacian_spectral_embedding`; the whole `*_game` family (`barabasi`, `barabasi_aging`, `recent_degree_aging`, `degree_sequence`, `callaway_traits`, `establishment`, `watts_strogatz`, `lastcit`, `cited_type`, `citing_cited_type`); the `st_*` connectivity / disjoint-paths wrappers (`st_vertex_connectivity`, `st_edge_connectivity`, `st_mincut_value`, `edge_disjoint_paths`, `vertex_disjoint_paths`, `connect_neighborhood`); and a handful of orphaned helpers (`Rx_igraph_get_n`, `Rx_igraph_get_directed`, `Rx_igraph_mybracket`, `Rx_igraph_check_finally_stack`, `Rx_igraph_vectorlist_int_to_SEXP`, `Rx_igraph_i_levc_arp`, plus the cascade-orphaned `Rx_igraph_i_levc_callback` and its two typedefs).

Why they're safe to remove:
- **Not registered** in `cpp11.cpp` (`CallEntries`), so not reachable via `.Call`.
- **Not referenced anywhere** — verified across `src/`, `R/`, `tools/`, and the yaml/templates.
- A **generated `R_igraph_*` replacement exists** for each (the R wrappers in `R/aaa-*.R` call the generated versions).

They still used `IGRAPH_R_CHECK` and *looked* top-level, which is precisely what made the file confusing. Removal was done programmatically and iterated to a fixpoint so cascade-orphaned helpers were caught too.

### 3. Fix the one genuine macro inconsistency
`Rz_SEXP_to_graph_ptr_list` is a helper (returns `igraph_error_t`, called from generated wrappers in `rinterface.c`) but used `IGRAPH_R_CHECK`. Switched to `IGRAPH_CHECK`; the callers already wrap it in `IGRAPH_R_CHECK`, so error semantics are preserved. The only remaining helper that uses `IGRAPH_R_CHECK` is `Rx_igraph_restore_pointer`, which returns `void` and therefore cannot propagate an error code — its marker documents why that's intentional.

## Verification

- Package **compiles and links cleanly** (`devtools::load_all`).
- Audit confirms **no top-level function uses plain `IGRAPH_CHECK`**, and **only the one documented `void`-returning helper uses `IGRAPH_R_CHECK`**.
- **Test suite**: `8173 PASS`. The handful of failures are pre-existing flaky/unseeded random-graph tests (`test-games.R` bipartite print assertions, `cluster_leading_eigen is deterministic`) — confirmed to fail/flake identically on the unmodified baseline, so they're unrelated to this change.
- **Functional spot-checks** (seeded) pass for graph-list operations (union/intersection/decompose — these exercise the modified `Rz_SEXP_to_graph_ptr_list`) and for every family whose dead duplicate was removed (PA/Barábasi, star, degseq, graphlets, spectral embedding, leading-eigenvector community, max-flow).

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -69,9 +69,35 @@ enum igraph_versions {
 
 #define R_IGRAPH_VERSION_VAR ".__igraph_version__."
 
+/******************************************************************************
+ * TWO KINDS OF FUNCTIONS IN THIS FILE
+ *
+ * This file mixes two functionally distinct kinds of functions. Every function
+ * definition below is annotated with a one-line marker saying which it is:
+ *
+ *   TOP-LEVEL  - called from R via `.Call`. Registered in the CallEntries[]
+ *                table in the generated src/cpp11.cpp. These run at the
+ *                boundary where an R error may be raised, so they MUST use
+ *                IGRAPH_R_CHECK() (which longjmps on failure and never returns
+ *                an error code). They take and return SEXP.
+ *
+ *   HELPER     - called only from other C code (including from top-level
+ *                functions). They MUST use IGRAPH_CHECK() and return their
+ *                igraph_error_t so the caller can propagate it; they MUST NOT
+ *                use IGRAPH_R_CHECK(). (A few helpers that cannot return an
+ *                error code, e.g. they return void or a pointer, deliberately
+ *                use IGRAPH_R_CHECK() and are only ever reached from top-level
+ *                contexts - their marker notes this.)
+ *
+ * Grep for "TOP-LEVEL" or "HELPER" to list each kind. The historical
+ * Rx_/Ry_/Rz_/Rw_ name prefixes do NOT indicate the kind - the marker does.
+ * Keeping the markers correct is the point of issue #2105 / #1056.
+ *****************************************************************************/
+
 /* The following three R_check_... functions must only be called from top-level C code,
  * i.e. in contexts where igraph_error() does NOT return. */
 
+/* HELPER: internal C; must use IGRAPH_CHECK */
 void R_check_int_scalar(SEXP value)
 {
   if (Rf_xlength(value) != 1) {
@@ -84,6 +110,7 @@ void R_check_int_scalar(SEXP value)
   }
 }
 
+/* HELPER: internal C; must use IGRAPH_CHECK */
 void R_check_real_scalar(SEXP value)
 {
   if (Rf_xlength(value) != 1) {
@@ -92,6 +119,7 @@ void R_check_real_scalar(SEXP value)
   }
 }
 
+/* HELPER: internal C; must use IGRAPH_CHECK */
 void R_check_bool_scalar(SEXP value)
 {
   if (Rf_xlength(value) != 1) {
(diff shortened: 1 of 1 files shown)


schochastics:
@krlmlr any todos left here? I think the split into files should be a followup, right?

krlmlr:
Thanks! Yes, please split as a follow-up. Need to declare the new C file in `Makeflags` or similar.

PR #2714, opened 2026-06-25
test: withr-free local seed and RNG-state leak fixes

### LINKED ISSUE #2713: local_seeds
Try to write a robust `igraph_local_seed` function

schochastics:
Fix #2713

## What & why

The test suite has an opt-in check (gated behind `IGRAPH_CHECK_RNG_STATE=true` in
`tests/testthat/setup.R`) that fails any test which leaves the global
`.Random.seed` changed. This PR makes the whole suite pass that check and moves
the suite's RNG-seed handling off `withr` onto in-house helpers.

Running the suite with `IGRAPH_CHECK_RNG_STATE=true`: **0 seed leaks, 0
failures, 0 warnings** (9739 passing).

## New helpers (`tests/testthat/helper.R`)

**`igraph_local_seed()`** — drop-in for `withr::local_seed()`. Sets the seed for
the current scope and restores the previous global RNG state when that scope
exits.
- Self-contained: snapshots and restores `RNGkind` + `.Random.seed` itself
  (uses `withr::defer` only as the deferral primitive), so it doesn't inherit
  `withr::local_seed`'s past seed-handling bugs.
- Restores **RNGkind first, then `.Random.seed` last** — changing the kind
  discards the seed, so the seed write must come last for an exact round-trip.
- Optional `rng_version` (e.g. `"3.5.0"`), folding in what used to be a separate
  `local_rng_version()` call. Pairing two independent deferred restores (kind +
  seed) ran them in the wrong order and leaked; a single ordered restore fixes it.

**`igraph_with_seed()`** — block form mirroring `withr::with_seed()`: set the
seed, evaluate the block, restore. Needed where a single test has several
independently seeded blocks with code between them (e.g. `test-print-classic.R`),
which a frame-scoped `igraph_local_seed()` can't express.

## setup.R

- Set a baseline seed (`set.seed(42)`) so `.Random.seed` always exists before
  tests run. Otherwise the first RNG-touching test *creates* it from nothing
  (many igraph C functions initialise the RNG without drawing from it, e.g.
  `voronoi_cells()`), which the check flagged — and restoring that absent state
  just shoved the "first creation" onto the next test (whack-a-mole).
- The seed-leak override now forwards the test block via
  `substitute()` + `eval(as.call(...))` instead of
  `rlang::inject(testthat::test_that(name, !!code))`. `inject()` eagerly
  processed **any** injection operator in the test body (`!!`, `!!!`, `{{`),
  e.g. `set_vertex_attrs(g, !!!attr_list)`, evaluating it in the wrong frame and
  erroring with `object 'attr_list' not found`.
- The check stays gated behind `IGRAPH_CHECK_RNG_STATE=true` (opt-in for CI).

## Test files

- Moved helper functions that were defined inside test files into a shared
  `helper-test-functions.R` (the override re-scopes test code, so file-local
  helpers were no longer reachable from the test body).
- Added `igraph_local_seed()` to tests that genuinely advance the RNG (unseeded
  `sample_*` fixtures and ARPACK eigensolvers).
- Replaced **all 875** `withr::local_seed()` call sites with
  `igraph_local_seed()` (including the 732 in `test-aaa-auto.R`).
- Replaced the **9** `withr::with_seed()` call sites with `igraph_with_seed()`
  (`test-print-classic.R`, `test-layout.R`, `test-iterators.R`). These never
  leaked — consistency cleanup; snapshots unchanged.
- Replaced every `local_rng_version("3.5.0")` + seed pairing with
  `igraph_local_seed(seed, rng_version = "3.5.0")`, then removed the now-unused
  `local_rng_version()` helper.
- `restore_rng_state()` suppresses warnings around the `RNGkind` restore so
  putting back the older `"Rounding"` sampler (`RNGversion("3.5.0")`) doesn't
  emit R's "non-uniform 'Rounding' sampler used" note.

## tools/AGENTS.md

Updated the test-writing guidance to use `igraph_local_seed`, so newly added
`test-aaa-auto.R` tests follow the convention.

## Notes
- `igraph_local_seed()`/`igraph_with_seed()` live in `helper.R` (not `setup.R`)
  so helper functions like `make_scan_graphs()` can see them.
- The suite no longer uses `withr` for any RNG-seed handling.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -6,13 +6,6 @@ skip_if_no_graphml <- function() {
   if (!has_graphml()) skip("No GraphML support")
 }
 
-local_rng_version <- function(version, .local_envir = parent.frame()) {
-  orig <- RNGkind()
-  withr::defer(do.call(RNGkind, as.list(orig)), envir = .local_envir)
-  suppressWarnings(RNGversion(version))
-  orig
-}
-
 # Restore the global RNG state (kind + .Random.seed) to a previously captured
 # snapshot. `kind` is the result of RNGkind() and `state` is .Random.seed before
 # the seed was set (or NULL if no RNG had been used yet). The kind is restored
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -6,48 +6,43 @@ skip_if_no_graphml <- function() {
   if (!has_graphml()) skip("No GraphML support")
 }
 
-# Restore the global RNG state (kind + .Random.seed) to a previously captured
-# snapshot. `kind` is the result of RNGkind() and `state` is .Random.seed before
-# the seed was set (or NULL if no RNG had been used yet). The kind is restored
-# first -- changing the RNG kind discards .Random.seed -- so that the seed
-# assignment is the final word and the state round-trips exactly.
-restore_rng_state <- function(kind, state) {
-  # suppressWarnings: restoring an older sampler (e.g. the "Rounding"
-  # sample.kind that RNGversion("3.5.0") selects) otherwise emits R's
-  # "non-uniform 'Rounding' sampler used" note. We're only putting back a state
-  # that was already in effect, so the note is noise.
-  suppressWarnings(do.call(RNGkind, as.list(kind)))
-  if (is.null(state)) {
-    if (exists(".Random.seed", envir = globalenv(), inherits = FALSE)) {
-      rm(".Random.seed", envir = globalenv())
-    }
-  } else {
-    assign(".Random.seed", state, envir = globalenv())
-  }
-}
-
-# Drop-in replacement for withr::local_seed() in tests. Sets the RNG seed (and
-# optionally pins the RNG version via `rng_version`, e.g. "3.5.0") and restores
-# the pre-call global RNG state when `.local_envir` exits (the current test_that
-# block, or the enclosing helper function when called from one). Self-contained:
-# it does its own save/restore, so it does not depend on withr's local_seed().
-# withr::defer is used only as the deferral primitive (it fires reliably in both
-# test blocks and function frames). Folding the version pin in here -- rather
-# than pairing a separate RNG-version helper with its own deferred restore --
-# keeps a single, correctly ordered restore point, so the seed never leaks.
+# Drop-in replacement for withr::local_seed() in tests.
+# Sets the RNG seed (and optionally pins the RNG version via `rng_version`,
+# e.g. "3.5.0") and restores the pre-call global RNG state when `.local_envir`
+# exits (the current test_that block, or the enclosing helper function when
+# called from one).
+#
+# Self-contained: it does its own save/restore, so it does not depend on withr's
+# local_seed() (whose RNG-version/kind restore has had ordering regressions, see
+# https://github.com/r-lib/withr/issues/167).
+# withr::defer is used only as the deferral primitive; it fires reliably in both
+# test blocks and function frames.
+#
+# The RNG version is pinned here rather than via a separate helper.
+# Pairing a separate RNG-version helper (with its own deferred restore) alongside
+# the seed restore ran the two restores in the wrong order and leaked the seed --
+# e.g. restoring the kind after the seed discards the seed again.
+# Folding the version pin into this one, correctly ordered restore point avoids
+# that, so the seed never leaks.
 igraph_local_seed <- function(
   seed,
   rng_version = NULL,
+  # kind/normal.kind/sample.kind mirror set.seed()'s arguments of the same names.
   kind = NULL,
   normal.kind = NULL,
   sample.kind = NULL,
   .local_envir = parent.frame()
 ) {
-  old_kind <- RNGkind()
-  old_seed <- get0(".Random.seed", envir = globalenv(), inherits = FALSE)
-  withr::defer(restore_rng_state(old_kind, old_seed), envir = .local_envir)
+  initial_kind <- RNGkind()
+  initial_seed <- get0(".Random.seed", envir = globalenv(), inherits = FALSE)
+  withr::defer(
+    restore_rng_state(initial_kind, initial_seed),
+    envir = .local_envir
+  )
 
   if (!is.null(rng_version)) {
+    # suppressWarnings: same "non-uniform 'Rounding' sampler used" note as in
+    # restore_rng_state() -- RNGversion("3.5.0") selects that older sampler.
     suppressWarnings(RNGversion(rng_version))
   }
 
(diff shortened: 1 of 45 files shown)


tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
   if (!has_graphml()) skip("No GraphML support")
 }
 
-local_rng_version <- function(version, .local_envir = parent.frame()) {
-  orig <- RNGkind()
-  withr::defer(do.call(RNGkind, as.list(orig)), envir = .local_envir)
-  suppressWarnings(RNGversion(version))
-  orig
+# Restore the global RNG state (kind + .Random.seed) to a previously captured
```
maelle: Could the line breaks be after sentences? 💅

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+# snapshot. `kind` is the result of RNGkind() and `state` is .Random.seed before
+# the seed was set (or NULL if no RNG had been used yet). The kind is restored
+# first -- changing the RNG kind discards .Random.seed -- so that the seed
+# assignment is the final word and the state round-trips exactly.
+restore_rng_state <- function(kind, state) {
+  # suppressWarnings: restoring an older sampler (e.g. the "Rounding"
+  # sample.kind that RNGversion("3.5.0") selects) otherwise emits R's
+  # "non-uniform 'Rounding' sampler used" note. We're only putting back a state
+  # that was already in effect, so the note is noise.
+  suppressWarnings(do.call(RNGkind, as.list(kind)))
```
maelle: how I wish we could suppress that warning exactly (with a class?) and remove the comment.

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+  if (is.null(state)) {
+    if (exists(".Random.seed", envir = globalenv(), inherits = FALSE)) {
+      rm(".Random.seed", envir = globalenv())
+    }
+  } else {
+    assign(".Random.seed", state, envir = globalenv())
+  }
+}
+
+# Drop-in replacement for withr::local_seed() in tests. Sets the RNG seed (and
```
maelle: put this function first, as it's the higher level one. (sort of https://style.tidyverse.org/package-files.html#organisation) 

For instance when reviewing, it'd make more sense to see this function first.

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+# optionally pins the RNG version via `rng_version`, e.g. "3.5.0") and restores
+# the pre-call global RNG state when `.local_envir` exits (the current test_that
+# block, or the enclosing helper function when called from one). Self-contained:
+# it does its own save/restore, so it does not depend on withr's local_seed().
```
maelle: link to some withr issue to explain our choice?

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+# withr::defer is used only as the deferral primitive (it fires reliably in both
+# test blocks and function frames). Folding the version pin in here -- rather
```
maelle: I must say this is a complicated sentence for me. Maybe rephrase in two sentences with an example for each?

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+# than pairing a separate RNG-version helper with its own deferred restore --
+# keeps a single, correctly ordered restore point, so the seed never leaks.
+igraph_local_seed <- function(
+  seed,
+  rng_version = NULL,
+  kind = NULL,
+  normal.kind = NULL,
```
maelle, line 32: why dots?

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+  sample.kind = NULL,
+  .local_envir = parent.frame()
+) {
+  old_kind <- RNGkind()
```
maelle: instead of old should we use "initial" or "starting"?

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+  old_seed <- get0(".Random.seed", envir = globalenv(), inherits = FALSE)
+  withr::defer(restore_rng_state(old_kind, old_seed), envir = .local_envir)
+
+  if (!is.null(rng_version)) {
+    suppressWarnings(RNGversion(rng_version))
```
maelle, line 46: here the warning is suppressed without any explanation

tests/testthat/helper.R
```diff
@@ -6,11 +6,70 @@ skip_if_no_graphml <- function() {
...
+  }
+
+  set.seed(
+    seed,
+    kind = kind,
+    normal.kind = normal.kind,
+    sample.kind = sample.kind
+  )
+
+  invisible(old_seed)
```
maelle: why? convention in withr?

tests/testthat/helper.R
maelle: because they're `set.seed()` arguments I suppose 😅

tests/testthat/setup.R
```diff
@@ -5,12 +5,35 @@
 # is pinned off because the 7-character graph id is non-deterministic.
 igraph_options(print.style = "classic", print.id = FALSE)
 
+# Establish a baseline global RNG state so `.Random.seed` always exists before
+# any test runs. Without this, the first test to touch the RNG would *create*
+# `.Random.seed` from nothing (many igraph C functions initialise the RNG even
```
maelle, line 10: is that bad design in those functions?

tests/testthat/setup.R
```diff
@@ -5,12 +5,35 @@
...
+# when they don't draw from it, e.g. voronoi_cells()), which the seed-leak check
+# below would flag. Restoring that absent state (as igraph_local_seed() faithfully
+# does when `before` was NULL) just pushes the creation onto the next test -- a
+# whack-a-mole. Pinning a baseline here means only tests that genuinely *advance*
+# the seed need igraph_local_seed().
+set.seed(42)
```
maelle, line 16: and this seed never leaks in an user's interactive session?

tests/testthat/setup.R
```diff
@@ -5,12 +5,35 @@
...
+
+# `igraph_local_seed()` (and its `restore_rng_state()` helper) live in helper.R
+# so they are visible both to test code and to helper functions that call them.
+
 if (Sys.getenv("IGRAPH_CHECK_RNG_STATE") == "true") {
   # Check that tests don't change the random seed
```
maelle: I don't understand why we need this if we're confident in our solution. 

And when will we set this env var? In a GHA workflow? Before each release?

tests/testthat/setup.R
```diff
@@ -19,3 +42,15 @@ if (Sys.getenv("IGRAPH_CHECK_RNG_STATE") == "true") {
     )
   }
 }
+
+# NOTE (why the override forwards `code` via substitute()/eval() instead of
```
maelle: Merge this NOTE and the comments before?

maelle (commented):
Thank you!! :game_die:

tests/testthat/setup.R
schochastics: There will be a specific runner that has this set to true (at least thats what Kirill was planing

tests/testthat/setup.R
schochastics: No this doesn leak. Test run in there own R throwaway R process

tests/testthat/setup.R
maelle: add this in a comment?

tests/testthat/helper.R
schochastics: R's "non-uniform 'Rounding' sampler used" note is a plain simpleWarning with no dedicated condition class, so there's nothing to catch by class — we'd have to match on the message string, which is fragile across locales and R versions. suppressWarnings around the exact one-line call is the narrowest reliable option, so I've kept it and left the explanatory comment. (This is the same regression withr itself worked around in [r-lib/withr#167](https://github.com/r-lib/withr/issues/167).)

tests/testthat/setup.R
schochastics: It's the igraph C core seeding its RNG on entry, not a per-function quirk. Several C routines initialise the RNG even when a particular call doesn't end up drawing from it (e.g. voronoi_cells()), so it's expected behaviour rather than something to fix in those wrappers.

tests/testthat/helper.R
```diff
@@ -6,11 +6,89 @@ skip_if_no_graphml <- function() {
...
+  }
+
+  set.seed(
+    seed,
+    kind = kind,
+    normal.kind = normal.kind,
+    sample.kind = sample.kind
+  )
+
+  # Return the previous seed invisibly, mirroring withr::local_seed()'s contract.
```
krlmlr:
```suggestion
  # Return the previous seed invisibly, mirroring withr::local_dir()'s
  # and withr::local_locale()'s contract.
  # Interestingly, withr::local_seed() returns the new seed.
```

tests/testthat/setup.R
krlmlr: Can be removed now entirely?

krlmlr (commented):
Thanks, please merge when green.

tests/testthat/setup.R
schochastics: The baseline exists for a different class of tests: ones that are not wrapped (because they aren't "random" tests) but call an igraph C function that defensively initialises R's RNG via `GetRNGstate`/`PutRNGstate` without ever drawing. Those functions:

create `.Random.seed` from nothing when none exists → the leak check flags it, but
leave an existing `.Random.seed` unchanged.
`set.seed(42)` guarantees `.Random.seed` always exists, so those init-only calls read-and-write-back the same value and don't get flagged — which means such tests don't need a pointless `igraph_local_seed()` wrapper.

PR #2732, opened 2026-07-06
chore: stop naming mode in indexing.R

maelle:
One of those was unused. It seemed clearer not to name any?


maelle:
Hide white space in the diff otherwise it's not helpful :smiley_cat:

schochastics (approved):

PR #2676, opened 2026-05-28
feat: Add attribute combination support to graph operators

schochastics:
fix #57

PR #2742, opened 2026-07-08
refactor: snake_case attr-comb options & arguments, soft-deprecate dotted names

schochastics:
Split out from #2676 per [review feedback](https://github.com/igraph/rigraph/pull/2676#pullrequestreview-4569626651) ("I'd like to review the *par* code in a separate PR"). #2676 kept the original **dotted** option/argument names; this PR carries the naming refactor on its own so it can be reviewed in isolation.

## What this does

- **igraph options**: rename `vertex.attr.comb` → `vertex_attr_combine`, `edge.attr.comb` → `edge_attr_combine`, and the new `graph.attr.comb` → `graph_attr_combine`. Reading/setting the old dotted keys still works via back-compatible aliasing in `igraph_opt()` / `igraph_options()`, emitting a single `lifecycle::deprecate_soft()` warning (`R/par.R`).
- **Function arguments**: `simplify()`, `as_undirected()` and `contract()` gain snake_case `edge_attr_combine` / `vertex_attr_combine` arguments via the argument-migration infrastructure (registry entries in `tools/migrations/attr-comb.R`, spliced by `Rscript tools/generate-migrations.R`; schema in `tools/migrations/README.md`); the dotted names keep working and soft-deprecate.
- **Graph operators**: `union()` / `intersection()` / `compose()` / `disjoint_union()` use the snake_case `*_combine` argument names.
- Updates the stimulus codegen defaults (`types-RR.yaml`), regenerated `*_impl` defaults, and the affected tests/snapshots.
- Points the stale `tools/migrations.R` schema references in sibling registry files and helper comments at `tools/migrations/README.md`, where the schema documentation lives since the registry split (#2779).

## Base branch

#2676 has merged, so this PR is based on `main` and rebased into a single commit on top of it; the attr-comb migration entries live in their own registry file `tools/migrations/attr-comb.R` so that the open per-topic migration PRs can append to `conversion.R`, `community.R`, etc. without conflicting here.


tests/testthat/_snaps/conversion.md
```diff
@@ -17,6 +17,13 @@
       Warning:
       `as.undirected()` was deprecated in igraph 2.1.0.
       i Please use `as_undirected()` instead.
+      Warning:
+      The igraph option `edge.attr.comb` was deprecated in igraph 3.0.0.
+      i Please use the `edge_attr_combine` option instead.
+      Warning:
+      Calling `as_undirected()` with positional or abbreviated arguments was deprecated in igraph 3.0.0.
+      i Detected call: as_undirected(graph, mode, edge.attr.comb)
+      i Use instead: as_undirected(graph, mode, edge_attr_combine = )
```
krlmlr, line 26: This doesn't look quite right, is this our code that triggers this deprecation message?

krlmlr (commented):
Thanks! Almost good to merge. The one snapshot flagged could be just a symptom of a pattern across the codebase.

krlmlr:
Needs conflict resolution.

krlmlr:
A clean session, checking if this PR introduced `...` to all signatures before the arg rename where it wasn't there before, would be worthwhile. Thanks for the effort!

PR #2745, opened 2026-07-11
fix: write GML without explicit vertex IDs

### LINKED ISSUE #2710: write_graph(g, format = "gml") fails because NULL id coerced to numeric(0)

write_graph(g, format = "gml") fails for any graph without an `id` attribute, raising:
```
Error in write_graph_gml_impl(...): 
  Size of id vector must match vertex count. Invalid value
Source: io/gml.c:1057
```
**Expected behavior**
With no id argument, the GML writer should generate default node ids and write the file (as it did in an earlier igraph version).

**Actual behavior**
It always errors at gml.c:1057.

**Apparent root cause**
The high-level write.graph.gml() guards its coercion:
```
if (!is.null(id)) {
  id <- as.numeric(id)
}
```
so a NULL id is passed through to the auto-generated `write_graph_gml_impl()`. That function, however, coerces unconditionally:
```id <- as.numeric(id)```
and as.numeric(NULL) returns numeric(0), not NULL. The C routine therefore receives a non-NULL, length-0 id vector instead of "no id supplied." In gml.c, myid becomes non-NULL with size 0, so the if (myid) branch runs and the size check (0 != vcount) fails. This is consistent with the error firing before any vertex data is examined, and with an explicit correctly-sized id avoiding it.

A possible fix would be to guard the coercion in `write_graph_gml_impl()` the same way `write.graph.gml()` does `(if (!is.null(id)) id <- as.numeric(id))`, so a NULL id reaches C as a null pointer.

**Workaround**
Explicitly specify ids:
```write_graph(g, file, format = "gml", id = seq_len(vcount(g)))```

### To reproduce

```
library(igraph)
write_graph(make_ring(5), tempfile(fileext = ".gml"), "gml")
```
> Error in write_graph_gml_impl(graph = graph, outstream = file, options = "default",  : 
  Size of id vector must match vertex count. Invalid value
Source: io/gml.c:1057

Supplying an explicit id works:
```
g <- make_ring(5)
write_graph(g, tempfile(fileext = ".gml"), "gml", id = seq_len(vcount(g)))   # no error
```

### System information

> sessionInfo()
R version 4.6.0 (2026-04-24)
Platform: aarch64-apple-darwin23
Running under: macOS Tahoe 26.5.1

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.6/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/Detroit
tzcode source: internal

attached base packages:

other attached packages:
[1] igraph_2.3.2

[1] compiler_4.6.0  magrittr_2.0.5  cli_3.6.6       tools_4.6.0     lifecycle_1.0.5
[6] pkgconfig_2.0.3 rlang_1.2.0

LeonidasZhak:
## Summary

Allow `write_graph(..., format = "gml")` to use the GML writer's default vertex IDs when `id` is omitted. Fixes #2710.

## Thanks to maintainers

Thanks for triaging the report, requesting a regression test, refreshing the branch, and re-running the checks.

## Issue or motivation

Writing a graph as GML currently fails unless the caller supplies one ID per vertex. Earlier versions generated node IDs when none were supplied, and the C writer still supports that path.

## Root cause

Stimulus treated the C writer's nullable `id` pointer as a required vector. The generated R wrapper converted `NULL` to `numeric(0)`, so C received an empty vector instead of a null pointer and rejected its length.

## Change

Mark `id` as optional in the Stimulus metadata and regenerate the R/C interface layers. The added regression test writes a graph without IDs and reads it back to confirm graph isomorphism.

The branch is now merged through `igraph/rigraph@a52ba87` and remains mergeable against the current `main`; the PR diff is still limited to the GML interface metadata, generated wrapper/C interface, and the focused regression test.

## Tests

- `air format --check .`
- Regenerated the Stimulus interface targets; the generated GML wrapper files remained reproducible.
- Focused `test-foreign.R`: 18 passed, 0 failed, 0 warnings.
- Full `testthat::test_local(reporter = "check", stop_on_failure = TRUE)`: 9208 passed, 0 failed, 3 skipped.
- `R CMD build`: succeeded, including vignette creation.
- `R CMD check --no-manual`: package tests and vignette rebuild passed; local status was `2 WARNINGs` from vendored GLPK/compiler diagnostics, with no errors or notes.
- GitHub Actions `rcc` run 33739775688, attempt 2: succeeded. Its jobs included the stock-R smoke test, matrix collection, pkgdown build, and R CMD check jobs on macOS, Windows, Windows ARM, Ubuntu, and Ubuntu ARM for R 4.2 through devel. The separate benchmark receive workflow was skipped.

## Scope

This does not change explicit IDs, other graph writers, GML serialization details, vendored GLPK, unrelated lifecycle/CI behavior, or the igraph C library.

I used AI assistance to help check for issues, refine the code, and run tests.


schochastics:
LGTM if CI is green

krlmlr:
Working on fixing checks on main, let's merge main into this branch when done to get it green.

LeonidasZhak:
## Summary

Merged the current `main` branch into this PR branch as requested. The GML fix itself is unchanged.

## Thanks to maintainers

Thanks for reviewing the fix and for the guidance on when to refresh the branch.

## Issue or motivation

The branch was seven commits behind `main`, including the recent package-cleanup changes, so its CI result no longer represented the current repository state.

## Root cause

The previous PR checks ran against the older base at `acd63244`. The branch needed the current `main` history before CI could reassess the patch.

## Change

Merged `igraph/rigraph@e772e9e7` without conflicts. Relative to current `main`, the PR still changes only the four original GML interface/test files.

## Tests

- `air format .`
- focused `test-foreign.R`: four non-skipped tests passed; six CRAN-gated tests skipped
- `NOT_CRAN=true testthat::test_local(..., stop_on_failure = TRUE)`
- full vignette build with `R CMD build`
- `R CMD check --no-manual`: tests and vignette rebuild passed; two existing vendored-GLPK warnings and one local `rgl`/X11 note remain

## Scope

This update only syncs the PR with `main`. It does not alter the GML patch, workflows, vendored GLPK, or unrelated current-`main` CI failures.

LeonidasZhak:
Quick follow-up on the CI gate: the PR is still mergeable and the GML patch is unchanged after the requested `main` sync. The `rcc` and `Continuous Benchmarks (Receive)` runs both currently show `action_required`, with no jobs or test logs created.

Could a maintainer approve or re-run those workflows when convenient? I’m happy to refresh the branch again if the current base requires it.

PR #2783, opened 2026-07-26
fix(migrations): detect supplied tail args with `missing()`

krlmlr:
The argument-coverage test sweep surfaced a real defect class in the
recovery machinery — found independently twice (games and layout agents),
and one instance is live on `main`.

## The bug

The generated ARG_HANDLE blocks passed `current = list(<tail args>)` plus a
re-evaluated `defaults` list into `migrate_recover_args()`, which detected
"argument supplied twice" by comparing values. That **forces every tail
promise before recovery** and **re-evaluates defaults**, which breaks three
ways:

1. **Cross-referencing defaults freeze.** `sample_last_cit()`'s
   `pref = (1:(agebins + 1))^-3` memoizes against the pre-recovery `agebins`,
   so the legacy positional call `sample_last_cit(20, 2, 5)` errors
   (on the open games PR).
2. **Irreproducible defaults are misread as conflicts.** Defaults like
   `V(graph)[1]`, `matrix(runif(...))` are never `identical()` across
   evaluations, so recovering such a slot errors "supplied more than once" —
   `diversity(g, w, vids)` is affected **on `main` today**, and
   `layout_as_star(g, 3)` on the layout PR.
3. **RNG state is consumed** during recovery when a default draws random
   numbers (`layout_with_drl`'s `seed`), making seeded reproduction
   impossible.

## The fix

Capture supplied-ness in the host frame with `missing()` —
`supplied = c(weights = !missing(weights), …)` — and pass that to the
helper. `missing()` never forces a promise, so tail defaults are not touched
until the function body first uses them; the value comparison (and the
`defaults` re-evaluation) disappears entirely. The generated blocks shrink
accordingly (−99 lines across the 66 blocks on `main`, all regenerated
here).

A second migration fixture with exactly the hazardous default kinds —
cross-reference (`breaks = bins * 2`), fresh environment (`new.env()`), and
RNG draw (`runif(1)`) — locks in all three behaviors, including that the
RNG state stays untouched by recovery.

## Validation

- `filter = "migration|generate-migrations"`: FAIL 0 | PASS 73
  (all existing recovery/error/snapshot semantics unchanged).
- Legacy `diversity(g, weights, vids)` now recovers with the single soft
  deprecation instead of erroring.
- Generator remains idempotent (re-run produces no diff).

## Follow-up once merged

The open topic PRs carry blocks generated by the previous generator; their
own CI stays green (the drift check runs each branch's committed generator),
but after this merges I will regenerate and push each open topic branch so
the whole tree converges on the `missing()`-based blocks.

---
_Generated by [Claude Code](https://claude.ai/code/session_016M32izVHZPfxAqemAe4BrX)_

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -441,22 +441,16 @@ betweenness <- function(
   if (...length() > 0L) {
     .arg_handle <- migrate_recover_args(
       list(...),
-      current = list(
-        directed = directed,
-        weights = weights,
-        normalized = normalized,
-        cutoff = cutoff
+      supplied = c(
+        directed = !missing(directed),
+        weights = !missing(weights),
+        normalized = !missing(normalized),
+        cutoff = !missing(cutoff)
       ),
       recover_new = c("directed", "weights", "normalized", "cutoff"),
       recover_old = c("directed", "weights", "normalized", "cutoff"),
       match_names = c("directed", "weights", "normalized", "cutoff"),
       match_to = c("directed", "weights", "normalized", "cutoff"),
-      defaults = list(
-        directed = TRUE,
-        weights = NULL,
-        normalized = FALSE,
-        cutoff = -1
-      ),
       head_args = c("graph", "v"),
       fn_name = "betweenness"
     )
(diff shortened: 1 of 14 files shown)


krlmlr:
Agreed to close.

PR #2757, opened 2026-07-26
docs: zoning rules for arguments around the ellipsis

krlmlr:
# Central PR: moving optional arguments past the ellipsis, repo-wide

Coordination point for applying the argument-migration pattern
(#2684, #2686, #2743) across all exported functions with optional arguments.
After #2779 landed the registry-split logic on `main`, this PR is **docs-only**:
the design rationale in `CONTRIBUTING.md`
(section *Argument Order and the Ellipsis*),
with pointers from `AGENTS.md`, `tools/README.md`,
and `tools/migrations/README.md`.

## The zoning rules (summary)

**Head (before `...`, positional):**

1. All required arguments, data first.
   This includes functions with three or more required arguments:
   a required argument after `...` would be keyword-only but still required,
   which hides it from the signature and worsens error messages.
2. At most a few *defining* optional arguments — those that say **what**
   is computed, not **how**:
   the objects operated on (`v` in `degree()`),
   model parameters (`power`, `m` in `sample_pa()`),
   or a variant-selecting enum (`type` in `as_adjacency_matrix()`).

**Tail (after `...`, keyword-only):** everything else, roughly ordered as
interpretation modifiers (`mode`, `directed`, `weights`, `loops`) →
output shape (`normalized`, `names`, `sparse`) →
algorithm/tuning (`algorithm`, `options`, callbacks) →
deprecated arguments last.

**Deliberate deviation from the tidyverse design guide:**
[dots-after-required](https://design.tidyverse.org/dots-after-required.html)
puts `...` immediately after the required arguments.
We allow a small head of defining optional arguments because idiomatic
positional calls (`degree(g, 1:3)`, `sample_pa(100, 1.5)`) are pervasive in
20 years of scripts, books, and teaching material; deprecating the idiomatic
form would create warning noise without a safety benefit.
The safety value of keyword-only arguments lies in the long tail of options.

**Invariants of the sweep:**

- No behavior changes: defaults stay untouched, no renames in this wave
  (renames remain expressible in the registry via the bare-symbol convention
  and can be layered per function family later).
- Old positional/abbreviated calls keep working through the generated
  ARG_HANDLE recovery, with a single `lifecycle::deprecate_soft("3.0.0", …)`.
- Deprecated functions are not touched.

## Topic PRs (226 functions, 21 PRs)

Each topic PR is a single commit on `main` (the split logic landed via #2779).

| PR | topic | functions |
|---|---|---|
| #2759 | structural-properties (incl. components, paths) | 37 |
[...]

Issue #1572, opened 2024-11-04
Standardize the handling of optional vertex parameters


szhorvat:
The handling, and default value, of optional vertex parameters should be standardized, and Stimulus should be updated accordingly.

There are currently two C functions which take an optional vertex parameter: `igraph_random_spanning_tree()` and `igraph_fundamental_cycles()`. Only the former is exposed at the moment. We need a decision on this issue before I expose the latter.

In C, these two functions interpret a negative vertex index as not passing a vertex.

In R, the situation is more complicated because vertices can be referred to either by a string name or by an integer index. Both must be supported.

IMO there are two reasonable options:

 1. Signal that no vertex is provided by passing a non-positive value. The default parameter value should be 0. `NULL` is not allowed, nor are passing multiple vertices.
 2. Signal that no vertex is provided by passing `NULL`. The default parameter value should be `NULL`. Non-positive integers are not allowed. Multiple vertices are not allowed.

We currently have (1) mostly implemented. The auto-generated code looks like this:

```
  if (!is.null(vid)) vid <- as_igraph_vs(graph, vid)
  if (length(vid) == 0) {
    stop("No vertex was specified")
  }
```

However, currently, `OPTIONAL VERTEX`, without an explicit default, uses `NULL` as the default, which triggers the "No vertex was specified" error. If we go with choice (1), Stimulus should be updated to use `0` instead of `NULL` as the default. If we go with choice (2), the autogen code template must be updated for this.

@krlmlr, can you make a decision on this? When that is done, can you adapt Stimulus, @Antonov548 ?

maelle:
On the R side, having `NULL` as default makes sense **for the interface**. Now, we can internally convert it to whatever the C side needs.

https://design.tidyverse.org/defaults-short-and-sweet.html

krlmlr:
I agree that the default on the interface should be `NULL` .

@Antonov548: Can you please help with Stimulus? Are the requirements clear?

Antonov548:
> igraph_fundamental_cycles

Sorry, I'm not follow anymore. From the all pull requests and discussion I have thoughts that changes to Stimulus are not required.
Is there still anything I need to update in Stimulus?

szhorvat:
Since we're going with option 2, I think you're right. But to be sure, it'd be great to actually implement this (by removing the default of 0 from `igraph_random_spanning_tree` and letting it use the `OPTIONAL` default of `NULL`) and make sure it all works.

krlmlr:
[...]

PR #2784, opened 2026-07-26
feat(migrations): require constant defaults, resolve complex ones in the body

krlmlr:
Reworked per review direction: **no grandfather mechanism**.
Two commits:

## 1. `refactor:` move the 23 non-constant defaults into the bodies

All non-constant defaults across the 22 migrated functions on `main`
become constant `NULL` in the signature,
resolved in the body once all arguments are available:

- `V(graph)` / `E(graph)` selectors (19 arguments):
  `alpha_centrality()`, `betweenness()`, `closeness()`, `diversity()`,
  `edge_betweenness()`, `harmonic_centrality()`, `page_rank()`,
  `power_centrality()`, `strength()`, `all_simple_paths()`,
  `all_shortest_paths()`, `constraint()`, `degree()`, `distances()`
  (`v` and `to`), `ego()`, `ego_size()`, `knn()`, `make_ego_graph()`,
  `shortest_paths()`, `which_mutual()`;
- `as_adjacency_matrix(sparse = igraph_opt("sparsematrices"))`;
- `max_bipartite_match(eps = .Machine$double.eps)`.

Body pattern, inserted right after the ARG_HANDLE block
(after `ensure_igraph()` where present):

```r
if (is.null(vids)) {
  vids <- V(graph)
}
```

**Behavior notes for review:**

- Passing the selector `NULL` explicitly now means the full vertex/edge
  set, same as omitting the argument.
  Previously an explicit `NULL` (including a computed `c()` that came out
  empty) was coerced to an *empty* selection — an undocumented, untested
  accident of `as_igraph_vs()`. The new semantics are pinned in
  `test-constant-defaults.R`; if you'd rather keep the accidental
  empty-selection meaning, say so and I'll switch the resolution to a
  `missing()` check instead.
- Watch the interplay with #1572, which settled on `NULL` meaning
  "*no* vertex" for optional *single-vertex* parameters
  (`random_spanning_tree(vid = NULL)`). Plural selectors saying
  "`NULL` = all" vs. singular saying "`NULL` = none" is a distinction
  worth an explicit line in the docs policy.
- This **fixes a live bug on `main`**: legacy positional recovery of
  `diversity()` died with a spurious "supplied more than once" error
  because re-evaluating its `V(graph)` default never compares
  `identical()`. With a constant default the recovery machinery needs no
  further changes — the defect class #2783 addressed disappears at the
  root, per "avoid nonconst defaults at all".

## 2. `feat:` the generator enforces the rule — no escape hatch

`is_constant_default()` (recursive AST check): constant = literals,
`NULL`/`TRUE`/`FALSE`/`NA*`/`Inf`/`NaN`, `c()`/`list()` of constants,
typed empty vectors (`integer()`, `numeric()`, ...),
unary sign, parentheses, the `deprecated()` sentinel.
Everything else errors at generation time —
option lookups, `V(graph)`, cross-references (`bins * 2`), RNG draws,
`rep()`, `.Machine$…`, bare symbols including `T`/`F`.
A leftover `nonconst_defaults` field errors with its own message.
Documented in `tools/migrations/README.md`
(and the matching rule prose is on #2757).

## Round 2: the whole package

After merging `main` (which migrated the games/layout/make/… topics
behind the ellipsis), the same treatment now covers **every exported,
non-deprecated, hand-written function** — the remaining 106 non-constant
defaults across 67 functions, migrated or not, plus the two
`plot_dendrogram()` S3 methods for coherence with their generic.
The registry `new` literals of the 32 affected migrated functions carry
the same `NULL` defaults, so the regenerated recovery blocks stay
constant.

| File | Functions | Defaults | What |
|---|---|---|---|
| games.R | 15 | 29 | `type.dist`, `pref.matrix`, `types`, `pref`, `agebins`, `p` — cross-references, incl. the `sample_*`/spec-wrapper pairs |
| layout.R | 11 | 26 | `vcount()`/`edge_density()`-derived constants, `drl` seed + options, `normalize()` limits |
| attributes.R | 9 | 9 | `index = V(graph)`/`E(graph)` accessors, getters and setters |
| print.R | 1 | 6 | `print.igraph()`'s six `igraph_opt()`-backed arguments |
| structural-properties.R | 6 | 6 | `bfs()`/`dfs()` `rho = parent.frame()`, `eids`/`v` selectors |
| make.R | 3 | 4 | `make_graph()` `n`/`dir` and the directed/undirected wrappers (`missing()` → `is.null()`) |
| embedding.R | 2 | 3 | `cvec` (references `weights`), `options = arpack_defaults()` |
| operators.R | 3 | 3 | `compose()`/`disjoint_union()` `graph.attr.comb`, `reverse_edges(eids = )` |
| plot.R | 1 | 3 | `mark.col`/`mark.border` (resolve after the `mark.groups` coercion), `mark.shape = 1/2` → `0.5` |
| cocitation.R, community.R, plot.shapes.R | 5 | 6 | `v` selectors, `contract()`/`plot_dendrogram()` options, `add_shape()` `clip`/`plot` |
| 11 more files | 11 | 11 | one each: `hits_scores()`, `plot_hierarchy()`, `as_undirected()`, `local_efficiency()`, `graphlet_proj()`, `sample_motifs()`, `indent_print()`, `similarity()`, `simplify()`, `stochastic_matrix()`, `count_triangles()` |

Special cases, handled deliberately: cross-referencing defaults resolve
in signature order (`area` before `repulserad`/`cellsize`, `types`
before `pref`, `agebins` before `pref`); `layout_with_drl()`'s random
default seed is now drawn inside the function with unchanged RNG stream
consumption for defaulted calls; `options = arpack_defaults()` resolves
in the body; `layout_with_gem(temp.min = 1/10)` and
`plot.igraph(mark.shape = 1/2)` became the literals `0.1`/`0.5` instead
of `NULL`; `sample_motifs(cut.prob = NULL)` needs no resolution because
the C layer already treats `NULL` as "no cuts" (matching `motifs()`);
and two places keep `NULL` as a *legal value* and therefore resolve via
`missing()` instead of `is.null()`: `print.igraph(max.lines = NULL)`
still prints everything, and `normalize(ymin = NULL)` still disables
normalization along that axis (as in `norm_coords()`). The
explicit-NULL flip described above now also applies to the
option-backed combination specs (`edge.attr.comb = NULL` etc. used to
mean an *empty* combination spec, dropping attributes).

### `c()` defaults were NULL in disguise

`c()` evaluates to `NULL`. The five empty-HRG fields that `fit_hrg()`
builds when no starting model is given (formerly `hrg()`'s signature
defaults `left = c()`, `right = c()`, …) therefore declared NULL values
that historically meant "empty sequence". Under the new doctrine NULL
is the resolve-in-body sentinel, so NULL-as-empty and NULL-as-default
cannot coexist: **empty-sequence defaults must be spelled as typed
empty vectors** (`numeric()` here — the fields are `as.numeric()`-coerced
on the spot; `layout_as_tree()`'s `root = numeric()` was already
spelled right and the checker now accepts it as a literal). Tests pass
`integer()` where an empty id selection is meant
(`max_degree(g, v = integer())` is 0). **Passing `NULL` explicitly to
mean "empty" is no longer supported for treated arguments — a
deliberate breaking change**: `max_degree(g, v = NULL)` now means *all*
vertices, pinned in `test-constant-defaults.R`.

## Remaining worklist (not this PR)

The exported hand-written remainder is **zero**: scanning every
function definition against the classifier finds no non-constant
default outside `R/aaa-*.R`. What is left is the Stimulus-generated
`_impl` layer — 103 defaults across 78 internal wrappers — and handling
those in the generator (constant-ness check there, or dropping defaults
from `_impl` signatures entirely) is still under discussion.

Validation: generator idempotent (generate → `air` → generate is a
byte-identical fixed point); full suite FAIL 0 · PASS 9250 (the only
error is the known sandbox network fetch in `test-foreign.R`);
gate-on `IGRAPH_LIFECYCLE_ERRORS=true` migration/constant-defaults
filter FAIL 0 · PASS 151; 85 Rd files regenerated via
`devtools::document()`.


krlmlr:
@schochastics @maelle: Running revdepchecks now. The `NULL` in combination with a "missing" check might look like `c()` (which is identical to `NULL`) means "empty set". Let's find out how badly it hurts.

krlmlr:
Claude says that revdep failures will all reproduce on main. This means to me that this PR is good for review as is, and that we deal with revdep problems elsewhere.

krlmlr:
@schochastics: Please squash when good.

krlmlr:
The revdep came back inconspicuous, this means any new failures must have come from previous commits. I'll ask Claude about those.

schochastics (approved):

Issue #2646, opened 2026-05-22
Revdep Failures


schochastics:
## Reverse dependency failures

32 reverse dependencies have live failures against the dev version of igraph. The full per-package analysis lives in [`revdep/problems-analysis.md`](https://github.com/igraph/rigraph/blob/main/revdep/problems-analysis.md). Grouped below by the underlying igraph change; each list tracks downstream PRs / issues.

Since #2834, each package's section is its own file — [`revdep/problems/<package>.md`](https://github.com/igraph/rigraph/tree/main/revdep/problems) — and `problems.md` is their concatenation.

### `bfs(father = )` / `dfs(father = )` defunct
The `father` argument was renamed to `parent` in igraph 2.2.0 and is now defunct; tidygraph passes it through to `igraph::bfs()` inside `map_bfs*` / `map_dfs*`, and the three packages below cascade through tidygraph.

- tidygraph https://github.com/thomasp85/tidygraph/pull/208
- [nat](https://github.com/natverse/nat)[-] [looks fixed](https://github.com/natverse/nat/blob/cf82fad17fe7cf2ed190f940c9f49efbc4156f7c/R/ngraph.R#L27-L35)
- dci (cascades through tidygraph) https://github.com/thomasp85/tidygraph/pull/208
- ggraph (cascades through tidygraph) https://github.com/thomasp85/tidygraph/pull/208
- scistreer (cascades through tidygraph) https://github.com/thomasp85/tidygraph/pull/208

### `tkplot()` removed
The interactive Tk plotter advanced to `deprecate_stop("3.0.0", "tkplot()")`; affected packages are direct callers and need to switch to `plot.igraph()`.

- Boptbd       [📧 EMAIL SENT]
- c3net        [📧 EMAIL SENT]
- ggm          [📧 EMAIL SENT] ✅
- optbdmaeAT   [📧 EMAIL SENT] 
- optrcdmaeAT  [📧 EMAIL SENT]

### `layout.spring()` defunct
Defunct since igraph 2.1.0; replacement is `layout_with_fr()`.

- comato [📧 EMAIL SENT]

### `hub.score()` defunct
Defunct since igraph 2.0.0; replacement is `hits_scores()$hub`.

- [ECoL](https://github.com/lpfgarcia/ECoL) https://github.com/lpfgarcia/ECoL/pull/66

### `get.edge()` removed
The function has been removed; replacement is `ends()`.

[...]

PR #2810, opened 2026-08-05
fix(migrations): ignore empty argument slots in `...`

krlmlr:
Supersedes #2808 — same fix for the four `argument is missing, with no default` revdep failures from #2646, restructured along the review comments there.

Prepared with Claude Code. The first commit is @schochastics' original from #2808, cherry-picked unchanged; the second is the review follow-up, so the delta against #2808 is reviewable on its own.

## The bug (unchanged from #2808)

A call may leave an argument slot empty — a trailing comma (`f(x, mode = "out", )`) or a skipped positional, which is what magrittr writes for `x %>% f(., , directed = FALSE)`. Under the old signatures those slots matched a formal by position and left it missing, so the formal's default applied.

Now that the optional arguments sit behind `...` (#2757–#2778), the empty slot lands in `...` instead, where the generated block's `list(...)` forces it:

```
Error in igraph::graph_from_adjacency_matrix(adjmatrix = x, mode = mode, weighted = TRUE, ) :
  argument is missing, with no default
Calls: ... -> <Anonymous> -> migrate_recover_args
```

Both shapes work on CRAN igraph 2.3.1 and error on `main` today:

```r
graph_from_adjacency_matrix(adjmatrix = m, mode = "undirected", weighted = TRUE, )
graph_from_data_frame(d, , directed = FALSE)
```

Four revdeps are affected: **tna** (31 test failures, plus examples and vignettes), **lagdynamics**, **modelbpp**, **NetSci**.

## What changed relative to #2808

Two review comments on #2808:

> * The entire arg checking code should be contained within the `if ()`, I don't understand the necessity of any code outside the `if`.
> * Use `list2()` or `pairlist()` or `pairlist2()` to avoid tripping over.

**`rlang::pairlist2(...)` replaces `list(...)`.** It hands the empty argument slots over as the missing argument rather than forcing them, so the error never arises in the first place. (`list2()` only tolerates a *trailing* empty and errors on `f(1, , 2)`; base `pairlist()` is `as.pairlist(list(...))` and forces just like `list()`. `pairlist2()` is the one that takes all of them.)

That removes the `migrate_capture_dots()` pre-pass — `substitute(...())`, `...elt()`, and a `list(values =, pos =)` structure threaded into `migrate_recover_args()`. The helper takes a plain dots list again and skips the empty slots itself, while still counting them towards the positional ordinal: an empty slot consumed a formal under the old signature, so `f(g, 5, , "in", TRUE)` still recovers into the *second* and *third* recoverable arguments.

**The generated block keeps `main`'s shape.** `migrate_recover_args()` returns `NULL` when nothing was recovered, which is what stops a `...` of nothing but empty slots from warning; `list2env()` and `lifecycle::deprecate_soft()` stay inline in the host frame, so the deprecation is still attributed to the user's frame with no caller-env plumbing. Against `main` the whole block is:

```diff
 if (...length() > 0L) {
   .arg_handle <- migrate_recover_args(
-    list(...),
+    rlang::pairlist2(...),
     current = list(weights = weights, type = type, directed = directed),
     ...
     fn_name = "migration_fixture"
   )
-  list2env(.arg_handle$values, environment())
-  lifecycle::deprecate_soft(
-    "3.0.0",
-    what = I(.arg_handle$what),
-    details = .arg_handle$details
-  )
+  if (!is.null(.arg_handle)) {
+    list2env(.arg_handle$values, environment())
+    lifecycle::deprecate_soft(
+      "3.0.0",
+      what = I(.arg_handle$what),
+      details = .arg_handle$details
+    )
+  }
 }
```

**`migrate_check_call_tags()` returns to the top of the block, as on `main`.** #2808 moved it behind the recovery gate; a forbidden prefix such as `di =` was an ambiguity error under the old signature, so it should be rejected whether or not the remaining slots happen to be empty.

## Changes

- `R/migrate-args.R` — `migrate_capture_dots()` removed; `migrate_recover_args()` skips empty slots and returns `NULL` when nothing was recovered.
- `tools/generate-migrations.R` — blocks pass `rlang::pairlist2(...)` and gate the rebinding on a non-`NULL` result.
- All 227 ARG_HANDLE blocks regenerated (output is still exactly what `air` formats, verified by regenerating after `air format .`).
- Tests in `tests/testthat/test-migration-fixture.R` for the trailing comma, a trailing comma after new-API arguments, an empty slot consuming its position, several empty slots, a `...` of nothing but empty slots, and `migrate_recover_args()` against `rlang::pairlist2()` directly.

## Verification

- `testthat::test_local()` green apart from two failures that need the environment rather than the code: `test-foreign.R:69` downloads from `github.com/igraph/graphsdb` (no network here) and `test-other.R:20` starts a `callr` subprocess that needs igraph *installed* rather than loaded via `pkgload`.
- Both reproducers above, plus `graph_from_data_frame(d, , )` (all slots empty — defaults kept, silent) and `graph_from_data_frame(d, FALSE)` (unaffected, `directed` is a head arg), behave as they do on CRAN 2.3.1.

### DIFF
--- a/R/adjacency.R
+++ b/R/adjacency.R
@@ -293,58 +293,61 @@ graph_from_adjacency_matrix <- function(
 ) {
   # BEGIN GENERATED ARG_HANDLE: graph_from_adjacency_matrix, do not edit, see tools/generate-migrations.R
   if (...length() > 0L) {
-    migrate_check_call_tags(
-      sys.call(),
-      c("a", "ad"),
-      "graph_from_adjacency_matrix"
-    )
-    .arg_handle <- migrate_recover_args(
-      list(...),
-      current = list(
-        mode = mode,
-        weighted = weighted,
-        diag = diag,
-        add.colnames = add.colnames,
-        add.rownames = add.rownames
-      ),
-      recover_new = c(
-        "mode",
-        "weighted",
-        "diag",
-        "add.colnames",
-        "add.rownames"
-      ),
-      recover_old = c(
-        "mode",
-        "weighted",
-        "diag",
-        "add.colnames",
-        "add.rownames"
-      ),
-      match_names = c(
-        "mode",
-        "weighted",
-        "diag",
-        "add.colnames",
-        "add.rownames"
-      ),
-      match_to = c("mode", "weighted", "diag", "add.colnames", "add.rownames"),
-      defaults = list(
-        mode = c("directed", "undirected", "max", "min", "upper", "lower", "plus"),
-        weighted = NULL,
-        diag = TRUE,
-        add.colnames = NULL,
-        add.rownames = NA
-      ),
-      head_args = c("adjmatrix"),
-      fn_name = "graph_from_adjacency_matrix"
-    )
-    list2env(.arg_handle$values, environment())
-    lifecycle::deprecate_soft(
-      "3.0.0",
-      what = I(.arg_handle$what),
-      details = .arg_handle$details
-    )
+    .migrate_dots <- migrate_capture_dots()
+    if (length(.migrate_dots$values) > 0L) {
+      migrate_check_call_tags(
+        sys.call(),
+        c("a", "ad"),
+        "graph_from_adjacency_matrix"
+      )
+      .arg_handle <- migrate_recover_args(
+        .migrate_dots,
+        current = list(
+          mode = mode,
+          weighted = weighted,
+          diag = diag,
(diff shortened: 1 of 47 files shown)


krlmlr:
Alternative to #2808 that looks a bit simpler. One small change is scheduled.

krlmlr:
Closing in favor of #2812, we can reiterate if for some reason full inlining isn't desired.

PR #2812, opened 2026-08-05
fix(migrations): recover legacy arguments against the old signature

krlmlr:
Supersedes #2808 and #2810 — one commit, based on `main`, nothing stacked. Prepared with Claude Code.

Fixes the four `argument is missing, with no default` revdep failures from #2646, by removing the runtime argument matcher rather than patching it.

## The bug

A call may leave an argument slot empty — a trailing comma (`f(x, mode = "out", )`) or a skipped positional, which is what magrittr writes for `x %>% f(., , directed = FALSE)`. Under the old signatures those slots matched a formal by position and left it missing, so the formal's default applied.

Now that the optional arguments sit behind `...` (#2757–#2778), the empty slot lands in `...`, where the generated block's `list(...)` forces it:

```
Error in igraph::graph_from_adjacency_matrix(adjmatrix = x, mode = mode, weighted = TRUE, ) :
  argument is missing, with no default
Calls: ... -> <Anonymous> -> migrate_recover_args
```

Four revdeps are affected: **tna** (31 test failures, plus examples and vignettes), **lagdynamics**, **modelbpp**, **NetSci**.

## The fix

`migrate_recover_args()` re-implemented base R's argument matching over `list(...)` and seven parallel configuration vectors. But that matching is something R already does, and did do before the migration — so the block declares the old signature and lets R match against it:

```r
# fmt: skip
if (...length() > 0L) {
  # Pre-3.0.0 signature: migration_fixture(graph, n, weight, kind, directed)
  .old_signature <- function(weight, kind, directed, ...) {
    if (...length() > 0L) {
      .arg_extra <- base::names(base::substitute(...()))
      .arg_extra <- .arg_extra[base::nzchar(.arg_extra)]
      if (base::length(.arg_extra) == 0L) cli::cli_abort("Too many arguments passed to {.fn migration_fixture}.", call = base::parent.frame())
      cli::cli_abort(base::c("Unexpected argument passed to {.fn migration_fixture}: {.arg {(.arg_extra)}}.", i = "Arguments after {.arg ...} must be spelled out in full."), call = base::parent.frame())
    }
    base::c(
      if (!base::missing(weight)) base::list(weights = weight),
      if (!base::missing(kind)) base::list(type = kind),
      if (!base::missing(directed)) base::list(directed = directed)
    )
  }
  .arg_handle <- .old_signature(...)
  if (base::length(.arg_handle) > 0L) {
    .arg_names <- base::names(.arg_handle)
    .arg_conflict <- base::intersect(.arg_names, base::c(
      if (!base::missing(weights)) "weights",
      if (!base::missing(type)) "type",
      if (!base::missing(directed)) "directed"
    ))
    if (base::length(.arg_conflict) > 0L) cli::cli_abort(base::c("Argument {.arg {(.arg_conflict)}} of {.fn migration_fixture} was supplied more than once.", i = "Pass it exactly once, by its new name {.arg {(.arg_conflict)}}."))
    base::list2env(.arg_handle, base::environment())
    lifecycle::deprecate_soft(
      "3.0.0",
      what = base::I("Calling `migration_fixture()` with positional or abbreviated arguments"),
      details = base::c(
        i = base::paste0("Detected call:  migration_fixture(", base::paste(base::c("graph", "n", base::c(weights = "weight", type = "kind", directed = "directed")[.arg_names]), collapse = ", "), ")"),
        i = base::paste0("Use instead:    migration_fixture(", base::paste(base::c("graph", "n", base::paste0(.arg_names, " = ")), collapse = ", "), ")")
      )
    )
  }
}
```

**#2646 falls out of it.** `...` is never materialised as a list, so an empty slot passes straight through into `.old_signature()`, leaves its formal missing, and still consumes its position — precisely the pre-migration behaviour. `f(g, 5, , "in", TRUE)` lands on the second and third recoverable arguments without anything having to arrange for it, and `rlang::pairlist2()` (which #2810 needed) is not involved.

**Everything is unrolled**, one line per argument: `if (!base::missing(<old>))` picks the value up, and `if (!base::missing(<new>))` reports it as already supplied. Asking `missing()` in the host frame answers "was this passed" directly, so the block needs neither the defaults nor a comparison against them.

**No runtime layer is left.** The guards and messages are inline too, so `R/migrate-args.R` is deleted; `lifecycle::deprecate_soft()` still sits in the host frame and resolves `user_env` to the user on its own.

**Every call is namespace-qualified.** The block is spliced into someone else's function body, where a formal may be named `names` or `c` — and a *missing* formal is worse than a shadowing one, since R forces the promise while looking for a function to call and reports `argument "names" is missing`. Qualifying unconditionally keeps every block the same shape instead of making it depend on the signature. Only operators and `...length()` are left bare — a formal cannot shadow those. The measured cost of `base::` on the hot `...length()` guard is ~80 ns per call.

## Behaviour is unchanged, down to the message text

The diff to the *existing* snapshots is empty; the new entries are additions. Two things were needed for that:

- **`.old_signature()` takes a trailing `...` of its own**, so an unknown or surplus argument lands there instead of tripping base R's "unused argument"; it is reported with the message the matcher produced. The tags come from `substitute(...())`, which forces nothing — a surplus slot may itself be empty.
- **Ambiguous abbreviations are rejected up front.** `.old_signature()` only knows the old names, so it would silently resolve some abbreviations the matcher rejected — `at` for `as_adjacency_matrix(attr =)`, which the migration renamed to `weights` while a deprecated `attr` formal remains. The generator enumerates those tags (48 functions) and the block tests the dot tags against them first.

One case is deliberately **stricter**: an argument passed explicitly at its own default value, together with a positional recovering into it, is now a conflict. `migration_fixture(g, 5, weights = NULL, 1:3)` supplies `weights` twice and errors, where comparing against the default could not tell the two apart and let the positional win silently.

## Tests

A third fixture, `migration_fixture_shadow(graph, ..., weights, names, c, attr)`, covers the two hazards a block faces in a real function body — both taken from `as_adjacency_matrix()`, which had no direct coverage: formals named `names` and `c` that shadow the block's own calls, and `attr` renamed to `weights` while a deprecated `attr` formal keeps the name. Every path through it is asserted deterministically and snapshotted: recovery with the shadowing formals missing (the dangerous case), recovery *of* a shadowing formal by abbreviation and by position, `attr =` binding the retained formal, and the ambiguous, unknown, surplus and supplied-twice errors.

The empty-slot behaviour is snapshotted alongside its deterministic tests, because the `Detected call` line is the readable form of the position bookkeeping:

```
> migration_fixture("g", 5, , "in", TRUE)
Warning:
Calling `migration_fixture()` with positional or abbreviated arguments was deprecated in igraph 3.0.0.
i Detected call: migration_fixture(graph, n, kind, directed)
i Use instead: migration_fixture(graph, n, type = , directed = )
```

One more edge is pinned: a slot past the last recoverable one is surplus whether or not it is empty, so `migration_fixture("g", 5, 1, 2, 3, )` errors — the old signature had nothing to match a sixth argument against either.

## Cost

The generated blocks grow by ~2000 lines across all 227 — 21 lines per block before against ~30, and 60 for the widest (`layout_with_fr()`, 17 recoverable arguments). The blocks open with `# fmt: skip` so `air` leaves the layout to the generator: that buys some of it back, and it removed the code that had to predict how `air` would wrap.

## Verification

- Full `testthat::test_local()` green apart from two failures that need the environment rather than the code: `test-foreign.R:69` downloads from `github.com/igraph/graphsdb` (no network here) and `test-other.R:20` starts a `callr` subprocess that needs igraph *installed* rather than loaded via `pkgload`.
- Both #2646 reproducers, plus `graph_from_data_frame(d, , )` (all slots empty — defaults kept, silent), behave as they do on CRAN 2.3.1.
- Generator output is untouched by `air format .`.


krlmlr:
Another idea for fixing: inlining and unrolling. Generated code doesn't really need static helper functions, can do everything from first principles. Work incoming.

krlmlr:
Can it really be as simple as this? My favorite so far. A bit longer, but wins by a margin when debugging. Need to iterate a bit, let me know what you think.

krlmlr:
One more change (new snapshot tests), then ready for review.

krlmlr:
Ready for review.

Revdeps running: https://github.com/igraph/rigraph/actions/runs/31091591505

schochastics (approved):
lgtm

PR #2822, opened 2026-08-07
DO NOT MERGE — test only: `windows-11-arm` without the smoke test

krlmlr:
**Do not merge.** This PR exists only to exercise #2821 on the affected platform, and it carries CI scaffolding that must never reach `main`.

It contains the single commit of #2821 plus one throwaway commit on top that:

- skips `rcc-smoke` (`if: false`), which also skips `rcc-smoke-check-matrix` and `rcc-suggests`, since both depend on it;
- drops `rcc-full`'s `needs: rcc-smoke` and the `if:` guard on the generated matrix, so it runs on its own;
- replaces the generated matrix with the single entry under test, `windows-11-arm` on R 4.6, whose `env` value is verbatim what `.github/workflows/versions-matrix/action.R` now emits for it;
- checks out the PR head directly instead of the SHA the smoke test would have produced.

So the only job that runs is `rcc: windows-11-arm (4.6)`. Expected outcome: `Status: 1 WARNING` for the flang unused-argument diagnostics, as on `main` today, but the job green, because `error_on` is now `"error"` for this entry.

The R-devel half of the breakage is #2823, with its own checker in #2824.

Close this once #2821 is settled.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -73,6 +73,11 @@ jobs:
     # mutate the repo. Revisit once the arm image is generally available.
     runs-on: ubuntu-26.04
 
+    # DO NOT MERGE -- this branch exists to exercise the windows-11-arm entry
+    # on its own. Skipping the smoke test also skips rcc-smoke-check-matrix and
+    # rcc-suggests, which both depend on it.
+    if: false
+
     outputs:
       sha: ${{ steps.commit.outputs.sha }}
       versions-matrix: ${{ steps.versions-matrix.outputs.matrix }}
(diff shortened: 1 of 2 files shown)


krlmlr:
Green: [`rcc: windows-11-arm (4.6)`](https://github.com/igraph/rigraph/actions/runs/31196070800), the only job in the run, one hour five minutes.

The install-time flang diagnostics are unchanged — nothing in the package can suppress them — so the check still ends in `Status: 1 WARNING`. The job passes because `error_on` is `"error"` for this entry now. I can't quote the status line: a passing job uploads no check artifact, and the check summary falls outside the retrievable log tail.

#2821 is merged, so this has served its purpose and can be closed.

---

PR #2823, opened 2026-08-07
fix(ci): Point the ccache-wrapped compiler variables at a wrapper script

krlmlr:
Prepared with Claude Code. One of two PRs unbreaking `rcc` on `main`; the `windows-11-arm` half is #2821.

`rcc: ubuntu-26.04 (devel)` started failing with R-devel r90366 (2026-08-06), with `Status: 1 NOTE` from `checking compiled code`, repeated dozens of times:

```
ccache: invalid option -- 'E'
Warning in system(cmd, intern = TRUE) :
  running command 'ccache -E -I/opt/R/devel/lib/R/include /tmp/.../file....h  -DR_INTERFACE_PTRS' had status 1
```

The install action writes `CC = ccache gcc -std=gnu2x` (and the same for the C++ variants) into `~/.R/Makevars`. That reads correctly only if the consumer treats the value as a whole command line. R-devel's `checking compiled code` preprocesses the R headers with the *first word* alone, so it runs `ccache -E ...`, which is ccache's own option parser rejecting `-E`. The resulting warning output is non-empty, so `R CMD check` turns it into a NOTE — one with no findings in it, which is the tell.

The fix writes one small wrapper script per variable and points the variable at it, so the value is a single token that runs ccache internally and works whichever way it is read.

Ruled out: `cc` on the runner resolving to ccache through the masquerade symlinks. `/usr/lib/ccache/cc -E …` works fine — ccache only rejects `-E` when invoked under its own name, which is what the log shows.

Verified locally that `R CMD SHLIB` still compiles and links through the wrapper, that `R CMD config CC` returns the wrapper, that `using C compiler: 'gcc ...'` is still reported correctly, and that both `ccache gcc … -E` and the wrapper accept the preprocessing invocation.

Testing: #2824 carries this commit and trims the matrix to `ubuntu-26.04` R-devel alone, without the smoke test. Do not merge that one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmSX7BpDELCDras4pvRGby)_

### DIFF
--- a/.github/workflows/install/action.yml
+++ b/.github/workflows/install/action.yml
@@ -169,7 +169,7 @@ runs:
 
     - name: Use ccache for compiling R code, and parallelize
       run: |
-        mkdir -p ~/.R
+        mkdir -p ~/.R ~/.R/bin
         # Infer each compiler default for the active R version from R CMD config,
         # then prepend ccache. Reading the resolved defaults preserves the
         # standard flags R selects per version (e.g. -std=gnu2x for C, -std=gnu++17
@@ -182,17 +182,36 @@ runs:
         # CXXnnSTD variable (CXX17STD = -std=gnu++17, ...) and applies it on top
         # at compile time, so wrapping only CXXnn with ccache is correct and we
         # must NOT pin -std here ourselves.
+        # The value is a wrapper script rather than a bare "ccache <compiler>"
+        # because not every consumer of these variables treats them as a full
+        # command line. R-devel's "checking compiled code" preprocesses the R
+        # headers with tools:::ccE(), which since r90356 reads the compiler from
+        # `R CMD config CC` and since r90359 truncates it at the first space
+        # (src/library/tools/R/apitools.R). For "ccache gcc -std=gnu2x" that
+        # leaves the bare "ccache", which clears the Sys.which() guard because
+        # ccache is a real program, then fails with "invalid option -- 'E'";
+        # R CMD check reports the resulting output as a NOTE. A single-token CC
+        # that is also a real compiler driver satisfies both readings.
         empty_makevars="$(mktemp)"
         for var in CC CXX CXX11 CXX14 CXX17 CXX20 CXX23; do
           val="$(R_MAKEVARS_USER="$empty_makevars" R CMD config "$var" 2>/dev/null || true)"
           if [ -n "$val" ]; then
-            echo "$var = ccache $val" >> ~/.R/Makevars
+            wrapper=~/.R/bin/ccache-"$var"
+            printf '#!/bin/sh\nexec ccache %s "$@"\n' "$val" > "$wrapper"
+            chmod +x "$wrapper"
+            echo "$var = $wrapper" >> ~/.R/Makevars
           fi
         done
         rm -f "$empty_makevars"
         echo 'MAKEFLAGS = -j2' >> ~/.R/Makevars
         echo "===== generated ~/.R/Makevars ====="
         cat ~/.R/Makevars
+        echo "===== generated ~/.R/bin wrappers ====="
+        for wrapper in ~/.R/bin/ccache-*; do
+          [ -f "$wrapper" ] || continue
+          echo "--- $wrapper"
+          cat "$wrapper"
+        done
         echo "==================================="
 
         echo 'CCACHE_SLOPPINESS=locale,time_macros' | tee -a $GITHUB_ENV


krlmlr:
This might well be a transient error.

krlmlr:
Possible, and worth settling with data rather than argument — I've opened #2825 as a control: same scaffolding as the checker #2824 (skip the smoke test, hardcode the matrix to the single `linux_devel` entry), but on unpatched `main`. Both run the same job at the same time against the same R-devel snapshot and differ only in the fix.

What I can say already:

The failure mode itself is not stochastic. The log shows the program `ccache` being run with `-E`, which is ccache's own option parser rejecting a compiler flag — it fails 100% of the time. I checked the alternative explanation that `cc` on the runner resolves to ccache through the masquerade symlinks: `/usr/lib/ccache/cc -E …` works fine, because under a compiler name ccache forwards to the real compiler. Only the literal name `ccache` errors. And it happened once per header, dozens of times in the one run, not intermittently within it.

What is genuinely uncertain is the *exposure*, and that's the part that could look transient. The check ran clean at 04:02 UTC and failed at 12:47 UTC the same day, with the failing run on `R Under development (unstable) (2026-08-06 r90366)`. I could not recover the revision from the earlier job — it succeeded, so there's no `00check.log` artifact, and the R banner falls outside the retrievable log tail. R-devel r90185 (2026-06-21) still hardcodes `cc` in `tools:::ccE`, so the behaviour changed somewhere between the two, and a snapshot rolling in between the two runs is the obvious candidate. That does mean it could equally roll back out.

Either way the change stands on its own: the value is only correct when read as a whole command line, and a single-token `CC` that runs ccache internally is correct under both readings. If the control comes back green, treat this as robustness rather than a fix and judge it accordingly.

---

krlmlr:
Found it in the r-svn mirror — it's a deliberate two-commit change in `src/library/tools/R/apitools.R`, both on 2026-08-06, so not transient.

**r90356** ([`4f07f6e`](https://github.com/r-devel/r-svn/commit/4f07f6e), ligges) — "fix Windows issues: use of `cc` generalized, avoid messing with Windows style paths". `ccE()` stopped hardcoding `cc`:

```r
-    if (Sys.which("cc") == "")
-        stop("'cc' is not on the path")
-    cmd <- sprintf("cc -E %s %s", include, tfile)
+    CC <- Rcmd(c("config", "CC"), stdout = TRUE)
+    if (Sys.which(CC) == "")
+        stop(paste(sQuote(CC, q = FALSE), "is not on the path"))
+    cmd <- sprintf("%s -E %s %s", CC, include, tfile)
```

At this revision our `CC` would already have broken the check, but differently: `Sys.which("ccache gcc -std=gnu2x")` is `""`, so it would have stopped with "is not on the path".

**r90359** ([`ca19f35`](https://github.com/r-devel/r-svn/commit/ca19f35), luke) — "Clear out command line arguments from CC obtained from configure":

```r
     CC <- Rcmd(c("config", "CC"), stdout = TRUE)
+    CC <- sub("\\s+.*", "", CC) ## clear out commant line arguments
```

That is exactly the "first word" behaviour I inferred from the log. `sub("\\s+.*", "", "ccache gcc -std=gnu2x")` is `"ccache"`; `Sys.which("ccache")` is non-empty, so it clears the guard, and then `ccache -E …` runs and fails.

The failing job was on r90366, which is past both. So the earlier clean run at 04:02 UTC was simply on a pre-r90356 snapshot — the exposure window opened that day and stays open. I'd expect the control #2825 to be red.

Worth noting where the gap is, if we want to report it upstream: the truncation is applied to the command as well as to the lookup, but `CC` is a command line, not a program name. `system()` would have run `ccache gcc -std=gnu2x -E …` quite happily — only `Sys.which()` needs the first word. Feeding the truncated value to `Sys.which()` alone and keeping the full string in `cmd` would fix it for every `CC` of the form "launcher compiler flags", which is the documented way to use ccache, distcc, and sccache from `~/.R/Makevars`.

Either way this PR is the right shape on our side: a single-token `CC` that is also a real compiler driver satisfies both readings, so it survives whichever way upstream settles.

---

krlmlr:
A/B is in, and it is not transient.

Both jobs ran within five minutes of each other, both on `R Under development (unstable) (2026-08-06 r90366)`, differing only in this commit:

- **control, unpatched `main`** ([#2825](https://github.com/igraph/rigraph/pull/2825), [job](https://github.com/igraph/rigraph/actions/runs/31196456844/job/92925733610)) — **failed**. `Status: 1 NOTE`, `checking compiled code ... NOTE`, with 46 `ccache: invalid option -- 'E'` lines in `00check.log`. Byte-for-byte the failure seen on `main`.
- **with the fix** ([#2824](https://github.com/igraph/rigraph/pull/2824), [job](https://github.com/igraph/rigraph/actions/runs/31196136443/job/92924696901)) — **passed**, in 17 minutes.

That matches the r-svn reading exactly: r90356 and r90359 are both in r90366, so every R-devel snapshot from here on hits it until upstream changes course.

I have amended the commit to cite the two revisions and the file, so the workaround is checkable against upstream later, and rebased the checker on top. No behavioural change — only the comment and the commit message.

---

PR #2831, opened 2026-08-08
ci(revdep2): Run several checks at once inside each shard

krlmlr:
Prepared with Claude Code.

## Where we were

Parallelism had one axis. `max-parallel` shards run at once — that is the sharding the whole workflow is built around — but inside a shard, `shard.R` called `rcmdcheck::rcmdcheck()` in a plain `for` loop, one package at a time. A hosted runner has four cores and one `R CMD check` keeps about one of them busy, so a shard left most of the machine idle for its entire 300-minute deadline.

The second axis is exactly what `revdepcheck::revdep_check(num_workers =)` does on a single machine, and it is what this adds.

## What changed

Each shard now runs `check-workers` checks at once, default 2.

A check goes into a `callr::r_bg()` background process running exactly the `rcmdcheck()` call the driver used to run inline, so the arguments, the timeout and the returned object are unchanged. Results are collected in the driver, one at a time, so nothing that reads or writes the shard's state ever runs concurrently. The checks are independent by construction: separate processes, a check directory each, and a library nothing writes to while a phase is running — the dev binary goes in *between* the phases, with the queue drained.

The queue is heaviest-first. With one worker that only moved the log around; with several it decides the makespan, since starting the long checks last is how a phase ends with one straggler and idle workers. It also means the packages a deadline defers are the cheap ones.

The mechanics live in `run_queue()` in `util.R` — a fixed-width queue with `start`/`alive`/`finish`/`skip`/`tick` callbacks — so the scheduling can be tested without running a single check.

## The plan is where the two axes meet

A package's weight is still what one check of it costs; that number goes into `plan.json` unchanged, because it is what the shard's own deadline test asks before starting a check. But the shard *sizing* now divides by the worker count, since that many packages advance per minute of a shard's deadline. On run 31270092803's numbers — 12263 check minutes, capacity 240, 20 lanes:

- 1 worker: 12263 min of shard time → 60 shards, 3 waves
- 2 workers: 6132 min → 40 shards, 2 waves
- 3 workers: 4088 min → 20 shards, 1 wave

Twenty fewer setups paid at two workers, and a wave less to wait through.

The shard reads the worker count back out of `plan.json` rather than its own environment: a shard running one at a time what was planned for two takes twice its estimate and defers the difference.

## The calibration stays honest by itself

Measured check times come from runs that ran under some worker count, so they already carry whatever contention that count caused. Where checks do not contend, the measurement stays put and dividing by the worker count buys the whole factor. Where they contend completely, the measurement doubles and the division gives back exactly what one worker would have done. Only the first run after the knob changes is estimated on the previous regime's numbers.

The worker count is recorded per shard in the timings artifact, so the regime behind a number is never a guess.

## Risk, and what watches it

Two concurrent checks share the runner's 16 GB, which is why the default is 2 rather than 4 — a check that compiles hard wants that headroom. The shard step now also gets the preflight's resource sampler and line-buffered output, so if two at once is too many the memory curve is in the log and in `resources.log` in the shard artifact, rather than being inferred from a bare `exit 143`. Lowering `check-workers` is then a one-input dispatch away.

## Testing

- `run_queue()`: concurrency never exceeds the cap (verified against a peak counter), `workers = 1` reproduces the old sequential order exactly, `skip()` drops items without starting or stalling them, and the all-skipped and empty-input cases terminate.
- The `callr` + `rcmdcheck` contract the shard depends on, against two real packages checked concurrently: both came back as `rcmdcheck` objects with parsed statuses, 53 s of check work in 27 s of wall clock. A check that cannot start returns the child's own condition, unwrapped from callr's, with `timed_out` correctly false.
- `plan.R` run end to end against CRAN at 1, 2 and 4 workers on the same five packages: 3, 2 and 1 shards, with per-package `weight_minutes` byte-identical across all three — the division moves the sizing and nothing else.

R sources parse and are `air`-formatted; the workflow parses as YAML and the changed step passes `bash -n`.

## A note on the branch

This is a second PR rather than more commits on #2830, which merged while this was being written; it is based on current `main` and includes it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1xpHRV7yVfgtJg4vp9P7z)_

### DIFF
--- a/.github/workflows/revdep2.yaml
+++ b/.github/workflows/revdep2.yaml
@@ -31,6 +31,12 @@
 # never a job failure; the shard defers what its deadline cannot fit and says
 # so.
 #
+# Parallelism has two axes and they multiply: `max-parallel` shards run at
+# once, and each shard runs `check-workers` checks at once (the runner has
+# four cores and one R CMD check keeps about one busy). The plan sizes shards
+# on both, so a shard holds as much check work as its deadline and its workers
+# can really get through.
+#
 # The collect job folds every shard artifact (and, on a retry, the untouched
 # results of the run being retried) into revdepcheck-style reports (README.md,
 # problems.md, failures.md, cran.md), a machine-readable manifest, and the
@@ -100,6 +106,10 @@ on:
         description: "Shards to run concurrently, and so the wave size the plan cuts: set it to the concurrency the account really has, never more (GitHub queues past its own limit anyway)"
         type: string
         default: ""
+      check-workers:
+        description: "Checks one shard runs at once; the runner has four cores and one check keeps about one busy, so this multiplies with max-parallel"
+        type: string
+        default: ""
       refresh-baseline:
         description: "Re-check the CRAN version even where a baseline is reusable"
         type: boolean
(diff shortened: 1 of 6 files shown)


krlmlr:
Using a simpler parallelization.

PR #2841, opened 2026-08-16
docs(revdep): Add reprexes for bootcluster, checked, htna and SEMdeep

krlmlr:
The four packages `revdep2` run 31048405399 flagged that the committed `revdep/` report does not. Each gets the two files this directory uses -- a runnable script and the reprex output -- and they turn out to be four different stories:

* **bootcluster** calls `sample_degseq(method = "simple.no.multiple")` from `scheme2.exp()`. Defunct since 2.1.0, same as qgraph and degreenet. Still errors on 2.3.3.9029. CRAN lists no URL or BugReports for it, so it is an email to the maintainer.
* **checked** passes `mode` positionally to `neighbors()`. That is soft-deprecated for 3.0.0 and only warns -- but the package's tests are snapshot tests, so 315 warnings become one failure. A one-word fix upstream.
* **htna** cascades through `tna:::as.igraph.matrix()`, whose call to `graph_from_adjacency_matrix()` ends in a trailing comma. That empty argument used to reach the migration handler and error; it does not any more, so the reproducer passes and is kept as a regression test rather than sent anywhere.
* **SEMdeep** fails inside `SEMgraph::SEMrun()` with a subscript out of bounds and no igraph function on the backtrace -- a silent result change, the same shape as SEMgraph's own open item. It is the one example here that is neither reduced nor rendered: SEMgraph needs Bioconductor packages that do not install where the others were rendered, so its `.md` carries the check log and a list of what to compare instead.

Rendered with reprex against a locally built igraph 2.3.3.9029, so the outputs are what the current development version actually does.

Claude-Session: https://claude.ai/code/session_01DdNk5vJPSUidKGdnAYt7Dv

### DIFF
--- a/revdep/examples/README.md
+++ b/revdep/examples/README.md
@@ -35,6 +35,18 @@ Each issue has two files:
 #### Group F — `tkplot()` defunct (igraph 3.0.0)
 13. **tkplot-defunct-issue** — shared reproducer for Boptbd, c3net, ggm, optbdmaeAT, optrcdmaeAT (all direct callers of `igraph::tkplot()`)
 
+#### Group C (continued) — new in the `revdep2` run 31048405399
+14. **bootcluster-sample-degseq-method-issue** — `sample_degseq(method = "simple.no.multiple")` defunct, same as qgraph/degreenet
+
+#### Group G — Optional arguments passed positionally (deprecated for 3.0.0)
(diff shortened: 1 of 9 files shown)


krlmlr:
Fold into #2646.

PR #2856, opened 2026-08-17
feat(ci): Add revdep3, pair checks in isolated containers on a shared revdepx core

krlmlr:
Proposes **revdep3**, the *pair engine*: the reverse-dependency check rebuilt on Docker containers. Its queue-engine sibling revdep4 (#2857) is **merged**; the two were designed as interoperable siblings on one shared core, and with that core now on `main` via #2857, this PR's diff is just the pair engine itself — `revdep3.yaml` + `revdep3/` (the branch has `main` merged in; the shared `.github/workflows/revdepx/` matched main byte-for-byte, and the committed `revdep/` record is main's).

## What revdep3 adds over the merged revdep4

Both engines isolate every check in its own container on the shared pipeline (plan → base → build → universe → shards → collect). They differ in where the concurrency sits:

- **revdep4 (merged)** runs a package's old and new halves *sequentially* and wins the time back across packages — 4 packages in flight per shard, one container each.
- **revdep3 (this PR)** keeps revdep2's *pair*: one package at a time per shard, its two halves running side by side in two containers. Per-package wall clock is halved; a package's verdict arrives in one step.

What the live runs measured about the trade: pair shards averaged 2.5 h for ~86 packages against queue shards' 4.05 h for ~172 — the pair engine buys per-package latency at moderately higher total runner cost. And because only two containers run at once, the pair engine is much gentler on the 4-core runners: the queue engine's measured per-half times run 1.5× the pair engine's wall clock at the median (p90 1.8×) from core contention, and the timeout analysis of the final revdep4 run showed all seven of its floor-timeouts were contention-driven — four of those packages finished in 421–715 s under this engine.

**The old half always runs fresh; a stored old result is a second opinion only.** Where the plan certifies an earlier run's old result as comparable (same revdep version, our CRAN version, container R, base-image tag, dependency fingerprint, within age), the fresh old check is compared against it: `baseline_agrees` on the manifest line, disagreements reported as drift. The stored result never substitutes for the check.

## Interoperability with the merged workflow

One artifact family (`revdepx-*`), one manifest/baseline/timings schema, one comparison code path, one universe-image lineage on GHCR, one concurrency group per ref. Either workflow reads the other's baselines as second opinions, consumes the other's timings, retries the other's runs (`retry-run` accepts a run id of either), and starts delta from whichever universe image was refreshed last. The second-opinion machinery's first live exercise ran this engine against revdep4's baselines: 3288 agree, 134 disagree — and the disagreements decomposed exactly into the two then-known infrastructure effects plus 31 true flakes (0.9%).

## Live-test results (this engine)

The validation run ([32196879628](https://github.com/igraph/rigraph/actions/runs/32196879628), `most`, depth 2, R 4.6.1) completed green end to end: 3435 reverse dependencies, 40 shards in two waves, all 45 jobs green, ~8.9 h wall clock — **3336 ok · 28 newly broken · 9 fail to install · 58 with uninstallable dependencies · 2 timeouts · 2 deferred**, and it was the first run to publish the universe image to GHCR (all 40 shards pulled it in minutes). It reproduced all 21 of revdep4's newly-broken and unmasked 4 more that revdep4's infrastructure failures had hidden. Its residual buckets (Bioconductor planner metadata, Rust and tcl toolchains, the memory cap) were fixed on both branches afterwards and validated by the reruns and the final revdep4 full run that produced the record now on `main`; this branch carries the identical fixes, so a fresh revdep3 run starts from the same platform.

## Notes for review

- The shared core arrived on `main` with #2857 and is untouched here; this PR was kept byte-identical to it throughout, so merging adds no second copy and no conflicts.
- Once merged, `revdep3.yaml`'s `workflow_dispatch` becomes registered like revdep4's; the temporary push trigger that drove the live run is long reverted — the workflow is dispatch-only with `strong`/depth-1 defaults and the check platform pinned at R 4.6.1.
- Whether to keep both engines or let revdep4 serve alone is a judgment call: the pair engine is the lower-latency, lower-contention option (better per-package timeouts, gentler on 4-core runners), the queue engine the cheaper-per-run one. They share everything but the shard driver, so the maintenance cost of keeping both is the ~600 lines under `revdep3/`.

---

https://claude.ai/code/session_01UB7YutLzVWU7xCTYUvF3kF


krlmlr:
Not proceeding.

Issue #2788, opened 2026-07-26
Comprehensive plan: argument renames for 3.0.0


krlmlr:
The argument-migration machinery built for 3.0.0
(#2757 rationale, #2779 registry, generated ARG_HANDLE recovery blocks)
makes argument renames feasible for the first time
without breaking user code:
a registry entry maps an old name to a new one,
legacy callers — positional, abbreviated, or using the old name —
are recovered, and a single `lifecycle::deprecate_soft("3.0.0", …)`
warns once per call while returning the correct result.
A rename is now a registry entry plus a signature/docs change.

This issue consolidates every open rename discussion
so the naming decisions can be made once, coherently,
and then executed mechanically through the registry.

Closes #691.
Closes #853.
Closes #692.
Closes #526.

## State of play (counts over the 529 exported, non-deprecated functions)

| family | spellings in use today | proposed direction |
|---|---|---|
| vertex selector | `v` 15 (betweenness, degree, distances, …) · `vids` 13 (closeness, strength, page_rank, …) · `nodes` 9 (ego family, constraint, alpha/power_centrality) · `vertex` 1 (cluster_spinglass) · `index` 11 (attribute accessors) | **decide: `v` vs `vids` vs `vertices`** — all three are on record (see below) |
| edge selector | `eids` 6 · `es` 4 · `e` 1 (edge_betweenness) | must mirror the vertex decision (#853: "If we go with `vids`, this should be `eids`. If we go with `v` this should be `e`.") |
| vertex count in generators | `nodes`-as-count 14 (sample_grg, sample_pref, sample_forestfire, centr_*_tmax, graph_from_graphdb, …) | `n` (62 functions already use it; #692 scheme, essentially uncontested) |
| weights | `weights` 52 · `weight` 1 (graph_from_graphnel) · `w` 2 (make_chordal_ring) · `vertex.weights` 3 vs `vertex_weights` 1 | `weights` (no existing issue; new finding) |
| algorithm selection | `method` 15 · `algorithm` 6 · `algo` 3 (page_rank, feedback_arc_set, feedback_vertex_set) · `implementation` 2 · `impl` 1 | `algorithm` for algorithm choice (#526); open: does `method` stay for comparison/isomorphism methods? |
[...]

Issue #526, opened 2022-04-07
Consistency in parameter names: algo vs algorithm


szhorvat:
At this moment, some functions use `algorithm` as a parameter names (e.g. `shortest_paths`, `sample_pa`, etc.), some others use `algo` (e.g. `page_rank`, `feedback_arc_set`, etc.), and yet others use `method` (e.g. `sample_degseq`, `sample_tree`, etc.). The naming should be made consistent across functions.

maelle:
Relevant lifecycle doc: https://lifecycle.r-lib.org/articles/communicate.html#renaming-an-argument

PR #2859, opened 2026-08-17
refactor: Rename `algo` to `algorithm` in `page_rank()`, `feedback_arc_set()` and `feedback_vertex_set()`

krlmlr:
Part of #2788 (PR 1 of a stack of 6, least controversial first); implements the `algo` → `algorithm` leg of #526.

## What this does

- Renames the `algo` argument to `algorithm` in `page_rank()`, `feedback_arc_set()` and `feedback_vertex_set()` — the three functions the #2788 survey lists under the `algo` spelling. All three already carry their 3.0.0 keyword-only migration entries with `algo` after `...`, so the rename is a pure registry change (`algo = algorithm` bare-symbol rename in `tools/migrations/`), and legacy `algo =` callers are recovered with the same single soft-deprecation as positional callers:

  ```
  Calling `page_rank()` with positional or abbreviated arguments was deprecated in igraph 3.0.0.
  i Detected call: page_rank(graph, algo)
  i Use instead: page_rank(graph, algorithm = )
  ```

- Abbreviations that could mean either spelling (`a`, `al`, `alg`) are rejected as ambiguous by the regenerated ARG_HANDLE guards.
- The deprecated `page.rank()` wrapper keeps its frozen `algo` formal (now documented locally with a deprecated badge, since `@inheritParams page_rank` no longer carries it) and forwards to `algorithm`.
- Tests: named uses updated, plus new coverage that the legacy name is recovered (snapshot for `page_rank(algo = )`, `expect_deprecated` for the two feedback functions).

## Out of scope

`method` (12 algorithm-choice + 2 comparison-metric functions), `implementation` (2) and `impl` (1) are the open "algorithm family" questions in #2788 and are not touched here.

https://claude.ai/code/session_01RTPj4qNv2FWui6etixZeuR

---
_Generated by [Claude Code](https://claude.ai/code/session_01RTPj4qNv2FWui6etixZeuR)_

### DIFF
--- a/R/centrality.R
+++ b/R/centrality.R
@@ -26,6 +26,8 @@ subgraph.centrality <- function(graph, diag = FALSE) {
 #' `page.rank()` was renamed to [page_rank()] to create a more
 #' consistent API.
 #' @inheritParams page_rank
+#' @param algo `r lifecycle::badge("deprecated")` Use `algorithm` in
+#'   [page_rank()] instead.
 #' @keywords internal
 #' @export
 page.rank <- function(
@@ -42,7 +44,7 @@ page.rank <- function(
   lifecycle::deprecate_warn("2.0.0", "page.rank()", "page_rank()")
   page_rank(
     graph = graph,
-    algo = algo,
+    algorithm = algo,
     vids = vids,
     directed = directed,
     damping = damping,
@@ -1884,7 +1886,7 @@ hub_score <- function(
 #'
 #' @param graph The graph object.
 #' @inheritParams rlang::args_dots_empty
-#' @param algo Character scalar, which implementation to use to carry out the
+#' @param algorithm Character scalar, which implementation to use to carry out the
 #'   calculation. The default is `"prpack"`, which uses the PRPACK library
 #'   (<https://github.com/dgleich/prpack>) to calculate PageRank scores
 #'   by solving a set of linear equations. This is a new implementation in igraph
@@ -1953,7 +1955,7 @@ hub_score <- function(
 page_rank <- function(
   graph,
   ...,
-  algo = c("prpack", "arpack"),
+  algorithm = c("prpack", "arpack"),
   vids = NULL,
   directed = TRUE,
   damping = 0.85,
@@ -1964,7 +1966,7 @@ page_rank <- function(
   # BEGIN GENERATED ARG_HANDLE: page_rank, do not edit, see tools/generate-migrations.R
   # fmt: skip
   if (...length() > 0L) {
-    .arg_ambiguous <- base::intersect(base::names(base::substitute(...())), base::c("d"))
+    .arg_ambiguous <- base::intersect(base::names(base::substitute(...())), base::c("a", "al", "alg", "d"))
     if (base::length(.arg_ambiguous) > 0L) cli::cli_abort("Argument {.arg {(.arg_ambiguous[[1L]])}} matches multiple arguments of {.fn page_rank}.")
     # Pre-3.0.0 signature: page_rank(graph, algo, vids, directed, damping, personalized, weights, options)
     .old_signature <- function(algo, vids, directed, damping, personalized, weights, options, ...) {
@@ -1975,7 +1977,7 @@ page_rank <- function(
         cli::cli_abort(base::c("Unexpected argument passed to {.fn page_rank}: {.arg {(.arg_extra)}}.", i = "Arguments after {.arg ...} must be spelled out in full."), call = base::parent.frame())
       }
       base::c(
-        if (!base::missing(algo)) base::list(algo = algo),
+        if (!base::missing(algo)) base::list(algorithm = algo),
         if (!base::missing(vids)) base::list(vids = vids),
         if (!base::missing(directed)) base::list(directed = directed),
         if (!base::missing(damping)) base::list(damping = damping),
@@ -1988,7 +1990,7 @@ page_rank <- function(
     if (base::length(.arg_handle) > 0L) {
       .arg_names <- base::names(.arg_handle)
       .arg_conflict <- base::intersect(.arg_names, base::c(
-        if (!base::missing(algo)) "algo",
+        if (!base::missing(algorithm)) "algorithm",
         if (!base::missing(vids)) "vids",
         if (!base::missing(directed)) "directed",
         if (!base::missing(damping)) "damping",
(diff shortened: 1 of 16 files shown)


schochastics:
Reviewed the rename itself and it looks correct: no drift in the generated blocks, no stale `algo =` callers left apart from the `*_impl()` C wrappers (which rightly keep it), and `page.rank()` forwards `algorithm = algo` by exact name so it doesn't double-warn.

I did find one regression in the generated ambiguity guard, fixed in 47d1fa6.

## The bug

`ambiguous_tags()` in `tools/generate-migrations.R` `charmatch()`ed a candidate prefix against `match_names` — the union of renamed-away old names and new tail names. For a rename whose old name is a prefix of the new one (`algo` → `algorithm`), `alg` matches both entries, so `charmatch()` returned 0 and the tag was emitted into the guard. But both candidates resolve to the *same* argument, so nothing was actually ambiguous:

```r
page_rank(g, alg = "arpack")       # soft deprecation before, hard error after
feedback_arc_set(g, alg = "exact_ip")  # newly guarded by this PR
```

The error message was also factually wrong ("matches multiple arguments of `page_rank()`"), and it contradicted `tools/migrations/README.md`. This isn't specific to these three functions — every rename in the #2788 stack where the old name prefixes the new one would hit it.

## The fix

A prefix is ambiguous only when the names it matches resolve to *different* targets, so the predicate now dedupes on `match_to`:

```r
hits <- startsWith(names, p)
length(unique(entry$match_to[hits])) > 1L
```

Regenerating narrows or drops four guards, and keeps the ones that matter:

- `page_rank()`: `c("a", "al", "alg", "d")` → `c("d")`. `d` is genuinely ambiguous (`damping` vs `directed`) and still errors; `alg =` recovers under the deprecation.
- `feedback_arc_set()` / `feedback_vertex_set()`: guard gone — it only ever held the false `algo`/`algorithm` prefixes.
- `migration_fixture()`: guard gone too. It was flagging `w`…`weigh` for exactly the same reason (`weight` → `weights`), so this was a pre-existing false positive in the fixture, unrelated to the rename.
- `migration_fixture_shadow()`: still rejects `at =`, the one case where the two candidates really do have distinct targets (`weights` vs the retained deprecated `attr`).

## Test changes

- `weig =` moves from the error-message snapshots to the deprecation snapshots in `test-migration-fixture.R`.
- New generator-level test for `ambiguous_tags()`: a prefix-rename yields no tags, two distinct targets still yield `"d"`.
- `test-centrality.R` gains `page_rank(star, alg = "prpack")` (recovers) and `page_rank(star, d = 0.5)` (still errors), so the distinction is pinned at the function level.
- Comments in `generate-migrations.R` and `tools/migrations/fixture.R` updated to state the rule correctly.

Full suite green locally: 9311 pass, 0 fail.

I kept this in the same PR rather than splitting it: the bug only becomes user-visible through this rename, and the fix changes these three functions' generated blocks, so a separate PR would leave this one mergeable-but-broken and need a rebase on the generated code anyway.

R/centrality.R
```diff
@@ -42,7 +44,7 @@ page.rank <- function(
   lifecycle::deprecate_warn("2.0.0", "page.rank()", "page_rank()")
```
maelle, line 44: When will we get rid of those functions? Could we try `deprecate_stop()` in a PR? They're dead weight we're carrying around.

R/centrality.R
```diff
@@ -1975,7 +1977,7 @@ page_rank <- function(
         cli::cli_abort(base::c("Unexpected argument passed to {.fn page_rank}: {.arg {(.arg_extra)}}.", i = "Arguments after {.arg ...} must be spelled out in full."), call = base::parent.frame())
       }
       base::c(
-        if (!base::missing(algo)) base::list(algo = algo),
+        if (!base::missing(algo)) base::list(algorithm = algo),
```
maelle, line 1980: why do we namespace base functions?

tests/testthat/_snaps/centrality.md
```diff
@@ -59,3 +59,13 @@
...
       ! Can't use unkown ARPACK options: unknown_thing1, unknown_thing2
 
+# page_rank(algo = ) is deprecated but still works
+
+    Code
+      res_legacy <- page_rank(star, algo = "prpack")
+    Condition
+      Warning:
+      Calling `page_rank()` with positional or abbreviated arguments was deprecated in igraph 3.0.0.
+      i Detected call: page_rank(graph, algo)
```
maelle, line 69:
```suggestion
      i Detected call: page_rank(graph, algo = )
```

I wish this looked like this so that the two lines might have real parallelism.

R/centrality.R
maelle: #2894

tests/testthat/test-migration-fixture.R
```diff
@@ -316,6 +315,37 @@ test_that("the generated block is in sync with the registry", {
   expect_identical(spliced$lines, lines)
 })
 
+test_that("ambiguous_tags() only flags prefixes with distinct targets", {
+  generator <- testthat::test_path("..", "..", "tools", "generate-migrations.R")
```
maelle, line 320:
```suggestion
  generator <- test_path("..", "..", "tools", "generate-migrations.R")
```

tests/testthat/test-migration-fixture.R
```diff
@@ -316,6 +315,37 @@ test_that("the generated block is in sync with the registry", {
...
+  skip_if_not(file.exists(generator))
+  gen_env <- new.env()
+  sys.source(generator, envir = gen_env)
```
maelle, line 323: could we have something simpler?

tests/testthat/_snaps/centrality.md
krlmlr: `algo =` in the given case, but `algo` if called with a positional argument?

tests/testthat/test-migration-fixture.R
krlmlr: Side notes: `tools/` is special, this is what is available to the `./configure` scripts.

Did we move the generator from `R/` in the past? Why? If we test it, it could well be a file in `R/` and rbuildignore-d (alongside the test file).

maelle:
https://github.com/ropensci/mctq/blob/main/R/.post-render.R  https://github.com/ropensci/mctq/blob/f871eb3a79e3a118d65959b213e119b1873f4d7d/.Rbuildignore#L22

PR #2870, opened 2026-08-28
docs: other roxygen2 warning

maelle:
Or should it be `Matrix::solve()`?


schochastics:
I think this should be `Matrix::solve()`

PR #2888, opened 2026-09-04
perf: Build the whole vertex-sequence list in one C pass

schochastics:
Split 3 of 4 out of #2696, per https://github.com/igraph/rigraph/pull/2696#issuecomment-4989953620 — this is the requested "C implementation of `create_vs_list()` **without ALTREP**" step.

**Base:** #2887 (review #2886 and #2887 first; this PR's diff is only the top commit). Last in the stack: #2696.

### What changed

After #2887, `create_vs_list()` still drove the per-element work from R — an `lapply` closure call, `as.integer()`, a name subset and `attributes<-`, once per sequence. That is the dominant remaining cost when a function returns thousands of `igraph.vs` objects.

`Rx_igraph_vs_list()` in `rinterface_extra.c` does the whole loop in one pass. R only builds `V(graph)` once (that is what mints the shared weak reference and graph id) and hands the pieces to C:

```r
create_vs_list <- function(graph, idx_list) {
  verts <- V(graph)
  .Call(
    Rx_igraph_vs_list,
    idx_list,
    if (is_named(graph)) vertex_attr(graph)$name else NULL,
    attr(verts, "env"),
    attr(verts, "graph")
  )
}
```

Each element gets a fresh integer payload, the corresponding names when the graph is named, the shared `env`/`graph` attributes, and the `igraph.vs` class. Names are built **eagerly** here — no ALTREP, as requested. That is the only thing #2696 changes on top.

Note the payload is duplicated when `Rf_coerceVector()` is a no-op, so a caller-owned integer vector is never mutated in place.

`max_cliques(sample_gnp(200, 0.16))` on a named graph: **5.6 ms -> 2.65 ms**.

### Where that leaves the stack

| | median | vs `main` |
|---|---|---|
| `main` | 24.7 ms | — |
| #2886 (shared weakref) | 7.8 ms | 3.2× |
| #2887 (`create_vs_list()`) | 5.6 ms | 4.4× |
| **this PR** (C loop) | **2.65 ms** | **9.3×** |

So the three merged PRs get essentially the whole win without any ALTREP. This is the point at which #2696's "if needed" can be judged on its own merits.

### Correctness

Verified against the `return.vs.es = FALSE` numeric path:

- integer payloads with identical values and names;
- no `names` attribute at all on unnamed graphs;
- `NA` and out-of-range IDs map to `NA_STRING`;
- double ID vectors coerced correctly;
- caller's input vectors unmutated;
- the shared weak reference still releases the graph: `get_vs_graph(seq)` is `NULL` after `rm(graph); gc()`.

Compiles without new warnings, and clean under `gctorture(TRUE)`.

Full suite: **FAIL 0 | WARN 0 | SKIP 7 | PASS 9299**.

### Also here

`cpp11::cpp_register()` added only the `Rx_igraph_vs_list` registration in `src/cpp11.cpp`. Two touchstone benchmarks for the batch path: `ego_order2_named` and `all_simple_paths_named`.

### DIFF
--- a/src/rinterface_extra.c
+++ b/src/rinterface_extra.c
@@ -2611,6 +2611,70 @@ static void *Rx_igraph_altrep_to(SEXP vec, Rboolean writeable) {
 static R_altrep_class_t Rx_igraph_altrep_from_class;
 static R_altrep_class_t Rx_igraph_altrep_to_class;
 
+/* Batch constructor for a list of vertex sequences.
+ *
+ * Builds the whole `lapply(idx_list, unsafe_create_vs, ...)` result in one C
+ * pass: for each vertex-ID vector it produces a fresh integer payload, attaches
+ * the corresponding vertex names (when the graph is named), and sets the shared
+ * `env` weak reference, the `graph` id and the `igraph.vs` class. This keeps the
+ * per-object R overhead (closure call, `as.integer`, name subset,
+ * `attributes<-`) out of the loop entirely.
+ *
+ *   idx_list  : VECSXP of vertex-ID vectors (integer or double)
+ *   names_src : graph's full vertex-name STRSXP, or NULL for unnamed graphs
+ *   env       : the shared weak reference (or env) to set as the "env" attr
+ *   graph_id  : graph id (character scalar), or NULL to skip the "graph" attr
+ */
+SEXP Rx_igraph_vs_list(SEXP idx_list, SEXP names_src, SEXP env, SEXP graph_id) {
+  R_xlen_t n=XLENGTH(idx_list);
+  int named=(TYPEOF(names_src) == STRSXP);
+  R_xlen_t nsource=named ? XLENGTH(names_src) : 0;
+  SEXP env_sym=Rf_install("env");
+  SEXP graph_sym=Rf_install("graph");
+  SEXP out=PROTECT(Rf_allocVector(VECSXP, n));
+  SEXP cls=PROTECT(Rf_mkString("igraph.vs"));
+
+  for (R_xlen_t i=0; i < n; i++) {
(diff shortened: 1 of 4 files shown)


touchstone/script.R
```diff
@@ -331,5 +331,50 @@ benchmark_run(
...
+# ---------------------------------------------------------------------------
+# Group #6 - batch construction of many vertex sequences
+# One shared graph reference and one hoisted name source, with the whole
+# per-element loop in C, instead of per-object R work. These two calls each
+# return thousands of vertex sequences from a single C core call.
+# ---------------------------------------------------------------------------
+
+# ego() returns one vertex sequence per node -- a few thousand sequences built
+# in one call. Exercises create_vs_list() through neighborhood().
+benchmark_run(
```
krlmlr: Bubble this up to the first PR?

src/rinterface_extra.c
```diff
@@ -2611,6 +2611,70 @@ static void *Rx_igraph_altrep_to(SEXP vec, Rboolean writeable) {
...
+      payload=PROTECT(Rf_duplicate(elt));
+    }
+
+    if (named) {
+      R_xlen_t len=XLENGTH(payload);
+      const int *pidx=INTEGER(payload);
+      SEXP nm=PROTECT(Rf_allocVector(STRSXP, len));
+      for (R_xlen_t k=0; k < len; k++) {
+        int j=pidx[k];
+        if (j == NA_INTEGER || j < 1 || j > nsource) {
```
krlmlr: Why do we need to compare with `nsource`? Is this an artefact of a subsetting operation gone sideways? Need to do some Git archaeology to understand.

PR #2908, opened 2026-09-15
docs: integrate the roclet to igraph proper

### LINKED ISSUE #2899: Consider integrating igraph.r2cdocs here
The logic is in a single file, it would simplify processes as we work on integrating v1.0.x .

https://github.com/igraph/igraph.r2cdocs/pull/12#issuecomment-5651393750

maelle:
Fix #2899


R/utils-roxygen2.R
```diff
@@ -0,0 +1,314 @@
...
+  subcomponent <- names(subcomponent(
+    treesitter_graph,
+    vertex,
+    mode = "out"
+  ))
+  subcomponent[endsWith(subcomponent, "_impl")]
+}
+
+
+# https://blog.r-hub.io/2021/07/30/cache/#function-factory
```
maelle: I instructed Claude to add caching through that approach; igraph.r2cdocs has a dependency on memoise.

maelle:
The current code has deps on various packages. Some of them we can remove like tibble or purrr, some of them we have to use (treesitter). Before I go about adding fixes / workarounds, I'd like to confirm we want to move the roclet infrastructure into this package.

```r
'::' or ':::' imports not declared from:
     ‘fs’ ‘purrr’ ‘roxygen2’ ‘tibble’ ‘treesitter’ ‘treesitter.r’ ‘xml2’
```

krlmlr:
Adding these as suggested packages shouldn't be a big deal.

krlmlr:
Or as "Enhanced"?

krlmlr:
Can also add memoise.

